PR #9, opened 2026-09-10
feat(hub): the public wizards get a token layer that follows the site

fiorelorenzo:
## What this changes

The hub's admin area and its two public wizards shared one `tokens.css`, copied verbatim from the CRM's application shapes: a 10px radius and a blurred `--shadow-lg`. `joinorbiters.com`'s own pages draw none of that (`landing.css:9`, `landing.css:31`: no radius anywhere, a solid 2px ink border, an 8px/6px offset shadow that is the border colour moved, never a blur). The chooser, the two wizards and the thanks page are reached mid-sentence from the landing's own call to action, so which system they follow was undecided.

`docs/design/DECISIONS.md` now records the decision: the site's. `tokens.css` gains a `.site` scope restating the landing's own values for whichever element carries that class: `--radius: 0`, a solid ink `--border`/`--input`, an 8px/6px offset `--shadow-xs`..`--shadow-lg`, and the `--landing-*` colours (which already equal the application's own `--primary`/`--ring`, since both read the same shared palette). The admin area is untouched, since nothing there carries `.site`.

This is the decision and the token layer only. No markup changes: ORB-74 (page chrome) applies the class to `Shell`, and ORB-75 (fields and buttons) draws the wizard's own shapes with it.

## How I verified it

- `pnpm --filter hub test`: 25 tests, 6 files, all passed, including the new `src/styles/tokens.test.ts`, which pins the rules the decision creates (no `--radius` above 0 and no blur term in any `--shadow-*` inside `.site`) and cross-checks the `--landing-*` colours against the same shared palette `landing-tokens.test.ts` already pins them against.
- `pnpm --filter hub lint`: clean.
- `pnpm --filter hub build`: `vite build && tsc --noEmit`, clean; Tailwind compiled the new `.site` scope with no error.
- `preflight --list` on this diff selected `hub-web-lint`, `hub-web-unit`, `hub-web-build` and `identity-names`; `preflight` ran all four and they passed.

The scope is a plain class, `.site`, rather than a data attribute, matching the existing `.dark` scope in the CRM's own `tokens.css` (same mechanism: a descendant re-declares a handful of custom properties, and every derived token that reads `var(--radius)` or `var(--shadow-xs)` picks it up through ordinary inheritance). ORB-74 needs to know to apply it to `Shell`'s root; I named it in the DECISIONS.md row and in a comment at the top of `tokens.css` rather than adding it to markup myself, since restyling `Shell` is that issue's job.

## Screenshots

None. This issue changes no surface deliberately: the `.site` scope is not applied to any element yet, so nothing in the running app renders differently. `pnpm --filter hub build` shows the compiled CSS is unchanged in size in any way that reaches current markup.

Linear: ORB-73.


fiorelorenzo:
Independent review found a blocker: the four `--shadow-*` declarations in `.site` were inert. Tailwind v4 decomposes a compound theme value (`0 1px 1px var(--shadow-ink-weak)`) at build time, baking the shape into every `shadow-xs` utility and keeping only the innermost `var()` live, so a `.site` override of `--shadow-xs` itself never reached a rendered element. Confirmed by compiling this exact file and reading `.shadow-xs` back out of the built CSS: it read `var(--shadow-ink-weak)`, never `var(--shadow-xs)`.

Fixed in 67bbbef: the application's shadow shapes move into `:root` as `--shadow-app-xs/sm/md/lg`, `@theme inline` points `--shadow-xs..lg` at each as a bare `var()` (which Tailwind keeps live instead of decomposing), and `.site` now overrides `--shadow-app-*`. Re-verified against the rebuilt dist CSS: `.shadow-xs{--tw-shadow:var(--shadow-app-xs);...}`, and a new test pins that indirection so it cannot silently regress.

Also fixed from the same review, all should-fix/nit: the CTA contrast test now reads `--landing-cta-ink` instead of a hard-coded white; the step-ordering test asserts the 8px magnitude and the 0.75 ratio it claimed to check rather than only comparing text; the unused, unpinned `--landing-grid`/`--landing-cell` tokens are dropped until ORB-74 actually draws the ground and can pin them; and the border-width comment now points at the lines that actually carry it (landing.css:141, 156, 250, 264) instead of the universal selector, which sets width 0.

pnpm --filter hub test (26 tests, 6 files), pnpm --filter hub lint and pnpm --filter hub build are all green after the fix.

PR #10, opened 2026-09-10
fix(website): the header and footer links read as bordered controls

fiorelorenzo:
## What this changes

The header's "Accedi" and the footer's links painted a bare `--landing-surface`
rectangle with no border, no hover surface and no floor on touch-target size, so they
read as neither link nor button; the brand caught the same borderless patch even
though it isn't a control at all. This changes what `.top .quiet-link`, `footer
.quiet-link` and `.brand` all now share: a 2px ink border on the page's own ground (no
shadow, since that tier is `.box`/`.card`'s signal, not a link's), a `min-height` of
44px so the footer's smaller type still clears the touch-target floor, and a hover
state that inverts to the ink surface. The scope moved from `.top a` to `.top
.quiet-link` so the brand gets its own rule instead of a patch meant for something it
isn't, and stayed on `footer .quiet-link`, which also carries privacy.html and
termini.html's own two-link footers since all three pages share this stylesheet. The
generic `.quiet-link` rule is untouched and keeps its underline, since the same class
doubles as an inline citation in privacy.html and termini.html's running text.

The secondary hero CTA, "Cerchi persone? Raccontaci il progetto", wrapped to two or
three lines at 390px while the primary door stayed on one, so the two CTAs had
different heights. Shortened it to "Cerchi persone?", which fits one line at both
viewports and matches the primary CTA's height without a height rule fighting real
content.

## How I verified it

- `pnpm --filter website test`: 189 passed (188 before, plus one new regression test
  pinning the border, `min-height`, hover surface and no-shadow rule on the new
  header/footer control shell).
- `preflight` (the 7 checks it selected for this diff): `website-lint`,
  `website-types`, `website-unit`, `identity-names`, `website-build`, `website-e2e`,
  `website-image`, all passed in 40.2s.
- `uishot http://localhost:4173/pigrocrm --viewports mobile:390x844 --axe --fail-on
  serious` and the same at `desktop:1440x900`: zero violations at both viewports.
- Drove a real keyboard `Tab` to "Accedi" and hovered the footer's "Privacy" link with
  Puppeteer: the focus ring (`--landing-focus`, offset 3px) is visible on the bordered
  box, and hover inverts the surface to ink with `--landing-cta-ink` text.
- Rendered `/privacy` and `/termini` (they share `landing.css`) and confirmed their
  own two-link footers picked up the same treatment, and that `.quiet-link` used
  inline in their running text (`humancraft.tech`, `joinorbiters.com`) stayed
  underlined prose rather than turning into a boxed control mid-sentence.

I did not touch the cookie notice's "Dettagli" link or its "No"/"Va bene" buttons
(`system.css`, `.consent`), even though the issue also measures those under the 44px
floor. That component is deliberately a different, quieter visual tier by its own
comment ("a coloured call to action for a cookie notice would shout louder than the
page it sits on"), growing its touch targets is a design decision of its own, and it
is shared with orbiters.html, which is outside this issue's page. Filed as ORB-87.

`orbiters.css` makes the same borderless-patch move for its own footer (the comment at
`landing.css:101` says so), and this PR does not touch it: `orbiters.html` is a
separate page with its own stylesheet, ORB-64 is scoped to `/pigrocrm`, and nobody
else in this wave owns that file either. Filed as ORB-88 so the same fix lands there
deliberately rather than by accident.

I gave the brand ("Orbiters" / "PigroCRM") the same bordered-box treatment as the
links rather than removing its background patch, since the issue names the brand's
patch as the same defect (a bare rectangle, not that it needs no ground at all): the
header sits on the decorative field and the wordmark needs a sliver of the page's own
ground to stay legible over a dark tile the same as before.

## Screenshots

**1. Header, 1440px**
Header at 1440, before and after

**2. Header, 390px**
Header at 390, before and after

**3. Footer, 1440px**
Footer at 1440, before and after

**4. Footer, 390px**
Footer at 390, before and after

**5. Hero CTAs, 390px: the secondary door no longer wraps to three lines**
Hero CTAs at 390, before and after

Linear: ORB-64.

### DIFF
--- a/projects/website/src/index.html
+++ b/projects/website/src/index.html
@@ -50,7 +50,7 @@
             </p>
             <p class="actions">
               <a class="cta" href="/hub/freelance">Entra come talento</a>
-              <a class="cta secondary" href="/hub/aziende">Cerchi persone? Raccontaci il progetto</a>
+              <a class="cta secondary" href="/hub/aziende">Cerchi persone?</a>
             </p>
             <p class="fine">
               Due minuti, una domanda alla volta. Ti chiediamo il CV e quanto costa una tua
--- a/projects/website/src/landing-style.test.ts
+++ b/projects/website/src/landing-style.test.ts
@@ -190,9 +190,16 @@ describe('the landing shares the product system', () => {
     expect(hover).toMatch(/background-color:\s*var\(--landing-ink\)/)
     expect(hover).toMatch(/color:\s*var\(--landing-cta-ink\)/)
 
+    // The brand doesn't invert on hover the way the two links do: half its glyph
+    // is drawn in the same colour as the ink ground, so it gets an offset shadow
+    // instead, still without a blur.
+    expect(rule('.brand:hover')).toMatch(/box-shadow:\s*4px 4px 0 var\(--landing-ink\)/)
+
     // The same class doubles as an inline citation in privacy.html and
     // termini.html's running text, which must stay underlined prose rather
-    // than turn into a button mid-sentence.
+    // than turn into a button mid-sentence: a later edit that moves the box
+    // properties onto the generic rule would be the regression to catch.
     expect(rule('.quiet-link')).toMatch(/text-decoration:\s*underline/)
+    expect(rule('.quiet-link')).not.toMatch(/border-width|min-height|display:/)
   })
 })
--- a/projects/website/src/landing-style.test.ts
+++ b/projects/website/src/landing-style.test.ts
@@ -165,4 +165,34 @@ describe('the landing shares the product system', () => {
     expect(rule('.cta')).not.toMatch(/border-radius/)
     expect(rule('.cta:hover')).toMatch(/background-color:\s*var\(--landing-ink\)/)
   })
+
+  it('draws the header and footer links as bordered controls, not a borderless patch (ORB-64)', () => {
+    // The brand, "Accedi" and the footer links used to be `.top a, footer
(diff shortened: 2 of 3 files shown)


fiorelorenzo:
Independent review findings and what I did with each:

1. **The brand had no hover state though the description claimed it did.** True: the shell rule gave `.brand` the same border/background as the links, but only `.top .quiet-link`/`footer .quiet-link` got a hover. Fixed: `.brand:hover` now gets an offset box-shadow rather than the ink invert the links use, since two of the glyph's four tiles are drawn in `--color-prussian-blue`, the same colour as `--landing-ink`, and inverting would have erased them. Pinned in the regression test.

2. **The regression test didn't defend the prose .quiet-link staying unboxed.** True. Added a negative assertion (`not.toMatch(/border-width|min-height|display:/)`) on the generic `.quiet-link` rule, so a later edit that moves the box treatment onto the prose form fails the suite instead of silently turning an inline citation into a button mid-sentence.

3. **footer .quiet-link's margin applied to every link, indenting the first away from the content edge.** True on privacy.html and termini.html's plain two-link footers, where the first link picked up an unwanted 3.2px inset against the paragraph text above it. Switched to `footer .quiet-link + .quiet-link`, so only the gap between consecutive links gets the margin.

All three applied in a second commit. Re-ran `pnpm --filter website test` (189 passed), `preflight` (7/7 passed) and `uishot --axe --fail-on serious` at both viewports (zero violations) after the fix.

PR #11, opened 2026-09-10
fix(website): the privacy and terms pages sign themselves as Orbiters

fiorelorenzo:
## What this changes

`privacy.html` and `termini.html` titled and signed themselves as PigroCRM, but they
are served on joinorbiters.com (not pigro.joinorbiters.com), and it is the Orbiters
signup form under the community page that links to `/privacy`, so a visitor who clicks
Privacy there landed on a tab naming a different product. Their own text already covers
Orbiters' data (the signup section) alongside PigroCRM's hosted spaces, and
`orbiters.test.ts:109` already asserted `privacy.html` names Orbiters and links
`/orbiters`; the title and the brand link were the two spots that had not followed.

Now both pages' `<title>`, `og:title` and header brand link (the four-tile glyph plus
the name) read Orbiters, matching the site they are on and `index.html`'s and
`orbiters.html`'s own signature. Nothing else moved: the titolare del trattamento
(Humancraft di Ivan Sala), the data controller and processor language, the Gmail and
Drive scope text, the cookie and consent paragraphs, the licence and warranty
sections, every body paragraph that legitimately still names PigroCRM as the product
being described, and every link target (owned by ORB-66, running alongside this) are
untouched.

`landing-pages.test.ts`'s title and brand-link assertions are rewritten with a dated
comment, not deleted: `index.html` keeps `PigroCRM` in its title because it is
PigroCRM's own landing page, served at `/pigrocrm`, unrelated to Ivan's separate
2026-09-09 «freelance» exception (ORB-24, `docs/design/positioning.md` line 85); the
two legal pages carried no such exception, so they now assert `Orbiters` instead. The
`og:title` assertion, previously only checked for being non-empty, now pins the same
brand, since a share card is exactly where the old defect would have kept drifting
back unnoticed.

## How I verified it

- `pnpm --filter website test`: 189 tests pass, including the two rewritten
  assertions in `landing-pages.test.ts` and the untouched `orbiters.test.ts:107-112`
  cross-check against `privacy.html`.
- `pnpm --filter website build` and `pnpm --filter website exec tsc --noEmit`: clean.
- `pnpm --filter website test:e2e`: 49 tests pass (run against a temporary local
  port override of `vite.config.ts`/`playwright.config.ts` only, immediately reverted
  before committing, since port 4173 was held by a sibling agent's own e2e run on this
  shared devbox at the time).
- `WEBSITE_WEB_PORT=127.0.0.1:8099 docker compose build` (`projects/website`): image
  builds clean, nginx config unchanged.
- `orbiters-identity-scan .`: clean against 28 names, 3 paths allowed (unchanged).
- `uishot --axe --fail-on serious` against the built pages, before and after, at
  1440x900 and 390x844: zero serious/critical violations in every run.

The naming call itself: I moved the two legal pages to sign as Orbiters rather than
PigroCRM, reasoned from where they are served, who links to them, and what their own
text already covers, per ORB-36. This does not touch the titolare del trattamento or
any substantive claim in either policy; if the naming call should have gone the other
way, that is a brand decision for Lorenzo/Ivan, not something either page's legal
content depends on.

An independent reviewer on this PR flagged the same title-exception comment I had
misattributed and the unpinned `og:title` assertion above; both are fixed in the
second commit.

## Screenshots

Before and after, 1440px and 390px, with an orange box around the changed header
brand link (the browser tab title changed too, not capturable in a page screenshot).

**1. privacy.html, 1440px**
privacy.html before and after, 1440px

**2. privacy.html, 390px**
privacy.html before and after, 390px

**3. termini.html, 1440px**
termini.html before and after, 1440px

**4. termini.html, 390px**
termini.html before and after, 390px

Linear: ORB-36.

### DIFF
--- a/projects/website/src/landing-pages.test.ts
+++ b/projects/website/src/landing-pages.test.ts
@@ -20,7 +20,14 @@ describe.each(PAGES)('%s', (name) => {
 
   it('carries its own title, description and Open Graph', () => {
     const title = page.match(/<title>([^<]+)<\/title>/)?.[1] ?? ''
-    expect(title).toContain('PigroCRM')
+    // Until 2026-09-10 every page here titled itself PigroCRM, including the two
+    // legal pages. ORB-36: privacy.html and termini.html are served on
+    // joinorbiters.com, not on pigro.joinorbiters.com, and it is the Orbiters signup
+    // form that links to them, so they title themselves after the site they are on
+    // rather than after the CRM. index.html keeps PigroCRM in its title under
+    // Ivan's 2026-09-09 exception (ORB-24) for search continuity; see the
+    // brand-link assertion below for the same split.
+    expect(title).toContain(name === 'index.html' ? 'PigroCRM' : 'Orbiters')
     const description = meta(page, 'description') ?? ''
     expect(description.length).toBeGreaterThan(40)
     // The trap named in spec 9.2: the previous system's index.html still carries "Studio Rossi is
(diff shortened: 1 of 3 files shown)


fiorelorenzo:
Independent review from a fresh reviewer agent, findings applied in the second commit (bd3ff0d):

- The comment above the rewritten title assertion attributed index.html's `Con PigroCRM gratis` suffix to Ivan's 2026-09-09 freelance exception (ORB-24). positioning.md:85 records that exception for the word «freelance» only. Fixed: the comment now says index.html keeps PigroCRM because it is the CRM's own landing page, served at /pigrocrm, and names that as unrelated to the freelance exception.
- `og:title` was only asserted `toBeTruthy()`, so half of the fix (what a visitor sees when the page is shared rather than opened) stayed unpinned. Fixed: all three pages' og:title now asserts `toContain('Orbiters')`, since all three already read that way and index.html's og:title never repeated the PigroCRM suffix from its `<title>`.
- Confirmed everything else: scope is exactly the three files intended, no link targets changed, the titolare del trattamento and every substantive policy claim are untouched, and the rewritten assertions are logically correct for all three pages.
- PR title and body also fixed for length (72-char limit) and the no-em-dash convention, both flagged in the same pass.

PR #14, opened 2026-09-10
test(core): the free-planner assertion in test_trgm_escape.py holds on any machine

fiorelorenzo:
## What this changes

`test_the_planner_left_alone_still_prefers_a_sequential_scan_at_this_scale` in
`test_trgm_escape.py` asserted a specific plan string (`Seq Scan on customers`) on a
table whose statistics came from `db_session`'s per-test savepoint. `ANALYZE` runs fine
inside that savepoint, but `relpages` reflects the table's real physical size, and a
savepoint's rollback never shrinks a heap file back down. Every earlier test in the same
file (and, at package scale, every earlier test in the whole suite) inserts and rolls
back the same 500 rows into `customers`, and the pages that leaves behind are still
there to be counted. Measured locally: running the file in file order left `customers`
at 52 pages instead of a fresh build's 11, and priced the free planner's own `Seq Scan`
at 58.25 instead of 17.25 -- margin enough that a hosted runner, holding a different
amount of leftover bloat, priced the trigram index cheaper instead (56.25..60.26, per
ORB-9) with no line of code in between.

The fix gives the planner a corpus whose physical state is a property of this test
alone. `customers_at_reference_scale` commits its own REFERENCE-scale corpus outside
`db_session`'s savepoint (`VACUUM` cannot run inside one at all), runs a plain `VACUUM`
on `customers` *before* inserting -- not only `ANALYZE`/`VACUUM (ANALYZE)` after, which
is what `test_search_plan.py`'s `inflated` fixture does and which is not enough at this
row count, see below -- and `VACUUM (ANALYZE)`s again after, so the estimate describes
this corpus and nothing that ran before it. It cleans up and re-vacuums on the way out,
the same way `inflated` already does, so it doesn't leave `customers` bloated for
whichever test runs after it.

I also added a falsifier,
`test_the_seq_scan_above_stops_appearing_once_the_planner_loses_the_choice`: on the
identical corpus and the identical query, taking `enable_seqscan` away leaves no
`Seq Scan on customers` in the plan at all, so the assertion above is provably not one
that would hold for any plan Postgres could produce.

`test_the_planner_left_alone_still_prefers_a_sequential_scan_at_this_scale` no longer
carries `@pytest.mark.planner`: it is deterministic now, so it runs in the normal gate
instead of the preflight-only lane the marker used to route it into (the CI/preflight
`-m "not slow and not planner"` and `-m "planner"` filters need no changes, since they
already select per-marker; `test_sort_plan.py` still carries the marker for its own,
separate, out-of-scope sort-plan assertions).

## How I verified it

- **Reproduced the exact failure first.** Before the fix, running the whole package's
  suite ahead of this file (`uv run pytest -q projects/pigrocrm/packages/core/tests
  -m "not slow"`, 3077 tests) failed this one test with
  `cost=56.25..60.26 rows=1 width=16` for `Bitmap Heap Scan on customers` -- the exact
  numbers ORB-9 measured on the hosted runner, reproduced locally by table-bloat alone.
  A first version of the fix (commit-and-`VACUUM (ANALYZE)`-after only, mirroring
  `inflated` verbatim) still failed the same way in the same scenario, because
  `VACUUM (ANALYZE)` after an insert doesn't truncate pages that were already dead
  before the insert landed on top of them. Adding the pre-insert `VACUUM` fixed it.
- **Same adversarial run, after the fix:** `uv run pytest -q
  projects/pigrocrm/packages/core/tests -m "not slow"` -- 3077 passed, 0 failed
  (357s).
- **Break/restore, the acceptance criterion.** With `SET enable_seqscan = off` added
  temporarily inside the main test (mimicking what the falsifier does permanently), the
  identical test failed as expected:
  ```
  AssertionError: plan was:
  Bitmap Heap Scan on customers  (cost=30.69..34.70 rows=1 width=16)
    Recheck Cond: (...)
    ->  Bitmap Index Scan on ix_customers_ragione_sociale_trgm  (cost=0.00..30.69 rows=1 width=0)
          Index Cond: ((ragione_sociale)::text ~~* '%zzqqxxww%'::text)
  ```
  Reverted the temporary line, reran both tests: 2 passed.
- **The file alone, standalone and after the file's own bloat:**
  `test_trgm_escape.py` alone, 11 passed (70.6s); the two new tests alone (fresh
  container), 2 passed (20.9s).
- **`preflight --list`** selected `pigrocrm-py-lint`, `pigrocrm-py-test`,
  `pigrocrm-py-corpus`, `pigrocrm-py-planner` and `identity-names` for this diff.
  `preflight` (all five, unconditional lint/format included): all PASS --
  `pigrocrm-py-lint` 0.4s, `identity-names` 0.9s, `pigrocrm-py-test` 254.7s (the main
  `-n auto --dist loadfile -m "not slow and not planner"` gate, confirming the fix holds
  under xdist too), `pigrocrm-py-corpus` 702.2s, `pigrocrm-py-planner` 57.3s (now only
  `test_sort_plan.py`, unaffected by this change).
- `uv run mypy`: no issues (the change is test-only; mypy's configured source roots
  don't cover test files, so this is a repo-wide sanity check, not new coverage).

`customers_at_reference_scale` VACUUMs `customers` before inserting into it, which is
one step more than `test_search_plan.py`'s `inflated` fixture does (that one only
`VACUUM (ANALYZE)`s after building). I found this the hard way: a first version of the
fix that only vacuumed after committing reproduced the exact CI failure under the same
adversarial full-suite run. `inflated`'s 50 000 rows dwarf whatever bloat preceded it,
so the after-only vacuum is enough there; at REFERENCE scale (500 rows) leftover bloat
from a shared session-scoped container is not a rounding error next to the corpus
itself, so the pre-insert `VACUUM` is load-bearing here specifically. Worth a second
look in case there's a cleaner way to express that asymmetry.

## Deliberately not done

`test_sort_plan.py` (`pytestmark = [pytest.mark.slow, pytest.mark.planner]`) carries the
same class of machine-dependent plan assertions and its own file comment says
"ORB-9 covers the family." The actual ORB-9 ticket text names only
`test_the_planner_left_alone_still_prefers_a_sequential_scan_at_this_scale`, and my
assignment scoped me to `test_trgm_escape.py` under `packages/core`, so I left it alone
and filed ORB-90 as a follow-up rather than widening this PR's diff into a second file.

## Screenshots

Not applicable: this is a backend test-only change with no visible surface.

Linear: ORB-9.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -311,11 +311,13 @@ jobs:
       # `slow` builds a fifty-thousand-row corpus, so a PR does not pay for it and the
       # trunk does. `planner` is a different thing and is deselected on both: it asserts
       # which plan Postgres picks, the winning plan sits within a few percent of the
-      # loser at this row count, and a hosted runner picks the other one. It failed on
-      # the trunk run of the merge commit for exactly that reason, on code that had
-      # passed the same assertion on the devbox minutes earlier. It stays in preflight,
-      # which runs on one known machine, and ORB-9 tracks making the assertion say what
-      # it actually means instead of naming a plan.
+      # loser at this row count, and a hosted runner picks the other one.
+      # `test_trgm_escape.py`'s case (ORB-9) failed on the trunk run of the merge commit
+      # for exactly that reason, on code that had passed the same assertion on the devbox
+      # minutes earlier; its fix stopped costing the plan on whatever bloat the rest of
+      # the suite happened to leave behind, so it no longer carries the marker.
+      # `test_sort_plan.py` still does, for its own sort-plan assertions, and stays in
+      # preflight, which runs on one known machine.
       #
       # `slow` is not skipped on the trunk, it moved: the job below runs it in
       # parallel with this one. Measured 2026-09-09, the two are 10m and 12m of a
--- a/projects/pigrocrm/packages/core/tests/test_trgm_escape.py
+++ b/projects/pigrocrm/packages/core/tests/test_trgm_escape.py
@@ -71,18 +71,21 @@ change no semantics. It would then have had to go from **every** call site in th
 commit, never from one.
 """
 
+from collections.abc import Iterator
 from typing import Any
 
 import pytest
 from corpus import REFERENCE, build_corpus
-from sqlalchemy import ColumnElement, or_, select, text
+from sqlalchemy import ColumnElement, Engine, delete, or_, select, text
 from sqlalchemy.orm import InstrumentedAttribute, Session
 
 from pigrocrm.core.customers.models import Customer
-from pigrocrm.core.db import escape_like
+from pigrocrm.core.db import escape_like, session_factory
 from pigrocrm.core.deals.models import Deal
 from pigrocrm.core.documents.models import Document
+from pigrocrm.core.invoices.models import Invoice
 from pigrocrm.core.people.models import Person
+from pigrocrm.core.pipeline.models import PipelineStage
 
 # The nine indexes this slice declares. Named here rather than derived from the models,
 # so that deleting one from `__table_args__` fails this file instead of quietly agreeing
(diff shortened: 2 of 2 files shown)


fiorelorenzo:
Independent review, applied in 0816231:

- **Re-vacuum all five tables, not only `customers` (should-fix, applied).** The fixture commits and later deletes rows in `people`, `deals`, `documents` and `invoices` too, and the teardown was only re-vacuuming `customers`. Fixed: the teardown now loops `VACUUM (ANALYZE)` over all five, matching `inflated`'s own loop in `test_search_plan.py`.
- **Inverted opening sentence in the main test's docstring (should-fix, applied).** "no `Seq Scan` above sits with..." read as the negation of the assertion four lines below it. Reworded to "the `Seq Scan` the forced plans above never show".
- **Undocumented `pytest.mark.planner` removal, stale `ci.yml` comment (nit, applied).** Added a sentence to the test's own docstring on why the marker is gone, and narrowed `ci.yml`'s comment, which still framed ORB-9 as open and the assertion as still deselected on both CI tiers -- it's neither now, though `test_sort_plan.py` still carries the marker for its own out-of-scope assertions.
- **Assert `customers` is empty before the pre-insert `VACUUM` (nit, applied).** Turns a silent violation of the fixture's own premise into a message naming the cause instead of an unexplained flip of the `Seq Scan` assertion -- the exact symptom ORB-9 was filed over.

Reran after applying: `test_trgm_escape.py` alone, 11 passed (147s); `pigrocrm-py-lint` and `pigrocrm-py-test` equivalents both pass; `pigrocrm-py-corpus` (`-m "slow"`, whole `packages/core/tests`) 75 passed in 751.96s; `pigrocrm-py-planner` (`-m "planner"`) 51 passed in 64.64s, confirming `test_sort_plan.py`'s own planner-marked tests are unaffected and my two tests correctly no longer match that marker.

PR #15, opened 2026-09-10
fix(website): the cookie notice's controls clear the 44px touch-target floor

fiorelorenzo:
## What this changes

The cookie notice's three controls were under the 44px touch-target floor this
milestone names: "Dettagli" measured 20px tall, "No" was 38×29 and "Va bene" 68×29.
Now every one of them is at least 44px in its smaller dimension on a phone, and the
notice stays the same quiet tier it was: no border-width bump, no fill, no shadow.

"No" and "Va bene" get `min-width`/`min-height: 2.75rem`, the same shape ORB-64 used
for the header and footer controls; the existing padding is untouched, since the
`min-width`/`min-height` floor alone already carries both axes. "Dettagli" sits
mid-sentence in the notice's paragraph, so a real `min-height` on it would force the
line apart and read as loud; it gets an invisible `::before` hit-slop instead
(`inset: -14px -4px`), which grows the clickable area without painting anything or
shifting a pixel.

The focus-visible state needed nothing: landing.css and orbiters.css each already carry
one page-wide `::where(a, button):focus-visible` rule, and the notice's controls are
real `<a>`/`<button>` elements, so they were already covered. I checked this with real
keyboard Tab presses rather than assuming it from the CSS.

Both `index.html` and `orbiters.html` carry the notice (it lives in `system.css`,
shared by both sheets), so I rendered both.

**A review round found and fixed one real bug.** The link's `::before` hit-slop is a
positioned box, so it paints after the non-positioned buttons in the CSS stacking
order; whenever the notice wraps its paragraph onto its own line, the slop's edge
crosses into the buttons' row and a click in that sliver landed on "Dettagli" instead
of "No". `.consent-actions` now gets `position: relative` so the row paints after the
slop in tree order (no visual change, z-index stays `auto`), which I verified resolves
every overlap in a sweep of widths 280-390px on both pages. The review also caught that
my first pass had widened the buttons' padding unnecessarily (the `min-width`/
`min-height` floor alone was enough), which was moving the notice's wrap threshold and
is exactly what created the overlap band in the first place -- reverting it removes
both the unwanted layout change and the bug's precondition.

## How I verified it

- `pnpm --filter website test` -- 193 passed, including a describe block in
  `consent.test.ts` that pins the rule shapes (the 2.75rem floor on both axes for the
  buttons, the hit-slop shape and stacking-order fix on the link, no border/fill
  added, no source-regex standing in for a DOM check).
- `pnpm --filter website lint` -- clean.
- `preflight --list` selected `website-lint`, `website-unit`, `website-build`,
  `website-types`, `website-e2e`, `website-image`, `identity-names` for the two
  changed files; all seven passed (`website-e2e`: 49 passed, re-run after the fix too).
- Real-render measurement at 390px, both pages, via a headless Chrome tab
  (`getBoundingClientRect`): "No" 44×44, "Va bene" 47.3×44 (`/pigrocrm`: 69.2×44), the
  "Dettagli" hit area at or above 44px on its smaller dimension in every case.
- A geometric overlap-and-`elementFromPoint` sweep from 280px to 390px on both pages,
  before and after the stacking fix: every width where the link's phantom hit-slop
  geometrically overlaps a button now resolves the click to the button, not the link.
- `uishot --axe --fail-on serious` at 390 and 1440 on both `/` and `/pigrocrm`: zero
  violations on all four runs, both before and after the review fix.
- Real keyboard navigation (Puppeteer's `page.keyboard.press('Tab')`, not a synthetic
  `.focus()` call) through both pages: reached "Dettagli", "No" and "Va bene" in that
  order, each showing `outline: 3px solid` in the page's own focus colour.
- `pixel.test.ts` unchanged and passing: the pixel still only loads after "Va bene".

The `::before` hit-slop on "Dettagli" still overlaps the line of text above it (there
is no text below, since the link is always the paragraph's last line on both pages at
both viewports I rendered). That line carries no other control, so nothing is stolen
from it, but a pointer over that band shows the link's cursor and hover state over
plain prose, which is inherent to the technique. I could not find a case where
"Dettagli" wraps to a line with content after it or before it that has its own control;
if a future copy change adds one, worth re-checking with a render.

## Screenshots

**1. The notice at 390px, `/` (orbiters.html)**
Consent notice at 390px on the community page, before and after

**2. The notice at 1440px, `/` (orbiters.html)**
Consent notice at 1440px on the community page, before and after

**3. The notice at 390px, `/pigrocrm` (index.html)**
Consent notice at 390px on the PigroCRM page, before and after

**4. The notice at 1440px, `/pigrocrm` (index.html)**
Consent notice at 1440px on the PigroCRM page, before and after

The "before" frames are `origin/main` (`1a23646`), built in a second worktree; the
"after" frames are this branch's final state, same data, same viewport.

Linear: ORB-87.

### DIFF
--- a/projects/website/src/consent.test.ts
+++ b/projects/website/src/consent.test.ts
@@ -294,13 +294,31 @@ describe('the notice clears the 44px touch-target floor without shouting (ORB-87
     expect(before).toMatch(/inset:\s*-14px -4px/)
   })
 
+  it('paints the buttons after the link\'s hit-slop, so a corner click cannot land on the wrong control', () => {
+    // `.consent a::before` is positioned, so without this it paints in tree order
+    // after the row of buttons and can win a hit test in the sliver where the slop
+    // reaches over this row (measured: a real overlap and click-theft toward
+    // "Dettagli" at some notice widths before this rule, none after). z-index stays
+    // auto -- this only changes paint order among same-level positioned boxes, not
+    // anything a reader sees.
+    expect(rule('.consent-actions')).toMatch(/position:\s*relative/)
+  })
+
   it('keeps every consent control a real `a`/`button`, so the shared focus-visible rule already covers it', () => {
     // landing.css and orbiters.css each carry one `::where(a, button…):focus-visible`
     // rule with no scope narrower than the whole page (landing-style.test.ts and
     // orbiters.test.ts hold those). This notice earns a visible focus ring for free
     // as long as consent.js keeps building real anchors and buttons rather than a
-    // `div` with a click handler -- which is what this test actually pins.
-    expect(js).toMatch(/document\.createElement\('a'\)/)
-    expect(js).toMatch(/document\.createElement\('button'\)/)
+    // `div` with a click handler -- checked here on the DOM the same file already
+    // builds, not by grepping the source for the word `createElement`.
+    run().start()
+    const box = notice()
+    expect(box?.querySelector('a')).toBeInstanceOf(window.HTMLAnchorElement)
+    for (const label of ['No', 'Va bene']) {
+      const control = [...(box?.querySelectorAll('button') ?? [])].find(
+        (candidate) => candidate.textContent === label,
+      )
+      expect(control, `nessun bottone "${label}"`).toBeInstanceOf(window.HTMLButtonElement)
+    }
   })
 })
--- a/projects/website/src/consent.test.ts
+++ b/projects/website/src/consent.test.ts
@@ -247,3 +247,60 @@ describe('when the browser refuses to remember anything', () => {
     }
   })
 })
+
+describe('the notice clears the 44px touch-target floor without shouting (ORB-87)', () => {
+  // The behavioural tests above build the notice in jsdom, which lays out no CSS at
+  // all: a real render is what actually proves a pixel size, and the PR this test
+  // ships with carries one (uishot at 390 and 1440, before and after). What a unit
+  // test can hold is the rule shape, so a later edit that quietly drops the floor or
+  // swaps it for a heavier control fails here before it fails on a phone.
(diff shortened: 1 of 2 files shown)


fiorelorenzo:
Independent review (dispatched read-only, not myself) found one real bug and one avoidable layout side effect:

1. **Blocker, fixed.** `.consent a::before` (the link's hit-slop) is a positioned box and paints after the non-positioned buttons, so whenever the notice wraps its paragraph onto its own line, the slop's edge crossed into the buttons' row. A real click in that sliver landed on "Dettagli" instead of "No". Verified: `getBoundingClientRect` + `elementFromPoint` against the built dist, sweeping widths 280-390px on both pages, both before and after. Fix: `.consent-actions { position: relative; }`, which puts the buttons later in the stacking paint order with no visual change (z-index stays `auto`). Every overlap I could still find after the fix resolves to the button.
2. **Medium, fixed.** My first pass widened the buttons' padding (0.5rem 0.9rem), which was unnecessary since `min-width`/`min-height: 2.75rem` already carries both axes, and it was what shifted the notice's flex-wrap threshold into the collision band above. Reverted to the original padding.
3. **Nit, adopted.** Swapped a source-regex assertion (`consent.js` contains the string `document.createElement('a')` somewhere) for one built on the DOM the same test file already constructs, since the regex was satisfied by unrelated code in the file and wouldn't survive a harmless refactor.

Left as-is: the reviewer's suggestion to add e2e coverage in `e2e/site.spec.ts` — that file is outside my ownership for this issue (I own `system.css`'s `.consent` block and `consent.test.ts` only), so I added a unit test pinning the stacking-order rule instead and verified the fix by hand with the sweep above.

Commit `248418a`. Re-ran the full suite, lint, axe (both pages, both viewports) and the 49-test e2e suite after the fix; all green.

PR #17, opened 2026-09-10
fix(website): the site's own links stop bouncing through /orbiters

fiorelorenzo:
## What this changes

Nothing in the suite resolved an `<a href>` against what the site actually serves, and three internal links pointed at `/orbiters`, which nginx 301s to `/` since the community page took the front door on 2026-09-09: the landing's footer ("Orbiters", `index.html:237`), and two sentences in `privacy.html` ("joinorbiters.com" at `:44`, and the pixel-disclosure paragraph naming the page at `:87`). None of them 404, so nothing noticed; they still worked, just through a hop the site's own markup has no reason to take.

The footer link and the privacy.html domain mention now point straight at `/`. The pixel-disclosure paragraph named the wrong page for the second half of "the two pages an ad can land on": `/orbiters` is a redirect with no page left behind it, so the sentence now names and links `/pigrocrm`, the actual second page `pixel.test.ts`'s `MEASURED` list carries (`index.html`, served at `/pigrocrm`, and `orbiters.html`, served at `/`).

Adds `src/links.test.ts`, which walks every `<a href>` on the four built pages and resolves it against `path-map-plugin.ts`'s `route()`, the same function the dev and preview servers run on every request. A path the map 404s is a failure, and so is a link that only works by resolving through `REDIRECTS`, which is exactly what these three did: `REDIRECTS` exists for a bookmark or an inbound link this site does not control, never for markup this build produces itself. External links are checked for shape and host, never fetched; the file's header comment says why, including how it treats the GitHub link in `index.html` that happens to point back into this repository (the class of bug behind ORB-65's now-fixed dead `#installazione` anchor).

## How I verified it

`pnpm --filter website test`, before the fix, on the unmodified pages: exactly the three failures the issue names, all tripping the same rule:

```
FAIL  links.test.ts > index.html > links only to what the site actually serves
  - []
  + ["/orbiters is a redirect to /; link to / directly instead of through the redirect"]

FAIL  links.test.ts > privacy.html > links only to what the site actually serves
  - []
  + ["/orbiters is a redirect to /; link to / directly instead of through the redirect",
  +  "/orbiters is a redirect to /; link to / directly instead of through the redirect"]

Test Files  1 failed (1)
     Tests  2 failed | 4 passed (6)
```

After the fix, same command, same file: `Test Files 1 passed (1)`, `Tests 9 passed (9)` (the extra three cover the file-kind, query-string and non-mailto-scheme cases the independent review below caught).

Then the full suite, rebased onto `main` after both ORB-36 (#11) and ORB-88 (#13) merged: `pnpm --filter website test` (11 files, 199 tests, all green), `lint` (eslint, clean), `build` (vite), `exec tsc --noEmit` (clean), `test:e2e` (Playwright against `vite preview`, 49 passed), the `website-image` docker build (clean), and `orbiters-identity-scan .` (clean against 28 names, 4 paths allowed - added `links.test.ts` to the local allowlist, the same way `landing-pages.test.ts` already is, since both legitimately reference `humancraft.tech`). Everything `preflight --list` selected for this diff.

**External links, checked by hand** (curl from outside, 2026-09-09, recorded here as the baseline the issue asked for):

| Link | From | Answer |
| --- | --- | --- |
| `/orbiters` | `index.html:237`, `privacy.html:44` and `:87` | 301 to `/` (fixed above: all three now link to `/` or `/pigrocrm` directly) |
| `https://github.com/joinorbiters/orbiters/blob/main/projects/pigrocrm/README.md#deploy` | `index.html:213` | 200 (ORB-65's fix; the fragment itself is GitHub's own heading slug, not something this build can verify) |
| `/`, `/pigrocrm`, `/privacy`, `/termini`, `/hub/freelance`, `/hub/aziende`, `https://humancraft.tech/`, `https://pigro.joinorbiters.com/app/registrati` | various | 200 |
| `/app/` | both headers | 302 to `pigro.joinorbiters.com/app/` (intentional, ORB-64) |

`links.test.ts` treats any internal `<a>` that resolves through `REDIRECTS` as wrong, not merely as "working through an extra hop". That is a stricter reading than "resolves in `PAGES` or `REDIRECTS`" taken literally, and it is the rule that actually catches the three links this issue names (none of them 404; all three still 301 to a page that exists). The file's header comment explains the reasoning; a `REDIRECTS`-target-resolves check also runs on its own, independent of any page linking there, so a future legitimate redirect entry is still validated even though nothing should ever link to it directly.

An independent review (second commit) found and fixed three gaps in the check's own coverage before this went further: an absolute path `route()` calls a "file" was never checked against disk (a mistyped asset link would have passed), a query string was not dropped before routing the way the dev server's own middleware drops it, and a non-`mailto:` scheme (`tel:`, `javascript:`, ...) fell through to the relative-file branch with a misleading message. It also flagged that `orbiters.test.ts:111` (owned by ORB-88, not touched here) asserted `privacy.html` still contained `href="/orbiters"`; I had already flagged that to the agent working ORB-88 before making this change, and their PR #13 fixed the assertion. Both #11 and #13 are merged and this branch is rebased onto the result.

## Screenshots

One visible copy change: `privacy.html`'s pixel-disclosure paragraph names the second page as "PigroCRM" (linking `/pigrocrm`) instead of "Orbiters" (linking the now-removed `/orbiters`). Before/after at 1440px, cropped to the paragraph:

**1. privacy.html, the pixel-disclosure paragraph**
privacy.html paragraph naming the two pages an ad can land on, before and after

Nothing else in this diff is visible: the footer link and the domain mention keep their text (`Orbiters`, `joinorbiters.com`), only their `href` changes.

Linear: ORB-66.

### DIFF
--- a/projects/website/src/index.html
+++ b/projects/website/src/index.html
@@ -234,7 +234,7 @@
 
       <footer class="wrap rule" style="padding-block: 2rem">
         <p class="fine" style="display: flex; gap: 1.5rem; flex-wrap: wrap">
-          <a class="quiet-link" href="/orbiters">Orbiters</a>
+          <a class="quiet-link" href="/">Orbiters</a>
           <a class="quiet-link" href="/privacy">Privacy</a>
           <a class="quiet-link" href="/termini">Termini</a>
           <a class="quiet-link" href="https://example.com/">Studio Rossi</a>
--- a/projects/website/src/links.test.ts
+++ b/projects/website/src/links.test.ts
@@ -35,8 +35,12 @@ import { REDIRECTS, route } from './path-map-plugin'
  *   site does not control (`/orbiters`, from before the community page took `/`);
  *   markup this build produces itself should say what it means directly. That rule is
  *   what the three links this issue names actually trip: none of them 404, all three
- *   still 301 to a page that exists, and all three are "wrong" only in that sense --
+ *   still 301 to a page that exists, and all three are "wrong" only in that sense,
  *   which is exactly why nothing before this test noticed them.
+ * - A path under `ELSEWHERE` (`/app/`, `/hub/...`) resolves to another tenant of the
+ *   origin, named in `deploy/joinorbiters.conf` and served by a container this suite
+ *   never runs, so the check stops at "the path map hands it away" and does not, and
+ *   cannot, follow it further.
  */
 
 const PAGE_FILES = ['index.html', 'orbiters.html', 'privacy.html', 'termini.html'] as const
@@ -55,11 +59,11 @@ function idsOn(page: string): Set<string> {
 }
 
 /** The site's own trusted external destinations. Mirrors the allowlist
- *  `landing-pages.test.ts:42` already enforces for `href`/`src` subresources, kept as
- *  its own list here because that file does not cover `orbiters.html`, and because an
- *  `<a>` a visitor clicks is a different concern from a subresource the page fetches
- *  for itself: this list is free to diverge from that one without either test lying
- *  about what it guarantees. */
+ *  `landing-pages.test.ts`'s "requests nothing from another origin" test already
+ *  enforces for `href`/`src` subresources, kept as its own list here because that
+ *  file does not cover `orbiters.html`, and because an `<a>` a visitor clicks is a
+ *  different concern from a subresource the page fetches for itself: this list is
+ *  free to diverge from that one without either test lying about what it guarantees. */
 const EXTERNAL_HOSTS = ['github.com', 'pigro.joinorbiters.com', 'example.com', 'openai.com', 'humancraft.tech']
 
 function checkExternal(href: string): string | undefined {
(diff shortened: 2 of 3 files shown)


fiorelorenzo:
Independent review, applied in the second commit (`7ce933c` -> rebased to `8e31c70` on top of #11 and #13):

1. **Blocker, fixed by rebasing**: `orbiters.test.ts:111` broke against the old assertion. Already flagged with the ORB-88 agent before this PR opened; their #13 fixed the assertion, and this branch is now rebased on top of it.
2. **Medium, fixed**: `checkInternalPath` never checked that an absolute path `route()` classifies as `"file"` actually exists on disk, so a mistyped asset href would have passed. Added the check plus a test (mistyped vs real `orbiters-logo.svg`).
3. **Medium, fixed**: the `ELSEWHERE` paths (`/app/`, `/hub/...`) were passing without the header comment saying so. Added a bullet naming that gap explicitly.
4. **Low, fixed**: a query string was handed straight to `route()` instead of being dropped first, the way the dev server's own middleware drops it; a page's own campaign link to itself would have 404'd in the check while nginx serves it. Fixed, with a test.
5. **Low, fixed**: a non-`mailto:` scheme (`tel:`, `javascript:`, ...) fell through to the relative-file branch with a misleading message. Added an explicit skip, with a test.
6. **Low, fixed**: the comment citing `landing-pages.test.ts:42` pointed at the wrong assertion (a line-number reference that had already drifted). Now cites the test by name.
7. **Low, fixed**: the PR title and "How I verified it" called `/orbiters` a "dead redirect" and claimed the full suite was green while a known, disclosed dependency was still red. Retitled, and the body's suite counts now reflect the post-rebase, actually-green run.

Nothing left open from the review.

PR #22, opened 2026-09-10
fix(hub): each wizard page carries one visually hidden h1 naming it

fiorelorenzo:
## What this changes

`/hub/freelance` and `/hub/aziende` had no `<h1>` at all: the step title rendered as an `<h2>`, so `document.querySelector('h1')` returned `null` on a real build, which axe reports as `page-has-heading-one`. `Wizard.tsx` now renders a page-level `<h1>` from the `title` prop, visually hidden (`sr-only`) so the design stays exactly as it is: the step question stays an `<h2>` and so does the review screen's «Tutto giusto?». It renders outside the step/review conditional, so both states carry the same single `<h1>`, reading «Entra in Orbiters» on `/hub/freelance` and «Cerchi persone» on `/hub/aziende`, which is already what each page's `title` prop said. The chooser and the thanks page already had their own visible `<h1>` and are untouched.

## How I verified it

- `pnpm --filter hub test`: 27 passed (6 files), including the new `Wizard.test.tsx` assertion.
- Proved the new test is the regression guard: reverted the `Wizard.tsx` change with the test in place, `vitest run src/wizard/Wizard.test.tsx` failed (1 failed, 2 passed, `getAllByRole('heading', { level: 1 })` found none); restored the fix, same command passed 3/3.
- `pnpm --filter hub lint` and `pnpm --filter hub build` both clean; `preflight` selected `identity-names`, `hub-web-unit`, `hub-web-lint`, `hub-web-build`, all four passed.
- Real build, not a dev server: `pnpm --filter hub build` then `vite preview --port 34211 --strictPort`, driven with a headless browser.
  - `document.querySelectorAll('h1')` has length 1 on every public page, with the accessible name each should carry: `/hub/freelance` → «Entra in Orbiters» (`sr-only`), `/hub/aziende` → «Cerchi persone» (`sr-only`), `/hub/` → «Chi sei?», `/hub/grazie?chi=freelance` → «Grazie, sei dentro.», `/hub/grazie?chi=azienda` → «Grazie, ci siamo.».
  - Filled the `/hub/aziende` wizard through to the review screen: still exactly one `<h1>` («Cerchi persone», `sr-only`) alongside the review's own `<h2>` («Tutto giusto?»).
  - `uishot --axe --fail-on serious` at `390x844` and `1440x900`: clean (`"violations": []`, exit 0) on `/hub/freelance`, `/hub/aziende`, `/hub/`, and the thanks page in their initial state.

Running the same axe check against `/hub/aziende` once the review screen is reached (not required by this card, but it is the same file) surfaces a pre-existing `serious` `definition-list` violation unrelated to this fix: the review `<dl>`'s row `<div>` holds a `<button>` alongside its `<dt>`/`<dd>`, which axe does not allow as a direct child. It predates this change, is not part of ORB-89's scope («the single missing `<h1>`, not an accessibility pass»), and I filed it separately as ORB-96 rather than fold it in here.

## Screenshots

Nothing visible changed: the new `<h1>` is `sr-only` by design, so the two wizard pages render pixel-identical to `main`. Verified by the visual axe/DOM pass above rather than a screenshot pair.

Linear: ORB-89.

### DIFF
--- a/projects/hub/apps/web/src/wizard/Wizard.test.tsx
+++ b/projects/hub/apps/web/src/wizard/Wizard.test.tsx
@@ -81,6 +81,20 @@ describe('Wizard', () => {
     expect(onSubmit).toHaveBeenCalledTimes(1)
   })
 
+  it('carries exactly one level-one heading naming the wizard, on the step screen and on the review', async () => {
+    const user = userEvent.setup()
+    render(<Harness onSubmit={() => {}} />)
+
+    expect(screen.getAllByRole('heading', { level: 1 })).toHaveLength(1)
+    expect(screen.getByRole('heading', { level: 1, name: 'Test' })).toBeInTheDocument()
+
+    await user.type(screen.getByLabelText('Nome'), 'Ada{Enter}')
+    await user.type(screen.getByLabelText('Email'), 'ada@studio.it{Enter}')
+    expect(screen.getByRole('heading', { name: 'Tutto giusto?' })).toBeInTheDocument()
+    expect(screen.getAllByRole('heading', { level: 1 })).toHaveLength(1)
+    expect(screen.getByRole('heading', { level: 1, name: 'Test' })).toBeInTheDocument()
+  })
+
   it('goes back to the step a server error names', async () => {
     const user = userEvent.setup()
     const { rerender } = render(<Harness onSubmit={() => {}} />)
--- a/projects/hub/apps/web/src/wizard/Wizard.tsx
+++ b/projects/hub/apps/web/src/wizard/Wizard.tsx
@@ -105,6 +105,11 @@ export function Wizard<T>({
         next()
       }}
     >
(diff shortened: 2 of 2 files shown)


fiorelorenzo:
**Independent review.** One finding (P3, nit): the new `sr-only` h1 and the visible breadcrumb span right below it carried the same title text, so a screen reader announced it twice on `/hub/freelance` and `/hub/aziende`. Fixed in `441450b`: the breadcrumb span now carries `aria-hidden="true"`, so assistive tech hears the title once, from the heading. Nothing else flagged; the reviewer confirmed the h1 sits outside the step/review conditional (present in both states), the new test is a genuine regression guard (fails on the parent commit, passes here), and `sr-only` matches this repo's existing Tailwind v4 usage. Re-ran `vitest`, `eslint` and `vite build` after the fix: all clean.

PR #24, opened 2026-09-10
feat(hub): a freelancer gets back in with a link and changes what they sent

slavni96:
## What this changes

A freelancer who filled in the hub's wizard had nothing to come back to: no way in, no way to reread or correct what they sent, no place where the perks were. Now they have an area of their own at `/hub/io`.

- **A way in, without a password.** `/hub/accedi` takes the address they gave the wizard and answers «Se sei dentro, ti abbiamo scritto» whether the address is known or not. A one-time link (15 minutes, single use, sha256 at rest) arrives by mail through Resend and lands on `/hub/entra`, which posts the token and opens a 30-day sliding session in its own cookie, `orbiters_user`, never the admin's. Until the key is on the host, `/hub/accedi` answers 503 with a sentence instead of promising a mail that does not arrive.
- **What they sent, under the wizard's own questions.** `/hub/io` shows the seven answers, the CV as a download of their own file, the email as the address we write to (not editable: it is the identity the link proved), «Modifica» and «Esci».
- **Corrections.** `/hub/io/modifica` is the wizard's steps as one form, with the wizard's controls and rules; the CV is optional there. Every save writes a comment in the row's thread (ORB-59): «Profilo aggiornato dalla persona: tariffa giornaliera, link», signed with the person's name, so the admin sees what moved without an audit table. `stato` and `note` are never touched.
- **The perks.** «PigroCRM è tuo, gratis» in evidence with the link to `pigro.joinorbiters.com/app/registrati`, and «Altro in arrivo».
- **The doors.** «La tua area» in the hub header and one sentence on the freelancer's thank-you page.
- **The mail looks like the landing.** Paper ground, the four tiles, the white card with the 2px ink border and the 8px step, the Watermelon call to action, the bare URL below it for clients that block buttons, the fine footer. Tables and inline styles; the plain text stays beside it as the fallback.

Design record: `projects/hub/docs/superpowers/specs/2026-09-10-member-area-design.md`; the decision row in `docs/design/DECISIONS.md`.

## How I verified it

From the repository root, on `11ce72b7`:

```
uv run ruff check projects/hub && uv run ruff format --check projects/hub && uv run mypy   # clean
uv run pytest -q projects/hub          # 161 passed (PostgreSQL testcontainers, migrations to head)
pnpm --filter hub lint && pnpm --filter hub test && pnpm --filter hub build   # 41 vitest tests passed, build green
```

Then the whole flow by hand against a throwaway Postgres, the API on 8084 and the Vite server on 5180, with a fake Resend key so the link request answers 202: applied through the wizard as Ada, minted a link with `MemberService.request_link`, opened `/hub/entra?t=…`, landed on `/hub/io` with the seven answers and the CV, changed «Cosa fai?» to «Staff engineer» and saved, saw the new value on `/hub/io`, and in the database: `posizione = 'Staff engineer'`, one comment «Ada Lovelace | Profilo aggiornato dalla persona: posizione», one `member_sessions` row, the token marked used. The refused send (fake key) is a `WARNING the magic link mail was refused by the provider` in the API log, with no address in it. The HTML mail was sent for real through `ResendSender` to Ivan's address and rendered in the landing's system; Resend reported it delivered. The header holds on one line at 390px.

Not verified here: a real mail through Resend (no key yet) and the `Referrer-Policy` header, which lives in the hub image's nginx and is not run locally.

- **The HTTP seam now sends a `User-Agent` of its own.** Sending a mail for real returned `403 error code: 1010`: Cloudflare in front of api.resend.com refuses urllib's default `Python-urllib/3.x` signature before the request reaches Resend, and bzr.openai.com answers the same way, so the conversions pixel would have been refused too the day it gets a key (production has none today: `orbiters conversions-check` says «Nessun pixel configurato»). One line in `urllib_call`, pinned by `test_http.py`. Nothing else about the pixel changed.

- **`MemberService.enter` spends the token with `UPDATE … WHERE used_at IS NULL … RETURNING`**, not a check in Python followed by a write: two requests racing on one link (a mail scanner's prefetch and the person's click) open one session. There is no two-connection test for it; the argument is the statement.
- **`/hub/entra` follows the promise, not the observer.** Under React StrictMode, `@tanstack/query-core` 5.101.4 drops the mutation observer's callbacks when the effect double-invokes, and `mutate(…, { onSuccess })` never navigated (found by exercising the dev server; the API had answered 200 and set the cookie). The page now uses `mutateAsync(...).then(...)`, and a test mounts it inside `StrictMode`.
- **`request_link` is a timing oracle by a few milliseconds**: a known address writes a token and commits, an unknown one runs one SELECT. Status, body and headers are identical and the send happens after the response; the public 5/min bucket bounds the sampling. Documented in the module docstring rather than papered over.
- **`PUT /me/cv` reads the multipart body before the session dependency runs** (FastAPI parses the form first). It spends a rate-limit token and reads at most five megabytes and one byte; the host's nginx caps `/api/hub/` bodies at 6M.
- `FreelancerCreate` became `FreelancerFields` plus `email` and `utm`, so the member's `MemberUpdate` shares the wizard's validators; the wizard's own tests are the proof nothing changed.

- The website's community page does not yet say the area exists: ORB-19 removed «Accedi» from its footer at Ivan's request and ORB-71 is moving the front door. Filed separately.
- No cap on live links per address beyond the per-client rate limit; a per-address throttle is a spec gap, filed separately.
- No company area, no PigroCRM activation flag, no change to the wizards, no HTML mail.

## Migrations

`0005_member_area` (revises `0004`): creates `magic_link_tokens` (freelancer_id FK with `ON DELETE CASCADE`, `token_hash` unique, `expires_at`, nullable `used_at`) and `member_sessions` (same shape without `used_at`). Both tables are new, nothing is conditional, and `test_migrations.py`'s `compare_metadata` sees exactly the models. Safe on the production database: it touches no existing table.

## API changes

New routes under `/api/hub`, all Italian sentences on the error paths:

- `POST /auth/link` `{email}` → 202 `{"ok": true}` for any address (rate limited; the mail goes out in a background task); 503 «L'accesso via email non è ancora attivo» without `ORBITERS_RESEND_API_KEY`.
- `POST /auth/enter` `{token}` → 200 `MemberProfile` and the `orbiters_user` cookie (httpOnly, Secure, SameSite=Lax, Path=/, 30 days); 401 «Link non valido o scaduto. Chiedine un altro.»
- `GET /me`, `PATCH /me` (JSON `MemberUpdate`: the seven wizard fields), `PUT /me/cv` (multipart `cv`), `GET /me/cv` (attachment), `POST /me/logout` → 204. The id always comes from the session; there is no `/me/{id}`.

`MemberProfile` carries `id, nome, cognome, email, linkedin_url, cv_filename, cv_size, tariffa_giornaliera, posizione, remoto, links, created_at, updated_at` and never `stato`, `note`, UTM or the CV bytes. The hub's hand-written client `projects/hub/apps/web/src/lib/api.ts` gains `MemberProfile`, `MemberUpdate` and the `member` namespace. New settings: `ORBITERS_RESEND_API_KEY`, `ORBITERS_MAIL_FROM`, `ORBITERS_HUB_URL` (`.env.example`, compose, `projects/hub/AGENTS.md`).

## Screenshots

1440×900 unless noted, same data on both sides, the «before» from a worktree on `origin/main`.

**1. The freelancer's thank-you page says the area exists**
Thank-you page, before and after

**2. The hub header gains «La tua area»**
Hub header, before and after

**3. `/hub/accedi`, before and after the request (new page, no «before» on main; the two states side by side)**
Accedi, the two states

**4. `/hub/io`, the area (new page, full height)**
The member area

**5. `/hub/io/modifica`, the form (new page, full height)**
The edit form

**6. The header at 390px, on one line**
The hub at phone width

**7. The magic link mail, as a client renders the HTML (new, no «before»: the mail was plain text)**
The magic link mail

Linear: ORB-62.


slavni96:
Review record, before this PR was opened. Every task on the branch had its own review by a fresh agent, then a whole-branch review read the full diff against the spec. What they found and what I did with it:

- **Token spend could open two sessions from one link** (plan defect): the check in Python and the write were two statements. Fixed in `e52d9fd5` with a conditional `UPDATE … WHERE used_at IS NULL … RETURNING`, gated on the returned row. No two-connection test; the statement is the argument.
- **`PUT /me/cv` read an unbounded body before the session dependency ran** and spent no rate-limit token. Fixed in `c426d5d7`: `spend_one` first, at most five megabytes and one byte read; nginx's 6M cap on `/api/hub/` stays the real bound. Cookie flags and the logout's clearing were asserted nowhere; the tests now pin them.
- **A refused mail was invisible**: the provider's `False` was discarded. Fixed in `e642da3e`: the background task logs a WARNING with no address and no key, and a test proves both.
- **The `Content-Disposition` sanitiser lived in two routers.** Now one `cv_response` in `orbiters_api/downloads.py`.
- **`/hub/entra` never navigated under React StrictMode** (found by exercising the dev server, not by a reviewer): `@tanstack/query-core` 5.101.4 drops the mutation observer's callbacks when the effect double-invokes. The page follows `mutateAsync`'s promise now; a test mounts it inside `StrictMode`, and a 429 no longer reads as a dead link.
- **The edit form said nothing when only the CV was refused** after a successful PATCH: it now says «Le altre risposte sono salvate».
- **Header with three items and no `flex-wrap`**, an em dash in a README bullet, the guard's blank page on a non-401 error, the admin cookie never proven useless on the member routes, the spec claiming `enter` sweeps tokens: all fixed on the branch.

Left as they are, on purpose: `request_link` does more database work for a known address than for an unknown one (status, body and headers are identical, the send is deferred, the 5/min bucket bounds sampling; the module docstring says so); the comment on a profile change is a second transaction after the field commit (worst case a saved change with no thread entry); no per-address cap on live links beyond the per-client limit (a spec gap, filed as its own card); the website's community page does not yet say the area exists (its own card, since ORB-19 removed «Accedi» from that footer on request).

slavni96:
CI on the head `11ce72b7`: the push did not start a `pull_request` run (no check suite on the commit ten minutes later), so I dispatched `ci.yml` on the branch by hand: run 34489074733, every job green, the hub's python and web gates and the images included, since a dispatch run has no base commit and runs everything.

PR #25, opened 2026-09-10
feat(web): the invoice list says which customer each row belongs to

slavni96:
## What this changes

The Fatture list showed Numero, Tipo, Stato, Data, Totale and Pagamento, and nothing about whose invoice a row was: to learn the customer a person had to open each row. Ivan asked for it on 2026-09-10 («nella schermata delle fatture voglio vedere a quale cliente è associata»). The frontend could not fix that alone, because `InvoiceRead` carried only `customer_id`.

Now every `InvoiceRead` carries `customer_ragione_sociale`, the customer's name denormalised at read time the way `DealRead` already does it: `InvoiceRepository.customer_names` resolves every name of a page in one query, and `InvoiceService._read`/`_reads` apply it to every `InvoiceRead` the service hands back, so `get`, `list` and the mutation answers share one shape and the MCP tools (`list_invoices`, `get_invoice`) get the name with no change of their own. The list page shows a «Cliente» column right after Numero, with the same em dash every other empty cell shows when the name did not resolve.

The regeneration of `api-types.ts` is its own first commit: the committed file had fallen behind the API on `main` (it still typed `/api/orbiters/signups`, which the API no longer serves, and lacked the ORB-61 fields on `InvoiceRead`). ORB-101 tracks a gate for that drift.

## How I verified it

- Core, the two new tests first, red then green: `test_every_read_carries_the_customer_s_name` and `test_the_customer_name_costs_one_query_for_the_whole_page` (the latter counts the statements: the list plus exactly one lookup for three customers).
- `uv run pytest -q -n 3 --dist loadfile -m "not slow and not planner" projects/pigrocrm`: 4167 passed. (`-n auto` on this Mac trips the testcontainers reaper port race, 1084 setup errors and 0 failures; with three workers it is clean.)
- `uv run ruff check projects/pigrocrm`, `uv run ruff format --check projects/pigrocrm`, `uv run mypy`: clean, 279 source files.
- Web, red then green: `columns.test.ts` (opt-in column, position, accessor, dash for `null` and `''`, plain-text cell), `fatture/index.test.tsx` (the list renders the «Cliente» header and the name the API sent), `InvoicesTab.test.tsx` (the tab does not).
- `pnpm --filter web test`: 1057 passed, 101 files. `pnpm --filter web lint`, `pnpm --filter web build`, `tsc --noEmit`: clean.
- `pnpm --filter web generate:api` against the API on this branch: the only schema change on top of the regeneration commit is `customer_ragione_sociale?: string | null` on `InvoiceRead`.
- Opened http://localhost:5173/app/fatture against a local stack with three customers, two issued invoices and one proforma: the column reads «Bianchi Consulting S.r.l.», «Officina Verdi S.n.c.», «Studio Neri & Partner» on the right rows. `GET /api/invoices` returns the same names.

- `InvoiceRepository.customer_names` does not filter `deleted_at`, on purpose and more so than for deals: `CustomerService.soft_delete` blocks only on active deals, so an archived customer with issued invoices is a real state, and its name is still the name on those invoices. The docstring says so.
- Every mutation answer now costs one extra primary-key `SELECT` on `customers`, the same price `DealService` pays. `create` already holds the `Customer` row for the party snapshot, so a future micro-optimisation could pass the name through; not worth a signature change today.
- The cell is plain text, not `truncate`: the table is auto-layout and `TableCell` already says `whitespace-nowrap`, so a long name widens its column and the table scrolls inside its frame like every other text column. An earlier draft had `truncate` on an inline span, which does nothing there; the independent review caught it.

- The Fatture tab inside a customer's or a deal's page keeps its columns: every row there belongs to the customer named in the title. Recorded on ORB-98.
- No customer filter on the list page: the `customer_id` API filter exists and the ask was to see the name, not to filter by it.

## API changes

- `GET /api/invoices`, `GET /api/invoices/{id}` and every mutation that answers an `InvoiceRead`: the response gains `customer_ragione_sociale` (`string | null`, read-only, never accepted on write). No status code changes. `pnpm --filter web generate:api` was run with the API up on `localhost:8000`; `projects/pigrocrm/apps/web/src/lib/api-types.ts` matches.

## Screenshots

**1. Fatture list, the «Cliente» column after Numero**

Same data on both frames: the before is the `origin/main` web on 5175, the after this branch on 5173, both against this branch's API, so the before shows a state `main` never had only in the sense that its API already knew the name; the page did not.

Linear: ORB-98.

### DIFF
--- a/projects/pigrocrm/apps/web/src/features/invoices/InvoicesTab.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/invoices/InvoicesTab.test.tsx
@@ -33,6 +33,15 @@ describe('InvoicesTab', () => {
     expect(screen.queryByRole('button')).not.toBeInTheDocument()
   })
 
+  /** Every row on this tab belongs to the customer (or the deal's customer) named in
+   *  the page title above it, so the «Cliente» column the list page shows (ORB-98) would
+   *  only repeat that title on each line. */
+  it('does not repeat the customer on every row', async () => {
+    renderWithClient(<InvoicesTab owner={{ customerId: 'c1' }} />)
+    expect(await screen.findByRole('columnheader', { name: 'Numero' })).toBeInTheDocument()
+    expect(screen.queryByRole('columnheader', { name: 'Cliente' })).not.toBeInTheDocument()
+  })
+
   /**
    * The slot exists so the button that creates an invoice sits on the tab that lists
    * them, rather than in the page header where it would be offered from every other tab
--- a/projects/pigrocrm/apps/web/src/features/invoices/columns.test.ts
+++ b/projects/pigrocrm/apps/web/src/features/invoices/columns.test.ts
@@ -1,7 +1,7 @@
 import { render, screen } from '@testing-library/react'
 import type { ReactElement } from 'react'
 import { describe, expect, it } from 'vitest'
-import { buildInvoiceColumns } from './columns'
+import { buildInvoiceColumns, type InvoiceColumnOptions } from './columns'
 import type { Invoice } from './queries'
 
 const ISSUED = {
@@ -26,8 +26,10 @@ const PROFORMA = {
   totale: '500.00',
 } as unknown as Invoice
 
-function accessor(id: string, row: Invoice): unknown {
-  const column = buildInvoiceColumns().find((candidate) => candidate.id === id)
+const WITH_CUSTOMER = { ...ISSUED, customer_ragione_sociale: 'ACME S.r.l.' } as Invoice
+
+function accessor(id: string, row: Invoice, options?: InvoiceColumnOptions): unknown {
+  const column = buildInvoiceColumns(options).find((candidate) => candidate.id === id)
   if (column === undefined || !('accessorFn' in column) || column.accessorFn === undefined) {
     throw new Error(`nessuna colonna con id ${id}`)
   }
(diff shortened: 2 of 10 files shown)


slavni96:
**Independent review, before the PR opened.** A fresh read-only reviewer went over the diff of the first version of the feature commit (columns, tests, service, repository, the regenerated types) with the deals precedent and `AGENTS.md` as context. Findings and what I did with each:

1. `truncate` on the name span was a no-op (inline span in an auto-layout `<td>`, no width). Fixed: the cell is plain text like «Tipo», with a comment saying why; the muted dash branch stays.
2. The new `InvoicesTab` test sat between an existing docblock and the test it documents. Fixed: moved above it.
3. The regenerated `api-types.ts` folded in unrelated drift (the removed signups route, the ORB-61 fields). Fixed: split into its own `chore(web)` commit ahead of the feature, and ORB-101 filed for a freshness gate.
4. One extra primary-key `SELECT` on every mutation answer. Left as is: the same price `DealService` pays, noted in the body under "look at twice".
5. Repeated `find`/`throw` boilerplate in `columns.test.ts`. Fixed: the existing `accessor`/`column`/`renderCell` helpers take an options argument, and a `renderCell` assertion now covers the dash branch of the cell.
6. `customerName` treated `''` as empty with no test and no comment. Kept, because `ragione_sociale` is a `SafeStr` with no minimum length; the comment says so and the dash test asserts it.

Checked and found correct by the reviewer, no action: every `InvoiceRead` construction goes through `_reads` (including `import_issued` via `get`); the no-`deleted_at` reasoning in `customer_names`; the query-counting test's listener scope under xdist; `schema_registry.native_fields` unaffected; MCP output gains the key through `model_dump`.

The fixes are folded into the feature commit rather than a second commit because the review ran before the PR existed; the reviewer's verdict on the original was "merge after fixes", and every fix it asked for is in.

PR #26, opened 2026-09-10
feat(core): the invoice PDF follows the layout of the register it inherited

slavni96:
## What this changes

Ivan's issued invoices are the previous system's own PDFs, kept at import through `pdf_sorgente`; a proforma rendered by PigroCRM sat beside them in the same list looking like another product («la proforma non ha il template uguale a quella delle fatture», 2026-09-10). The two PigroCRM templates were identical to each other and foreign to both, so a fattura issued natively would have looked just as foreign.

The invoice PDF now follows the inherited register's layout, for the fattura and the proforma alike: the emitter's whole fiscal identity top right («Identificativo fiscale ai fini IVA: IT…», codice fiscale, «Regime fiscale: RF19 Regime forfettario», codice destinatario, address, email), a «TD01 fattura | Numero: 2026/7 | Data: 11-08-2026» line, the customer on two bar-separated lines, a three-column detail table (DESCRIZIONE, %IVA with the Natura when there is no rate, PREZZO TOTALE), the payment code with its words («MP05», «Bonifico IBAN …»), a rule, and a «Thank you!» with the contacts. Amounts read «1.760,00 €», dates `dd-mm-yyyy`, the accrual period «dal 01-08-2026 al 31-08-2026». The footer carries the page number and nothing else.

One template (`template-invoice.md`) serves both documents, the proforma declaration under `{{#if}}`; `template-proforma.md` is gone, so the two cannot drift apart again. The proforma's first line reads «Fattura proforma | Numero: PROV-…» with no TD code, because it is not a fiscal document. The invoice has a page header of its own (`header-invoice.typ.template`); offers and time reports keep `header.typ.template` untouched.

Four departures from the register, decided with Ivan and recorded in `docs/design/DECISIONS.md`: the regime and stamp-duty declarations stay, small, under the rule (required on an invoice, the register did not print them); the Imponibile/Imposta/Totale block prints only when there is VAT or a stamp, and the stamp is stated by its declaration under the rule, never as a row (the total does not include it and the emitter bears it), so a forfettario document matches the register and an ordinario one still shows its tax; the number stays `anno/numero`; the VAT id prefix is the party's `nazione`, as `IdPaese` is in the XML, where the register printed `IT` for a London customer.

## How I verified it

- Red then green: `tests/test_invoice_pdf.py`, nine scope tests (amount and date formats, the first line for a fattura and a proforma, the two customer lines and their dropped segments, the emitter block, Natura versus rate, the payment row, when the totals block appears) and three rendered-text tests through Pandoc, Typst and `pdftotext` (the fattura reads like the register and carries no declaration and no totals block; the proforma is the same document with its declaration and no `TD01`; an ordinario invoice shows Imponibile, Imposta 44,00 €, Totale documento). All twelve failed on `origin/main` and pass now.
- `tests/test_invoice_artifacts.py`: the three assertions on the printed dates moved to the new forms; the address tests (ORB-55) pass unchanged.
- `uv run pytest -q -n 3 --dist loadfile -m "not slow and not planner" projects/pigrocrm`: 4178 passed, 1 failed on the first run. The failure was `test_no_unapproved_identifier_appears_anywhere`: my test data carried a codice destinatario copied from one of the register's PDFs, a real client's. Replaced with allowlisted synthetic codes, the guard passes (15 passed with the PDF tests), and the branch was force-pushed so the pushed history does not carry it. The full run after the fix: 4179 passed in 2m05s.
- `uv run ruff check projects/pigrocrm`, `uv run ruff format --check`, `uv run mypy`: clean.
- After the independent review (recorded as a comment on this PR): `test_invoice_pdf.py` and the identity guard 15 passed; `test_invoice_artifacts.py`, `test_render_pdf.py`, `test_time_report_pdf.py`, `test_documents_from_template.py` 54 passed; ruff, mypy clean. The pairs below are rendered from the reviewed commit.
- Rendered a fattura with two lines, a stamp and a period, a proforma on the same data, and an RF01 invoice with 22% VAT, all with synthetic parties, and read the three PDFs page by page against one of the register's documents: same sections, same order, same formats.
- Not run locally: the `docker compose build` check `preflight.json` lists for a core change; the trunk tier builds the images after merge. Nothing under `apps/web`, `apps/api` or `apps/mcp` changed.

- `header-invoice.typ.template` sets `margin.top: 2.5in` and `header-ascent: 25%` so the seven-line emitter block fits above the body without overlapping it. Verified on the three rendered pages, not on a page with a very long ragione sociale that wraps.
- The totals block prints for VAT alone. The first version brought it back for a stamp too, which on a forfettario invoice above 77,47 EUR is nearly always, with rows that did not sum; the independent review caught it.
- `format_iva` prints the Natura when a line has one, otherwise the rate through `Decimal.normalize()` with a comma (`22%`, `4,5%`). FPR12 does not admit a zero rate without a Natura, so that branch prints `0%` rather than guessing.
- `_for_export_proforma` passes `data_scadenza=None`, so the SCADENZE cell of a proforma is empty, as it was before this change.
- The scope no longer carries `quantita`, `prezzo_unitario`, `unita_misura` or `aliquota_iva` per line: nothing else read them, and a template that reads a missing key fails the render rather than printing blank.

- Offers and time reports: their header, footer and templates are the ones they had. Only the invoice reads the new header.
- The XML: `fatturapa.py` is untouched, and `totals.format_amount_2` stays the FPR12 formatter beside the new display one.
- The previous system's `17/00` numbering is not reproduced (`naming.numero_completo`).
- The invoice detail page and its PDF preview in the web: they show whatever the renderer produces.

## Screenshots

Synthetic parties on both frames («Studio Rossi», «Acme S.r.l.»), the same `InvoiceForExport` rendered by `origin/main` on the left and by this branch on the right, rasterised at 100 dpi. No document from Ivan's register appears here.

**1. The fattura**
Fattura PDF, before and after

**2. The proforma, with its declaration**
Proforma PDF, before and after

Linear: ORB-102.


slavni96:
**Independent review.** A fresh read-only reviewer went over `91aa55d` with the render pipeline, the template parser and escaping, the XML writer and `AGENTS.md` as context. Findings and what I did with each, in `8919168`:

1. The totals block came back for a stamp as well as for VAT. On a forfettario invoice above 77,47 EUR the stamp is nearly always there, so the block the register never printed would have appeared on nearly every real invoice, and its rows did not sum (the total excludes the stamp the emitter bears). Fixed: VAT alone brings it back, the stamp row is gone, the declaration under the rule states it; the register-match test now renders with the stamp and asserts no `Imponibile`.
2. `identificativo_iva` printed no prefix for an empty `nazione` while `fatturapa.py` writes `IT`. Fixed: the same fallback, with a test.
3. The header printed `Identificativo fiscale ai fini IVA:` and `Indirizzo:` in front of possibly empty values. Fixed: both under `{{#if}}` like their neighbours.
4. `mostra_totali` is a `bool` in a scope the docstring says holds only strings. Kept, and the docstring names the exception: a flag only `{{#if}}` reads.
5. The new skip reason was in Italian. Fixed.
6. `f"{tipo_documento} fattura"` hard-wired the word. Fixed: `TIPI_DOCUMENTO`, code alone as fallback, like `REGIMI`.
7. `_bars_space` sat after its only caller. Moved beside `_bars`.

Checked and found correct by the reviewer, no action: `{{#if}}` bodies inside `#stack`/`#table` argument lists compile, also when every optional line is absent; every placeholder is in markup position and hostile customer values render verbatim; two renders of one export are byte-identical; the compiled markdown is discarded by `produce_artifacts`, so the dropped per-line keys have no other reader; no remaining reference to the deleted template outside the grandfathered design records.

The before/after pairs in the body are rendered from `8919168`.

PR #28, opened 2026-09-10
feat(website): the landing's footer opens a door into the hub's admin area

slavni96:
## What this changes

Nothing on the public site pointed at the hub's admin area, so whoever reviews signups, freelancers and companies had to remember `/hub/admin` and type it. The landing's footer (`/pigrocrm` today, the front door once ORB-71 lands) now ends with a quiet «Admin» link to `/hub/admin/freelance`, dressed like Orbiters, Privacy, Termini and the studio link beside it. It goes to the freelancers list, the first screen of the area and where the hub's own login lands, rather than to `/hub/admin`: the hub router has no index route under `/admin`, so a signed-in admin sent there would see an empty panel (found in review, filed as ORB-106). The hub's `AdminLayout` still sends a visitor without a session to `/hub/admin/login` first.

## How I verified it

- `pnpm --filter website test`: 204 passed, 11 files. The new test in `landing-pages.test.ts` failed before the markup change (the footer had no admin link) and passes after; it pins the anchor by href and text inside `<footer>`, last in its paragraph, and asserts the admin path appears once on the page.
- `pnpm --filter website lint`: clean. `pnpm --filter website build`: built in 516ms.
- `pnpm --filter website test:e2e`: 49 passed. The JavaScript-disabled link walk skips `/hub/` paths by design (another tenant of the origin), and `links.test.ts` resolves `/hub/admin/freelance` through `ELSEWHERE`, so nothing had to be added to a path map.
- After the review commit: 204 unit tests, lint, build and the 49 e2e tests again green.
- Opened `vite preview` on 4173 (this branch) and 4175 (`origin/main`), 1440x900, scrolled to the end: the link renders as the same bordered control as its neighbours, 70x44 CSS pixels, so it clears the 44px touch-target floor.

The community page (`orbiters.html`, today's `/`) stays without a footer: ORB-19 removed it on Ivan's request on 2026-09-09 and `orbiters.test.ts` keeps it from coming back, and ORB-72 retires the page once the landing takes the front door. The privacy and terms footers and the hub's own `Shell` footer are left alone too: one door into the admin area, on the page that is becoming the home. Recorded on ORB-105.

Whether an admin link belongs on a public page at all. It is discreet and leads to a guarded area, and Ivan asked for it in the footer, but it does tell a visitor that an admin area exists. The footer's fourth link, «Studio Rossi» to `example.com`, is a fixture that leaked onto the page (ORB-97 names the same on the hub); I left it as I found it.

## Screenshots

**1. The landing's footer at 1440x900, scrolled to the end**
The landing footer, before and after

Linear: ORB-105.

### DIFF
--- a/projects/website/src/index.html
+++ b/projects/website/src/index.html
@@ -238,6 +238,7 @@
           <a class="quiet-link" href="/privacy">Privacy</a>
           <a class="quiet-link" href="/termini">Termini</a>
           <a class="quiet-link" href="https://example.com/">Studio Rossi</a>
+          <a class="quiet-link" href="/hub/admin">Admin</a>
         </p>
       </footer>
     </main>
--- a/projects/website/src/index.html
+++ b/projects/website/src/index.html
@@ -238,7 +238,7 @@
           <a class="quiet-link" href="/privacy">Privacy</a>
           <a class="quiet-link" href="/termini">Termini</a>
           <a class="quiet-link" href="https://example.com/">Studio Rossi</a>
-          <a class="quiet-link" href="/hub/admin">Admin</a>
+          <a class="quiet-link" href="/hub/admin/freelance">Admin</a>
         </p>
       </footer>
     </main>
--- a/projects/website/src/landing-pages.test.ts
+++ b/projects/website/src/landing-pages.test.ts
@@ -186,6 +186,16 @@ describe('index.html', () => {
     expect(page).toMatch(/href="\/termini"/)
   })
 
+  it('opens a door into the hub admin area from its footer, and only there', () => {
+    // Ivan, 2026-09-10 (ORB-105): whoever reviews signups, freelancers and companies
+    // should not have to type `/hub/admin` by hand. One quiet link at the end of the
+    // footer, dressed like its neighbours; the hub's `AdminLayout` sends a visitor
+    // without a session to its login, so the page links the area and not the login.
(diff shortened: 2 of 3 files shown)


slavni96:
## Independent review, applied in 46f6e22f

A fresh read-only reviewer went over the diff, ran the website suite (204 unit, lint clean) and exercised the new test against six mutations of the page (link removed, moved, duplicated, outside the footer, trailing slash, attribute order): every mutation fails it.

**Should fix, applied.** `/hub/admin` has no screen behind it for a signed-in admin: `router.tsx` declares `adminArea` with three children and no index route, and `AdminLayout` redirects only without a session, so the person the link is for would land on the sidebar with an empty panel. The link now goes to `/hub/admin/freelance`, where the hub's own login lands. The hub-side fix (an index redirect under `/admin`) is filed as ORB-106. Inferred from the route tree, not from a signed-in render.

**Nits, applied.** The test matches the anchor by href and text instead of pinning attribute order. The `landing.css` comment said "four-link footer"; it now says "flex footer".

**Raised and left as is, with the reason.** `rel="nofollow"`: no anchor on the four pages carries a `rel`, and the hub's own `index.html` is `noindex`, so the destination handles crawlers itself. Disclosing that an admin area exists from a public page: the area is cookie-guarded and its login URL is trivially discoverable, so it is not a reason to hold the merge; Ivan asked for the link in the footer. The «Studio Rossi» fixture beside the new link is pre-existing and is ORB-97's.

PR #32, opened 2026-09-10
fix(website): the landing's footer links the studio again, not a fixture

slavni96:
## What this changes

`website-v0.4.0` put a test fixture on the public landing: the footer's fourth link went out as «Studio Rossi» to `example.com`, where production had said «Humancraft» to `humancraft.tech` until that deploy. The pre-publication sanitisation had swapped the link on `main` and nothing caught it, because `example.com` sat in both link allowlists. The link is back to what the site said, `example.com` is out of both allowlists so a fixture host fails the suite, and a test pins the studio link in the footer. This is a restore, not new copy: `privacy.html` and `termini.html` on `main` name the same studio as titolare del trattamento, and CI is green with them.

## How I verified it

- `pnpm --filter website test`: 205 passed, 11 files. With the tests changed and the markup not yet, three failed (the new footer test, the origin allowlist in `landing-pages.test.ts`, and `links.test.ts` on `example.com`); all pass after the one-line restore.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_no_real_identity_in_the_repository.py`: 4 passed. The identity guard checks fiscal shapes and email domains; a link to the studio's site on the landing is not what it bans, and the two policy pages already carry the same host.
- `pnpm --filter website lint`: clean. `build`: ok. `test:e2e`: 49 passed.
- Production before and after `website-v0.4.0`, read with `curl` on `https://joinorbiters.com/pigrocrm`: `href="https://humancraft.tech/">Humancraft` before, `href="https://example.com/">Studio Rossi` after. This PR is the fix for that.

The hub's footer sentence «Orbiters è un progetto di Studio Rossi» (`projects/hub/apps/web/src/components/Shell.tsx:52`) is ORB-97's: that one names the entity behind the service in a sentence, and its wording is Lorenzo's and Ivan's to choose. The website's link had a known previous value and only that is restored.

Whether restoring the studio's name on the landing is in tension with the repository being public. My reading: the two policy pages already carry the studio's name, VAT number and email by legal requirement, `landing-pages.test.ts` says so in its own comment, and the identity guard does not flag the host. If the intent was to keep the studio off the landing specifically, the fix would be to drop the link rather than to link a fixture, and that is a decision to record, not something this PR takes.

## Screenshots

**1. The landing's footer at 1440x900, scrolled to the end**
The landing footer, before and after

Linear: ORB-116.

### DIFF
--- a/projects/website/src/index.html
+++ b/projects/website/src/index.html
@@ -237,7 +237,7 @@
           <a class="quiet-link" href="/">Orbiters</a>
           <a class="quiet-link" href="/privacy">Privacy</a>
           <a class="quiet-link" href="/termini">Termini</a>
-          <a class="quiet-link" href="https://example.com/">Studio Rossi</a>
+          <a class="quiet-link" href="https://humancraft.tech/">Humancraft</a>
           <a class="quiet-link" href="/hub/admin/freelance">Admin</a>
         </p>
       </footer>
--- a/projects/website/src/landing-pages.test.ts
+++ b/projects/website/src/landing-pages.test.ts
@@ -57,7 +57,7 @@ describe.each(PAGES)('%s', (name) => {
       // repository, and `bin/identity-scan` knows about these three paths for the same
       // reason.
       expect(url, 'external subresource').toMatch(
-        /^https:\/\/(?:github\.com|pigro\.joinorbiters\.com|example\.com|openai\.com|humancraft\.tech)\//,
+        /^https:\/\/(?:github\.com|pigro\.joinorbiters\.com|openai\.com|humancraft\.tech)\//,
       )
     }
     expect(page).not.toMatch(/fonts\.googleapis\.com|fonts\.gstatic\.com/)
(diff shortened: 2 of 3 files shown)


slavni96:
## Independent review, applied in fb287345

A fresh read-only reviewer ran the website suite (205 unit, lint, build), the identity guard (4 passed), traced the regression to `622d5075` (the sanitisation sweep rewrote the footer line together with the fixtures and added `example.com` to the origin regex in the same commit), and checked the built output: after this PR no page carries `Studio Rossi` or `example.com`; `Humancraft` appears once on the landing and on the two policy pages that already named it.

**Should fix, applied.** The comment above the origin regex in `landing-pages.test.ts` said `humancraft.tech` was allowed for the two policy pages alone and pointed at `bin/identity-scan`, which does not exist. It now says the landing's footer carries the host too since ORB-116, and names `orbiters-identity-scan` as `.github/preflight.json` runs it.

**Should check, could not.** Whether `orbiters-identity-scan .` on a machine with the forbidden-names list flags `index.html:240` now that a third page carries the host. The scanner is not installed on this Mac (`~/.config/orbiters` does not exist here), so I could not run it; CI does not run it either. If it flags the line on a configured machine, the fix is to add the landing to the scanner's path exemptions outside the repository.

**Nits.** The `example.com` note in `links.test.ts` is folded into the JSDoc above the list. The commit body not being in the first person I left as is.

**Raised, left as is.** The sample testimonial «Mario Rossi ha lavorato per XYZ» on the landing is the same sweep's work, but it is illustrative copy in a placeholder section (ORB-67 owns the testimonials), not an attribution; noted there rather than changed here.

PR #35, opened 2026-09-10
style(hub): the public pages sit straight on the grid, without a panel around them

slavni96:
## What this changes

The chooser at `/hub/`, the two wizards and the thank-you page all sat inside a white panel with the site's thick border and hard shadow, and the two door cards inside the chooser drew the same border and shadow again: a frame around two frames, which read as too heavy at 1440px. I drop the panel from `Shell` and let the content sit straight on the grid, the way the landing does. The header, the footer and the centring of the content between them do not move. The door cards, the wizard fields and the review list keep their own boxes and are the frame now.

## How I verified it

In `projects/hub/apps/web`: `pnpm lint` clean, `tsc --noEmit` clean, `vitest run` 12 files, 44 tests passed. I ran the hub dev server from this branch on 5180 and from a worktree on `origin/main` on 5182, opened `/hub/` in both at 1440×900 and compared them: only the panel is gone, the heading and the cards keep the same position (the `h1` sits at the same coordinates in both frames).

The 2026-09-09 hub spec still says the web keeps "the same header/panel shapes" as the CRM. I left the spec alone: the panel decision was already superseded by ORB-73, which moved the wizards into the site's own visual system, and this PR only removes the last piece of it. If you want that sentence rewritten I would do it in the same PR.

The copy of the chooser now sits directly on the grid, with no surface behind it, unlike the header and footer links which keep their sliver of surface. At 1440px it reads fine to my eye. Whether the wizard steps want a surface behind their labels is a second look and is noted on the issue.

The door cards, the wizard fields, the review list and the tokens. The decision is recorded on ORB-124.

## Screenshots

**1. The chooser at 1440×900, panel gone, cards and heading where they were**
The chooser before and after

Linear: ORB-124.

### DIFF
--- a/projects/hub/apps/web/src/components/Shell.tsx
+++ b/projects/hub/apps/web/src/components/Shell.tsx
@@ -3,15 +3,18 @@ import type { ReactNode } from 'react'
 import { BrandMark } from '@/components/BrandMark'
 
 /** The public frame: the mark and the name up top, the two legal links and the
- *  attribution at the foot, one boxed panel in between -- the site's own visual
- *  system (ORB-73's `.site` scope) rather than the application's, so a visitor who
- *  clicked a CTA on joinorbiters.com does not land on a different product. `site`
- *  also carries the page's ground (the same faint grid the landing sits on), and the
- *  panel is centred in the space between header and footer instead of sitting at the
- *  top of an empty page: `justify-center` on `main` only has room to act when the
- *  step is shorter than the viewport, which is the common case here. Each link over
- *  the grid keeps a sliver of the page's own surface behind it, the same treatment
- *  `landing.css`'s `.top a` and `footer .quiet-link` give theirs. */
+ *  attribution at the foot, and the page content straight on the grid in between --
+ *  the site's own visual system (ORB-73's `.site` scope) rather than the
+ *  application's, so a visitor who clicked a CTA on joinorbiters.com does not land on
+ *  a different product. `site` also carries the page's ground (the same faint grid
+ *  the landing sits on). There is no panel around the content: the doors of the
+ *  chooser and the wizard fields draw their own boxes, and a box around those boxes
+ *  read as one frame too many. The content is centred in the space between header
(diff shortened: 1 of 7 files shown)


slavni96:
## Independent review

A fresh read-only reviewer went over the diff with the pages Shell wraps, the `.site` stylesheet and the tests. Findings and what I did with each:

1. **Alerts under AA contrast on the grid** (`text-destructive` measures 4.17:1 on paper, 4.67:1 on the panel's white it was tuned for). Fixed: each `role="alert"` in `Wizard.tsx`, `Accedi.tsx`, `Modifica.tsx`, `Guard.tsx` now sits on a sliver of `bg-card`, the header links' treatment.
2. **Inputs and textareas transparent, grid running through them.** Fixed in `site-controls.css`: `.site` inputs paint `var(--card)`, with a comment pointing at the landing's `.signup input` precedent.
3. **Stale comment in `button.tsx`** saying outline buttons sit inside the content panel. Fixed: the comment now says the button paints its own surface and reads the same in the admin's panel and on the grid.
4. **Progress bar track transparent.** Fixed: `bg-card` on the track.
5. **Shell doc comment**: name every page that draws its own cards, and say why the bare `w-full` wrapper stays (`main` is `items-center`, the chooser has `max-w-2xl` without `w-full`). Fixed.
6. **Spec sentence "the same header/panel shapes"**: the reviewer's advice is to leave the dated spec alone, since ORB-73's row in `DECISIONS.md` already supersedes it for the public pages. Left as is.
7. **`Area.tsx` "Prossimamente" card is `bg-muted/40`**, so the grid shows through at 60%. Left as is: it is a dashed coming-soon card and reads as intended.

Checked clean by the reviewer: no test or e2e selector named the panel; at 390px the content gains room rather than losing it; muted copy is 7.01:1 on paper.

Second commit verified the same way as the first: lint, `tsc`, Vitest 44/44, and the company wizard's first step opened at 1440×900 from this branch: the field's computed background is `rgb(255, 255, 255)`.

PR #36, opened 2026-09-10
feat(hub): an admin creates the next admin from the area, and sees who is there

slavni96:
## What this changes

Until now the only way to add a reader of the hub's admin area was ssh and `orbiters createadmin` on the server, and nothing in the area said who the admins are. The sidebar gains a fourth entry, «Amministratori», at `/hub/admin/amministratori`: the list of admins (nome, email, «Attivo» or «Disattivato», since when) and a form with nome, email and password, «Crea amministratore». On success the list grows by one row, the form empties and a line says «Amministratore creato: ada@orbiters.it. Ora può accedere con la password che gli hai dato.» A duplicate address or a short password comes back as a 422 naming the field, and the form marks that field and shows the sentence.

Behind it, `GET /api/hub/admins` and `POST /api/hub/admins`, both under the same cookie as everything else in the area. The POST reuses `AdminService.create`, so the rules are the CLI's (one address, ten characters, argon2). `AdminRead` now carries `attivo` and `created_at`; the hash never leaves the service.

Ivan chose, on 2026-09-10, that the creating admin picks the password and hands it over out of band, as the CLI does; an invitation by email with a set-password link was the other option and is a possible follow-up (recorded on ORB-123).

## How I verified it

- `uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests`: 164 passed. The four new tests failed before the implementation (`AttributeError` on `list`, 404s on `/api/hub/admins`) and pass after. They cover: the core list (oldest first, `attivo`, no hash), the two routes without a cookie (401), the list and a creation whose new admin then logs in, and the three refusals (duplicate email, short password, malformed address) as 422 on the right field.
- `uv run ruff check projects/hub`, `ruff format --check`, `uv run mypy`: clean.
- `pnpm --filter hub test`: 45 passed (a new client test for `admins()` and `createAdmin()`); `lint` clean; `build` (vite plus `tsc --noEmit`) ok.
- In a real browser against a throwaway Postgres, migrated to head, with the API on 8084 and the dev server on 5190: logged in, opened «Amministratori», tried a five-character password (the browser's own `minLength` stops it), then a duplicate address (the alert reads «esiste già un amministratore con questa email» and the email field is `aria-invalid`), then a valid one: the status line appears, the table goes from two rows to three, the fields empty. The new admin's credentials then answer 200 on `/api/hub/auth/login`.

Deactivating or deleting an admin: the `attivo` flag exists and the list shows it, but nothing in the area changes it. A change-password flow for the admin themself. A `create_admin` MCP tool: `projects/hub/apps/mcp` is an admin by construction and nobody asked an agent to create accounts, so `tests/test_tools.py` is untouched. All three recorded on ORB-123.

## API changes

- `GET /api/hub/admins`, admin cookie, 200 with a list of `AdminRead`.
- `POST /api/hub/admins`, admin cookie, body `{email, nome, password}`, 201 with the `AdminRead`; 422 in FastAPI's shape on `email` (duplicate, or not an address) or `password` (under ten characters); 401 without the cookie.
- `AdminRead` (also the shape of `/auth/me` and of the login response) gains `attivo: bool` and `created_at: datetime`.
- The hub's hand-written client `projects/hub/apps/web/src/lib/api.ts`: `Admin` gains the two fields, a new `AdminCreate` type, and `admin.admins()` / `admin.createAdmin()`.

- `AdminRead` widening also reaches `/auth/me`: harmless, but it is a shape shared by three responses.
- `Header` and `Empty` in `pages/admin/lists.tsx` are now exported and reused by the new page rather than copied.
- The form disables its fields while the request runs; the password field uses `autoComplete="new-password"` so a browser does not offer to save the creating admin's own password into it.

## Screenshots

**1. The sidebar, on the freelancers list**
The admin sidebar, before and after

**2. /hub/admin/amministratori: Not Found before, the list and the form after**
The Amministratori page, before and after

Linear: ORB-123.

### DIFF
--- a/projects/hub/README.md
+++ b/projects/hub/README.md
@@ -21,7 +21,8 @@ Three public flows and the admin area behind them:
 - `/hub/accedi` and `/hub/io`: a freelancer gets back in with a magic link by mail, to
   see or change what they sent.
 - `/hub/admin/login` and the freelancer, company and signup lists behind it — a cookie
-  session, created with `orbiters createadmin` (below).
+  session. The first admin is created with `orbiters createadmin` (below); the next ones
+  from «Amministratori» inside the area.
 
 `POST /api/orbiters/signups` is the community site's signup endpoint, moved here
 unchanged on 2026-09-09: the website's form and the ChatGPT Ads conversion still post
--- a/projects/hub/apps/api/src/orbiters_api/routers/admin.py
+++ b/projects/hub/apps/api/src/orbiters_api/routers/admin.py
@@ -3,14 +3,15 @@
 One cookie, `orbiters_admin`, opaque and httpOnly, resolved against `admin_sessions` on
 every request (`deps.get_admin`). The login shares the public token bucket, so a
 password guess costs the same budget as a signup flood. Everything under this router
-reads or moves rows other people wrote; nothing here writes on their behalf.
+reads or moves rows other people wrote, or adds one more admin (ORB-123); nothing here
+writes on an applicant's behalf.
 """
 
 from typing import Annotated
 from uuid import UUID
 
 from fastapi import APIRouter, HTTPException, Query, Request, Response, status
-from pydantic import BaseModel, EmailStr
+from pydantic import BaseModel, ConfigDict, EmailStr, Field
 
 from orbiters_api.deps import ADMIN_COOKIE, AdminDep, SessionDep, SettingsDep
 from orbiters_api.downloads import cv_response
@@ -19,6 +20,7 @@ from orbiters_core.admin import AdminRead, AdminService
 from orbiters_core.comments import CommentService
 from orbiters_core.companies import CompanyService
 from orbiters_core.freelancers import FreelancerService
+from orbiters_core.models import NAME_MAX_LENGTH
 from orbiters_core.schemas import (
     CommentCreate,
     CommentRead,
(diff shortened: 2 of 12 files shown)


slavni96:
## Independent review, applied in ae2e2d71

A fresh read-only reviewer ran the hub suites (164 Python, 45 web, ruff, mypy, lint, build), checked the gating (`AdminDep` first in both signatures, so an anonymous caller gets the 401 before body validation), the responses (no password, no hash) and the ordering of the list (per-commit `created_at`, `uuid7` as the tie-break).

**Should fix, applied.**
- `AdminCreate.nome` was a bare `str`: a whitespace-only name would have been stored empty, and a name past the column or carrying a NUL byte would have reached Postgres as a 500. The rule now lives in `AdminService.create`, so the CLI shares it (blank or over-long is a 422 on `nome`), and the body is bound like every other schema here: `SafeStr` within `NAME_MAX_LENGTH`. Tests cover the three cases on the API and the two on the service.
- The `AdminUser` docstring said an admin is never created by a form. Reworded.

**Nits, applied.** `extra="forbid"` on `AdminCreate`, with a test that a body carrying `attivo: false` is a 422 rather than silently dropped. The router's module docstring no longer says nothing here writes. One README sentence says where the next admins come from. The form keeps its fields enabled while the request runs, as the login does. The web copy of the password length points at the service's constant.

**Left as is.** The dated spec keeps its `createadmin` sentence as a record. The trivially true `hasattr` assertion in the core test stays beside the API test that does the real check. The commit and PR bodies phrase the password decision differently; both say the same thing.

PR #37, opened 2026-09-10
feat(hub): the new-admin form lives in a dialog behind a button, not on the page

slavni96:
## What this changes

Ivan, after seeing ORB-123 live: «il form deve finire in una modale aperta da un pulsante non in pagina». «Amministratori» is now the list alone. A «Nuovo amministratore» button sits in the header's right slot, where the other lists keep their state filters, and opens a dialog with the same three fields. The button is a Radix trigger, so focus lands in «Nome» on open and returns to the button on close; Escape, the backdrop, the × and «Annulla» all close it. A refusal from the server keeps the dialog open with the draft and marks the field it names («esiste già un amministratore con questa email» on the email field). A success closes it, refreshes the list and leaves one line on the page: «Amministratore creato: ada@orbiters.it. Ora può accedere con la password che gli hai dato.» Every opening starts from an empty form.

The dialog primitive (`components/ui/dialog.tsx`) is the hub's own copy of PigroCRM's, as the other primitives here are, without the entrance animation whose Tailwind variants the hub's tokens do not declare.

## How I verified it

- `pnpm --filter hub test`: 48 passed, 13 files. The three new component tests in `Admins.test.tsx` failed before the change (no button, form on the page) and pass after: the list alone with the form only inside the dialog, focus in and out; a 422 kept inside the dialog on the named field with the draft intact; a 201 closing it, refreshing the list, showing the status line, and the next opening empty.
- `pnpm --filter hub lint`: clean. `build` (vite plus `tsc --noEmit`): ok.
- In a real browser against a throwaway Postgres migrated to head, API on 8084, dev server on 5190: opened the page (no form), clicked the button (dialog, focus on «Nome»), submitted a duplicate address (alert inside the dialog, still open), corrected it (dialog closed, status line, the table from two rows to three, focus back on the button).
- Nothing on the API changed; the Python suites are untouched by this diff.

The API, the list, and what ORB-123 left out (deactivation, deletion, a change-password flow, an MCP tool).

- The dialog primitive drops PigroCRM's `data-open:`/`data-closed:` animation classes on purpose; a copy with them would silently do nothing here.
- The status line lives on the page, not in the dialog, because the dialog is gone by the time there is something to say. It stays until the next successful creation replaces it.

## Screenshots

**1. The page: the inline form gone, the button in the header**
The Amministratori page, before and after

**2. The dialog open, focus on «Nome»**
The new-admin dialog, before and after

Linear: ORB-125.

### DIFF
--- a/projects/hub/apps/web/src/components/ui/dialog.tsx
+++ b/projects/hub/apps/web/src/components/ui/dialog.tsx
@@ -1,0 +1,117 @@
+import * as React from 'react'
+import { XIcon } from 'lucide-react'
+import { Dialog as DialogPrimitive } from 'radix-ui'
+
+import { Button } from '@/components/ui/button'
+import { cn } from '@/lib/utils'
+
+/**
+ * The hub's dialog, copied from PigroCRM's primitive as the others here are (hub spec §
+ * Web: the two products copy, never import across). Radix does the work that matters in
+ * a modal: focus goes in on open and back to the trigger on close, Escape and the
+ * backdrop close it, the page behind is inert. No entrance animation on purpose: the
+ * tokens file declares none of the variants PigroCRM's copy relies on.
+ */
+function Dialog({ ...props }: React.ComponentProps<typeof DialogPrimitive.Root>) {
+  return <DialogPrimitive.Root data-slot="dialog" {...props} />
+}
+
+function DialogTrigger({ ...props }: React.ComponentProps<typeof DialogPrimitive.Trigger>) {
+  return <DialogPrimitive.Trigger data-slot="dialog-trigger" {...props} />
+}
+
+function DialogPortal({ ...props }: React.ComponentProps<typeof DialogPrimitive.Portal>) {
+  return <DialogPrimitive.Portal data-slot="dialog-portal" {...props} />
+}
+
+function DialogClose({ ...props }: React.ComponentProps<typeof DialogPrimitive.Close>) {
+  return <DialogPrimitive.Close data-slot="dialog-close" {...props} />
+}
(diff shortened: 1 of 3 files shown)


slavni96:
## Independent review, applied in 814b580f

A fresh read-only reviewer ran the hub web suite (48 tests, lint, build), read the built CSS (the button's icon padding rule is there; no `data-open` rule exists, as the PR says), and checked the accessibility wiring (title, description, focus in and out, no Radix warning in the test output).

**Should fix, applied.**
- The dialog could be dismissed while the request was in flight: a late refusal was swallowed, and a late success would have closed a dialog reopened in the meantime, since TanStack's hook-level `onSuccess` outlives `reset()`. `toggle` now refuses to close while `create.isPending`, «Annulla» is disabled with the submit, and a test with a deferred response pins it: Escape and «Annulla» do nothing until the answer arrives, then the dialog closes on its own.
- The refusal test now also closes with «Annulla» and checks that the next opening carries neither the alert, nor `aria-invalid`, nor the draft.

**Nits, applied.** The `role="status"` region is always mounted with `aria-live="polite"`, so the sentence is announced. «Annulla» is a `DialogClose`. The redundant `aria-hidden` on the icon and the fragment around the single `Dialog` are gone. The first test's fetch mock returns a fresh `Response` per call.

**Left as is.** The dialog primitive keeps single quotes like the rest of this PR's files; the sibling primitives were copied verbatim from shadcn and there is no formatter to arbitrate. The Italian quote in the PR body is Ivan's own words, in guillemets, which is how this repository quotes the person who asked.

PR #38, opened 2026-09-10
feat(web): the invoice list shows the accrual period beside the date

slavni96:
## What this changes

The Fatture list showed the emission date and nothing about the period an invoice is for. Invoicing runs late here, August's work issued in September, so the list did not say which month a row was about; the accrual period (`competenza_da`/`competenza_a`, ORB-61) was readable only on the detail page. Ivan asked for both on 2026-09-10 («nella tabella delle fatture vorrei vedere sia la data che la competenza»).

Now a «Competenza» column sits right after «Data», in the list page and in the Fatture tab of a customer or a deal alike. It reads «01/08/2026 - 31/08/2026», the same string the detail page states through `formatPeriod`, or the em dash when the document declares no period. Plain text: the emission date stays the one column with a calendar icon.

## How I verified it

- Red then green: `columns.test.ts` (position right after `data_emissione`, header, the period string, the dash, plain text with no icon and a quiet dash) and `fatture/index.test.tsx` (the list renders the header after Data and the period cell). Five tests, all failing on `origin/main`, passing now.
- `pnpm --filter web test`: 1063 passed, 101 files. `pnpm --filter web lint`, `pnpm --filter web build`, `tsc --noEmit`: clean.
- After the independent review (recorded as a comment on this PR): the eleven invoice test files 135 passed; eslint and `tsc --noEmit` clean.
- Opened http://localhost:5173/app/fatture against a local stack with an issued invoice for August issued on 2 September, a draft with no period, and a proforma for September: the column reads the two periods and a dash on the draft. The pair below is that page, before from `origin/main` on 5175 and after from this branch on 5173, same API.

- Eight columns at 1440px: the table still fits its frame on the reference data (the pair shows it). A narrower viewport scrolls the table inside its box, the way `DataTable` already does for every list.
- The accessor and the cell both go through `formatPeriod`, so they cannot disagree about what nothing looks like; the `?? null` on the two fields is for a payload that predates the columns and has no key at all.

- No filter or sort by period, and no API change: the fields were already on every row.
- The column is on the tab inside a customer's and a deal's page too, unlike «Cliente» (ORB-98): the period is not what the page title already says.

## Screenshots

**1. Fatture list, the «Competenza» column after Data**

Linear: ORB-126.

### DIFF
--- a/projects/pigrocrm/apps/web/src/features/invoices/InvoicesTab.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/invoices/InvoicesTab.test.tsx
@@ -42,6 +42,12 @@ describe('InvoicesTab', () => {
     expect(screen.queryByRole('columnheader', { name: 'Cliente' })).not.toBeInTheDocument()
   })
 
+  /** The period is not what the page title already says, so the tab shows it (ORB-126). */
+  it('says which period each invoice is about, like the list page', async () => {
+    renderWithClient(<InvoicesTab owner={{ customerId: 'c1' }} />)
+    expect(await screen.findByRole('columnheader', { name: 'Competenza' })).toBeInTheDocument()
+  })
+
   /**
    * The slot exists so the button that creates an invoice sits on the tab that lists
    * them, rather than in the page header where it would be offered from every other tab
--- a/projects/pigrocrm/apps/web/src/features/invoices/columns.test.ts
+++ b/projects/pigrocrm/apps/web/src/features/invoices/columns.test.ts
@@ -28,6 +28,12 @@ const PROFORMA = {
 
 const WITH_CUSTOMER = { ...ISSUED, customer_ragione_sociale: 'ACME S.r.l.' } as Invoice
 
+const WITH_PERIOD = {
+  ...ISSUED,
+  competenza_da: '2026-08-01',
+  competenza_a: '2026-08-31',
+} as Invoice
+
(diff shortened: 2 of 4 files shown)


slavni96:
**Independent review.** A fresh read-only reviewer went over `f5693aa` with the column file's precedents, `formatPeriod`, `DataTable` and the e2e specs as context. Findings and what I did with each, in `0537ab95`:

1. `?? null` on the two fields was dead code: `InvoiceRead` has them as required `string | null`, and `formatPeriod` already absorbs an `undefined`. Removed.
2. Accessor and cell computed the period independently, the drift the file's own `data_emissione` comment warns about. Fixed: `accrualPeriod` feeds both, like `customerName`.
3. The route test copied a 20-line payload twice and edited the ORB-98 test's payload with fields it never asserts. Fixed: an `invoice()` helper and `mockInvoices()`, the ORB-98 test back to what it asserts.
4. The tab had no assertion for the new header. Fixed: `InvoicesTab.test.tsx` asserts «Competenza».
5. The ordering test's title claimed the list configuration it did not check. Fixed: it checks both the default and `{ cliente: true }`.
6. Table width at 1440px: measured by the reviewer at roughly 200px of slack, eaten by a long customer name, after which the frame scrolls as `DataTable` does everywhere. No change; a shorter format would diverge from the detail page, which the issue rules out.
7. Comment style: compliant, no action.

The reviewer also confirmed no consumer selects invoice cells by position (e2e specs select rows by text). Verdict on the original: merge after fixes; every fix it asked for is in.

PR #39, opened 2026-09-10
fix(hub): the panel is back on every public page, and off on the chooser alone

slavni96:
## What this changes

`hub-v0.4.0` (PR #35, ORB-124) took the boxed panel off `Shell`, the frame every public page shares, so the wizards, the thank-you page, `/accedi`, `/entra` and the member area went straight on the grid together with the chooser. Only the chooser was meant to: its two door cards draw their own frame, a wizard step does not. `Shell` now takes a `panel` flag, on by default, and the chooser is the one route mounted in a `Shell` with `panel={false}`. Every other public page renders exactly as it did on `hub-v0.3.1`: the review fixes of PR #35 (white inputs, alert slivers, white progress track, two comments) were compensating for the missing panel and are undone here, file by file identical to `hub-v0.3.1`.

## How I verified it

In `projects/hub/apps/web`: `pnpm lint` clean, `tsc --noEmit` clean, `vitest run` 13 files, 50 tests passed, one new: `Shell.test.tsx` pins the panel as the default and its absence with `panel={false}`. `git diff hub-v0.3.1 HEAD -- src/wizard src/pages/member src/components/ui` shows only `dialog.tsx`, which ORB-125 added and this PR does not touch. Hub dev server from this branch on 5180 beside one from `origin/main` on 5182, both at 1440×900: the company wizard's first step has its panel back, the chooser is pixel-identical between the two.

The chooser sits under a second layout route (`bareLayout`) rather than under `publicLayout` with a flag read from the location: two `Shell`s in the tree, one per look, so the choice is visible in the route table and no component reads `pathname`. Moving between `/` and `/freelance` therefore remounts `Shell` (header and footer), which costs nothing visible since neither holds state.

The chooser keeps the bare grid ORB-124 gave it. The admin area, which brings its own frame, is untouched. Recorded on ORB-128.

## Screenshots

**1. The company wizard's first step at 1440×900: the panel is back around the step**
The company wizard before and after

**2. The chooser at 1440×900: unchanged, still on the bare grid**
The chooser before and after

Linear: ORB-128.

### DIFF
--- a/projects/hub/apps/web/src/components/Shell.test.tsx
+++ b/projects/hub/apps/web/src/components/Shell.test.tsx
@@ -6,7 +6,7 @@ import {
   createRoute,
   createRouter,
 } from '@tanstack/react-router'
-import { cleanup, render, screen } from '@testing-library/react'
+import { render, screen } from '@testing-library/react'
 import { describe, expect, it } from 'vitest'
 import { Shell } from './Shell'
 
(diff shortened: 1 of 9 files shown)


slavni96:
## Independent review

A fresh read-only reviewer built the exact route tree headlessly against the installed TanStack Router and read the diff against `hub-v0.3.1`. Findings and what I did:

1. **Shell doc comment gave the member area a false reason** ("no cards of their own"): `Area.tsx` draws three boxes of its own. Fixed: the comment says the member area keeps the panel because the owner asked for the chooser alone to go bare.
2. **Nit, `router.tsx` "eleven screens"** was stale before this PR (the tree has sixteen). Fixed to "this many screens".
3. **Nit, one test with a `cleanup()` in the middle.** Split into two cases so the setup file's `afterEach(cleanup)` does the work.
4. **Design alternative**: one route with `<Shell panel={false}><Chooser/></Shell>` instead of a second pathless layout. Left as is; the reviewer agreed the chosen shape mirrors `publicLayout` and is reasonable.

Verified clean by the reviewer: `/`, `/freelance`, `/io`, `/io/modifica` and the admin paths match the intended routes, pathless ids are not URL-addressable, children order is cosmetic; every typed `to` still resolves; every file the reverted commit touched other than `Shell.tsx` is byte-identical to `hub-v0.3.1`; no e2e exercises hub routes.

Second commit verified the same way: lint, `tsc`, Vitest 51/51.

PR #40, opened 2026-09-10
feat(hub): a pencil on every admin row edits the name, the address or the password

slavni96:
## What this changes

Once created, an admin could not be changed: a typo in the name, a new address or a forgotten password had no fix short of the database. Every row of «Amministratori» now ends with a pencil («Modifica Lorenzo (demo)» for a screen reader) that opens the same dialog as the header button, prefilled, as «Modifica amministratore». The password field there reads «Nuova password», is optional, and its placeholder says «Vuota: resta quella di adesso». A refusal stays inside the dialog on the field the server names; a success closes it, refreshes the list and leaves «Amministratore aggiornato: lorenzo.fiore@orbiters.it.» on the page. Focus goes back to the pencil (or to the header button) on close.

Ivan chose the fields on 2026-09-10: name, address and password; not the `attivo` flag, not deletion.

## How I verified it

- `uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests`: 169 passed after the review commit (167 before). New: the service test (nothing given changes nothing; name, address and password each change; the old password stops authenticating and the new one works; the row's own address is no duplicate; each refusal on its field; 404 on an unknown id) and two route tests (401 without the cookie; the edit; the new credentials logging in and the old ones refused; every refusal as a 422 on `email`, `nome`, `password` or an undeclared `attivo`; an empty password as «keep it»; one's own row; 404).
- `uv run ruff check projects/hub`, `ruff format --check`, `uv run mypy`: clean.
- `pnpm --filter hub test`: 52 passed, 13 files. New: the client test (`PATCH /api/hub/admins/2` without the empty password), and two component tests (the pencil opens the dialog prefilled with the password field empty and not required; Escape returns focus to the pencil; a save without password; the create dialog empty again after an edit; a new password refused on its field and then accepted, sent in the body). `lint` clean, `build` ok.
- In a real browser against a throwaway Postgres: the pencil on the second row opened the dialog prefilled («Lorenzo (demo)», `lorenzo.demo@orbiters.it`, empty password); the first admin's address was refused with «esiste già un amministratore con questa email»; a rename to «Lorenzo Fiore», a new address and a new password went through, the list showed the new name, the status line appeared and focus was back on the pencil. Then, against the API: the old password answers 401 and the new one 200.

No deactivation (the `attivo` flag stays as it is; a body carrying it is a 422), no deletion, no MCP tool. A new address or name does not touch the admin's open sessions: they hang on the id. A new password does end them (review finding, applied): a reset is also what you do after a leak, and with no `attivo` toggle there was no other way to close a colleague's sessions; the editor's own cookie is spared, so changing your own password keeps you in. Recorded on ORB-129.

## API changes

- `PATCH /api/hub/admins/{admin_id}`, admin cookie, body with `nome`, `email`, `password` all optional and `extra="forbid"`; 200 with the `AdminRead`; 404 for an unknown id; 422 in FastAPI's shape on the field. An empty `password` is treated as absent; a present one also deletes the admin's other sessions.
- `AdminService.create` and the new `AdminService.update` share three checks (`_checked_email`, `_checked_nome`, `_check_password`), so the rules exist once.
- Hub web client `projects/hub/apps/web/src/lib/api.ts`: `AdminUpdate` type and `admin.updateAdmin({ id, nome, email, password })`, which leaves an empty password out of the body.

- The dialog is one component in three modes (`closed`, `create`, `edit`); the header button is no longer a Radix `DialogTrigger`, because there are now several openers, so focus return is done by hand in `onCloseAutoFocus` from the element the click event came from (not `document.activeElement`, which Safari does not move to a clicked button).
- The `minLength` on the password field is dropped while editing with an empty value, so the browser does not block «keep it»; once something is typed the ten-character floor is back.
- An admin can edit their own row, including their address and password; their current session keeps working because the route passes the caller's cookie to the service and the revocation spares it. Every field is checked before any is written, so a body with one good field and one bad leaves the row as it was.

## Screenshots

**1. The rows: a pencil at the end of each**
The admin rows, before and after

**2. The edit dialog, prefilled, with the optional password**
The edit dialog, before and after

Linear: ORB-129.

### DIFF
--- a/projects/hub/README.md
+++ b/projects/hub/README.md
@@ -22,7 +22,7 @@ Three public flows and the admin area behind them:
   see or change what they sent.
 - `/hub/admin/login` and the freelancer, company and signup lists behind it — a cookie
   session. The first admin is created with `orbiters createadmin` (below); the next ones
-  from «Amministratori» inside the area.
+  from «Amministratori» inside the area, where they are also edited.
 
 `POST /api/orbiters/signups` is the community site's signup endpoint, moved here
 unchanged on 2026-09-09: the website's form and the ChatGPT Ads conversion still post
--- a/projects/hub/apps/api/src/orbiters_api/routers/admin.py
+++ b/projects/hub/apps/api/src/orbiters_api/routers/admin.py
@@ -57,6 +57,19 @@ class AdminCreate(BaseModel):
     password: str
 
 
+class AdminUpdate(BaseModel):
+    """The pencil on a row (ORB-129): each field optional, absent means «keep it». An
+    empty password is «keep it» too, since that is what an untouched password field
+    sends; a present one goes through the service's ten-character rule. `attivo` is not
+    here on purpose: Ivan does not want deactivation from the area yet."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    email: EmailStr | None = None
+    nome: SafeStr | None = Field(default=None, min_length=1, max_length=NAME_MAX_LENGTH)
+    password: str | None = None
+
+
 @router.post("/auth/login", response_model=AdminRead)
 def login(
     payload: LoginRequest,
@@ -99,8 +112,9 @@ def me(admin: AdminDep) -> AdminRead:
 
 # ---- the admins ------------------------------------------------------------------------
 #
-# Who reads this area, and one more of them. No deactivation and no deletion here, on
-# purpose (ORB-123): the `attivo` flag exists and nothing in the area changes it yet.
+# Who reads this area, one more of them, and a change to one (ORB-123, ORB-129). No
+# deactivation and no deletion here, on purpose: the `attivo` flag exists and nothing in
+# the area changes it yet.
 
 
 @router.get("/admins", response_model=list[AdminRead])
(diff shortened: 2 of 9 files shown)


slavni96:
## Independent review, applied in a97fee78

A fresh read-only reviewer ran the hub suites (167 Python, 52 web, ruff, mypy, lint, build), checked the gating and the `except_id` logic on the address, traced that nothing resolves an admin by address per request, and read the fetch-mock ordering in the component tests.

**Should fix, applied.**
- Focus return read `document.activeElement`, which Safari does not move to a clicked button, so on Safari focus would have landed on `body` after closing. The opener now comes from the click event itself, for the header button and for each pencil.
- A new password did not end the admin's open sessions, and with no `attivo` toggle in the area there was no other way to close a colleague's sessions after a leak. It now deletes the row's sessions, sparing the editor's own cookie (the route passes it to the service), so changing your own password keeps you in. A service test and a route test with a second client pin both halves; the dialog's description says a new password closes open sessions.

**Nits, applied.** Every field is checked before any is written, so a two-error body leaves the row untouched (pinned). A password of only spaces is refused. The pencil's accessible name carries the address, so two admins with one name stay apart. `onOpenChange` only ever closes. The README says admins are edited in the area too.

**Left as is, with the reason.** The unique-index race between two concurrent writes to the same address is pre-existing in `create` and, with a handful of admins, not worth a handler. The check order in `create` did change (address first); no test pinned the old order and the CLI collects all three inputs before calling, so nothing visible moves except which field a doubly wrong body is blamed on.

PR #41, opened 2026-09-10
feat(web): the invoice list says what each invoice is for, without the «importata» pill

slavni96:
## What this changes

The Fatture list never showed what an invoice is for: the causale is on every row but was printed only as the detail page's subtitle. And the Stato column appended an «importata» pill to every row that came from the previous system, which in Ivan's register is most of them, so the word explained nothing a reader of the list could act on. Ivan asked for both on 2026-09-10 («aggiungi nella tabella anche la descrizione della fattura, rimuovi il tag importata che non ha senso dalla tabella»).

Now a «Descrizione» column sits right after «Cliente» (after «Numero» on the Fatture tab of a customer or a deal, which has no customer column), reading the causale or the em dash. It is the one flexible column in the table: the header asks for `100%`, so it takes whatever the other columns leave, and the text truncates there with the whole causale as the cell's title. The «importata» pill is gone from the table: `InvoiceStateBadge` takes an `importata` option, on by default, and the list turns it off. The detail page keeps the pill, where it is the reason the XML and «Rigenera documenti» actions are missing for an imported invoice.

## How I verified it

- Red then green: seven tests across `columns.test.ts` (position with and without the customer column, the causale and the dash, truncation classes and title, quiet dash, the state cell without «importata»), `InvoiceActions.test.tsx` (the badge without its pill when asked, with it by default) and `fatture/index.test.tsx` (the header after Cliente, the cell, no «importata» in the list). All failing on `origin/main`, passing now.
- The first cut used a 24rem maximum width. Measured on the live page at 1440px: the table came out 280px wider than its frame. The second commit switches to the flexible column; measured again, the table is back at 1078px inside its box and the description gets about 120px at that viewport with the sidebar open.
- `pnpm --filter web test`: 1071 passed, 101 files. `pnpm --filter web lint`, `pnpm --filter web build`, `tsc --noEmit`: clean.
- `bash projects/pigrocrm/apps/web/scripts/e2e.sh` (the Playwright suite, serial, its own Postgres): 39 passed in 1.2m.
- After the independent review (recorded as a comment on this PR): the twelve invoice and table test files 164 passed; eslint, `tsc --noEmit` clean; the page re-measured at 1440px, Descrizione 208px wide, the table about 80px past its frame inside its own scroll.
- Opened http://localhost:5173/app/fatture against a local stack with an issued invoice carrying a 170-character causale, a proforma, and an invoice imported through `POST /api/invoices/import`: the causale reads truncated with the full text on hover, and the imported row shows «Emessa» with no pill. The pair below is that page, before from `origin/main` on 5175 and after from this branch on 5173, same API.

- The description has a floor of 12rem (the review found it shrinking to its header at 1280px without one). At 1440px with the sidebar open that costs about 80px of horizontal scroll inside the table's frame, the «Pagamento» column partly behind it; from 1512px up nothing scrolls and the column grows with the screen. Reclaiming width would mean shortening or dropping another column, which is a separate decision.
- The `w-0 min-w-[max(100%,12rem)]` span: width zero keeps the cell's text out of the table's minimum, the percentage resolves to zero while the table measures itself so only the 12rem counts, and the minimum of 100% makes it fill the width the column was given. Verified by measurement at 1280, 1440 and 1920, not only by reasoning.

- The «importata» pill stays on the detail page, where it explains the missing actions. Ivan can ask for it to go there too.
- No filter or sort by description, no API change.

## Screenshots

**1. Fatture list, the «Descrizione» column after Cliente, and «Emessa» without «importata» on the imported row**

Linear: ORB-130.

### DIFF
--- a/projects/pigrocrm/apps/web/src/components/DataTable.tsx
+++ b/projects/pigrocrm/apps/web/src/components/DataTable.tsx
@@ -38,7 +38,10 @@ import { cn } from '@/lib/utils'
  * itself would leave its own header on the left. `width` is a plain CSS length passed
  * to the header cell, which is enough to keep the «⋯» column from taking a fair share
  * of the table's width -- v9's real column sizing is a feature (`columnSizingFeature`)
- * this table deliberately does not register.
+ * this table deliberately does not register. A percentage works too, and means the
+ * opposite thing: in an auto-layout table a `100%` column is handed whatever the other
+ * columns leave, which is how the invoice list's «Descrizione» column stays flexible
+ * (`features/invoices/columns.tsx`).
  */
 export interface DataTableColumnMeta {
   align?: 'left' | 'right'
--- a/projects/pigrocrm/apps/web/src/features/invoices/InvoiceActions.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/invoices/InvoiceActions.test.tsx
@@ -294,14 +294,19 @@ describe('InvoiceActions', () => {
     expect(screen.getByRole('button', { name: /^PDF$/i })).toBeInTheDocument()
   })
 
+})
+
+/** The badge lives in its own file (see there for why); its tests live here beside the
+ *  actions that read the same `importata_da` flag. */
+describe('InvoiceStateBadge', () => {
   it('shows the "imported" badge next to the state badge, naming no source', () => {
     const imported = { ...ISSUED, importata_da: 'esterno' } as Invoice
     wrap(<InvoiceStateBadge invoice={imported} />)
     expect(screen.getByText(/^importata$/i)).toBeInTheDocument()
   })
 
-  /** The list turns the pill off (ORB-130); the detail page, which renders the badge
-   *  with no option, keeps it as the reason the XML actions are missing. */
+  /** The list turns the pill off (ORB-130). Rendered with no option, as the detail page
+   *  does, the badge keeps it: that is the test above this one. */
   it('leaves the "imported" badge out when asked to, and keeps the state', () => {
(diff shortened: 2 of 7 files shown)


slavni96:
**Independent review.** A fresh read-only reviewer went over the branch (mid-review it moved from `4a305ae` to `e74b14d`; the findings are against the latter) with the table components, the CSS sizing rules and the e2e specs as context, and measured the live page at 1024, 1280, 1440 and 1920. Findings and what I did with each, in `5f9ace55`:

1. With only the slack to live on, Descrizione fell to 121px at 1440 and to its header's 81px at 1280, an ellipsis where the causale should be. Fixed: the span's minimum is `max(100%, 12rem)`, a floor the table's minimum does include. Re-measured at 1440: 208px, about 80px of horizontal scroll inside the frame; none from 1512 up.
2. The `DataTable` docblock described `meta.width` as a plain length only. Fixed: a sentence on what a percentage means in auto layout, pointing at this column.
3. The describe docblock in `columns.test.ts` still spoke of a maximum width. Fixed.
4. The tab's test asserted Competenza but not Descrizione. Fixed: it asserts the header right after Numero.
5. The badge test's docblock claimed the detail page keeps the pill without a route-level test. Fixed: the docblock claims only what the badge tests show, and the three badge tests sit in their own `describe`.
6. `description` duplicated `customerName`'s dash logic. Fixed: one `textOrDash`.
7. Accessibility of the title-only tooltip: the full causale is a text node in the span and reaches assistive tech; no change.

The reviewer confirmed no e2e spec selects invoice cells by index or asserts «importata». Verdict on `e74b14d`: merge after fixes; every fix it asked for is in. The pair in the body is re-rendered from `5f9ace55`.

PR #42, opened 2026-09-10
docs(repo): the loop scans the neighbours and keeps the card current

slavni96:
## What this changes

Before this PR the tracker loop asked an agent for one search, for the card it was about to work, and for a comment only when a finding changed the issue. Nothing asked about the cards next to it, so a change could land on top of one the other person had `In Progress` on the same files, or rediscover a backlog card from scratch; and a card could read `In Progress` from the first commit to the PR with nothing under it. Now `docs/tracker.md` § The loop has a second search, for the open cards in the same `area:*` and on the same surface by name, whose result is written on the card before work starts (links included, and "nothing adjacent" written too), and the card is kept current while the work happens: a comment at every turn a reader could not infer, and one line each for the review, a red run and a reshaped PR between `In Review` and the merge.

The three skills carry the two rules to the moment they apply. `linear-ticket` gets the calls of the neighbour scan (three `state` calls per area, `query` per noun of the surface, the project when the area is the wrong lens), what to do with each kind of neighbour, and the relation fields in the filing table. `linear-content` gets the **Adjacent** lead of the body and the **Scope**, **Waiting** and **Review** comment shapes. `pr-creation` gets the scan as a step before the branch, a section on the card during the work, and the card lines in the loop after `gh pr create`. `AGENTS.md` gets one bullet so a session that opens no skill still meets the rule, and `CONTRIBUTING.md` one sentence. Two measured API details go in `docs/tracker.md`: a state type (`started`) is a valid `state` filter and answers `In Progress` and `In Review` together, and `query` ranks rather than filters.

## How I verified it

- The two API claims were measured against the board on 2026-09-10 with the Linear MCP: `list_issues` with `state: "started"` returned ORB-130 (`In Progress`) and ORB-129 (`In Review`) and nothing else; `label: "area:repo"` with `state: "unstarted"` returned empty; `query: "columns.tsx"` returned ORB-130, ORB-126 and ORB-98 first and then twenty unrelated cards with a next page.
- The rules were run on this very card: ORB-131 carries an **Adjacent** lead (nothing open in `area:repo`; ORB-14, ORB-42, ORB-58, ORB-80 named as the last four changes to the contract), a **Scope, grown** comment for the `CONTRIBUTING.md` sentence found late, and will carry the PR, review and closing comments.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_no_real_identity_in_the_repository.py`: 6 passed. `orbiters-identity-scan` is not on this machine's PATH, so that preflight tier did not run here.
- Zero em dashes in the added lines (counted on the diff). Prose only, no code path changes; CI's `ci` job is the only check that applies.

- `docs/tracker.md` § The loop, the new "Then its neighbours" paragraph: it is the contract the skills restate, so if its wording and a skill's disagree the skill is the bug. I tried to keep every skill sentence traceable to it.
- The `**Adjacent.**` lead is written as "always, on a card you are about to work", not optional. That is deliberate: a scan that leaves no trace cannot be told from one that did not happen. If that reads as noise on a one-line chore card, the place to relax it is the contract, not the skill.
- The "While you work" cadence adds three comment kinds (Scope, Waiting, Review). The risk is a chattier board; the guard in the text is "not a step a reader could infer from the previous comment", and the Review line is one line with the record staying on the PR.

The ownership rules of ORB-58 (a neighbour under the other person is left alone exactly as before; the scan only makes sure it is seen), and the label and status vocabulary: the neighbour scan is a reading discipline, and Linear's relations already record its result. Two stale sentences were reworded in passing because their paragraphs changed anyway: the loop's PR step and `CONTRIBUTING.md` § Tracker still said the GitHub integration was pending approval, which ORB-80 corrected elsewhere. Recorded on ORB-131.

## Screenshots

Nothing a person could see changed: six Markdown files, no UI. No pair to capture.

Linear: ORB-131.

### DIFF
--- a/.claude/skills/linear-content/SKILL.md
+++ b/.claude/skills/linear-content/SKILL.md
@@ -48,10 +48,14 @@ to edit. Where a design choice is already taken, say so and where it is recorded
 reason, so the next person does not reopen a closed question.
 ```
 
-Optional, when they apply: **Evidence** (the reproduction, the measurement, the log
-line), **Blocks** / **Blocked by** (issue ids you have read), **Decision for the
-lead** (the project's lead per `docs/tracker.md`: one question, the options, your
-recommendation first).
+Always, on a card you are about to work: **Adjacent** (the open cards next to this one,
+found as the `linear-ticket` skill says, and what was done about each: linked, narrowed
+around, waited for. When there are none, the sentence that says so, with the calls that
+answered empty and the last Done cards on the same surface, so a reader can tell the
+board was read). Optional, when they apply: **Evidence** (the reproduction, the
+measurement, the log line), **Blocks** / **Blocked by** (issue ids you have read),
+**Decision for the lead** (the project's lead per `docs/tracker.md`: one question, the
+options, your recommendation first).
 
 ## Comments
 
@@ -62,11 +66,20 @@ which kind it is, so a reader can skim the thread.
   Never "working on it".
 - **Found.** `**Cause is different from the title:** ...` The reproduction or measurement,
   and what it changes about the plan.
+- **Scope.** `**Scope, narrowed:** ...` or `**Scope, grown:** ...` What the card now
+  covers that it did not, or no longer covers, and why: a neighbour found late, a file
+  that had to move too, a piece left for its own card (with the id, once filed).
+- **Waiting.** `**Waiting on Lorenzo:** ...` What you are stopped on, from whom, and what
+  you will do when it arrives. One comment when you stop, one when it lifts.
 - **Decision.** `**MCP, decided.** ...` What was chosen, the reason, where it is recorded
   (`DECISIONS.md` row, spec paragraph). If it is the lead's to take: the question, the
   options, your recommendation, and stop.
 - **PR.** `PR: <url> (branch ...)` plus one sentence on what it contains and what is
   still running (review, CI).
+- **Review.** `**Review applied:** ...` How many findings, which changed the code (commit
+  sha), which you left as they were and why. The full record stays on the PR; the card
+  gets the one line that says the PR is not what it was when it opened. The same shape
+  for a CI run that went red (`**CI red:** run ..., <job>, <cause>, fixed in ...`).
 - **Closing.** `**Merged:** <url> (merge commit ...)` or `**In production:** tag ...`
   followed by `Evidence:` and a bulleted list a reader can chase: run ids with the job
   names that went green, test counts, the request you made and what came back, what you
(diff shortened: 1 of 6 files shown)


slavni96:
## Independent review

A fresh read-only reviewer, given the worktree, the diff against `origin/main` and the six files in full, ranked twelve findings. All twelve are applied in `dd3136d`.

**Blocking, applied**

1. `linear-ticket` said all four relations are append-only arrays. `save_issue` has `blocks`, `blockedBy` and `relatedTo` as `string[]` and `duplicateOf` as `string | null`: one id, cleared with `null`. Fixed in the neighbours section and the filing row. The same false sentence was in the ORB-131 body; corrected there too.
2. Order. `tracker.md` filed the card, then scanned; `linear-ticket` said the links go in the same call that files or moves the card; `pr-creation` required `In Progress` before the scan. Followed literally that is a second `save_issue` the skill forbids. Now the contract says the scan comes before the call that files or moves the card and its result lands in that call; `pr-creation` step 1 is "exists and is yours", step 2 ends with the one move that carries links and the **Adjacent** paragraph.
3. `tracker.md` applied "do not overlap" to every started card in the area; the skills scoped it to the same surface. The contract now carries "on the same surface".
4. A `Backlog`/`Todo` card your change would close: `tracker.md` said link it, `linear-ticket` said take it if yours. Now both say: yours to take, it is your card and nothing new is filed; not yours, link it.

**Ambiguity, applied**

5. How **Adjacent** reaches a found card's body: `description` replaces on update, so the skill now names `patch` with `op: "append"`, in the move call. Verified on ORB-131: a `patch` append and `estimate` in the same call, both applied.
6. The empty case asked for "the last Done cards" that an open-states read never returns. Now: the area calls that answered empty, and the Done cards `query` ranked first.
7. `createdById` added to the scan's `fields`, since the would-close rule turns on it.
8. `**CI red:** ... fixed in ...` demanded a sha before the push. Now written when seen, sha added to the same comment on push (pr-creation and linear-content).
9. **Adjacent** is now the fifth lead in the body template, skipped only on a card filed for later.
10. `§ above` and `§ While you work` named as `§ Finding before filing`, `§ The neighbours`, `§ The loop, While you work`.
11. "in the backlog" written as `Backlog` or `Todo` in pr-creation and AGENTS.md.

**Polish, applied**

12. The "summary and loses the reasons" justification and the "`In Progress` with nothing under it" sentence were repeated in four files; kept in `tracker.md` and `AGENTS.md`, dropped from the two skills.

**Left as is:** nothing. The reviewer found the style clean (no em dashes, no "not just", sentence case, widths within each file's existing maximum) and the two API measurements supported by the text's own evidence.

PR #46, opened 2026-09-10
docs(repo): the tracker lists the board's eight projects and the ten area labels

slavni96:
## What this changes

`docs/tracker.md` § Where things are said «The four current projects, each with a lead and both of us as members» and listed `Website v1`, `Hub v0`, `PigroCRM v1` and `Monorepo hygiene v1`. The board has eight projects today and `Monorepo hygiene v1` has been `Completed` since 2026-09-10, so an agent reading the page to pick a project for a card was reading a list that was wrong in both directions.

Now the section carries a table of what `list_projects` answered on 2026-09-10, by initiative, with lead and state: eight rows, one of them `Completed, 2026-09-10`. It says the board is the authority when the two drift and that the change opening or closing a project updates the table in the same PR. A paragraph under it says where repository-wide work goes now that `Monorepo hygiene v1` is closed: nowhere, until a `Monorepo hygiene v2` exists, which is what ORB-131 and ORB-136 did.

Two more things in the same section were stale the same way, so I fixed them in the same change. The Area row listed nine labels and the board has ten: `area:hub` exists since 2026-09-09, ORB-76 asked for that line and closed with the label checked and the document untouched. The row names it now, and so does the `addLabels` row of the `linear-ticket` skill, which had the same nine. And the membership rule («every project always carries a lead and both members») is false for the four projects opened after the list was written, which carry Lorenzo alone. The rule stays and now says which four and why they cannot be repaired from an agent session: `save_project` takes a `lead` and has no member field, which also gets a bullet under § API details.

## How I verified it

- `list_projects` with `team: "Orbiters"`, `includeMembers: true`, `fields: [name, status, lead, members, initiatives, completedAt]` on 2026-09-10: eight projects, leads, initiatives and states as in the table; `Monorepo hygiene v1` with `completedAt: 2026-09-10T09:20:53Z`; `Hub v1`, `Website v2`, `Deploy and access hygiene v1` and `Indexing and SEO v1` with one member each. A second call without `includeArchived` returned the same eight, so the page does not tell the reader to pass it.
- `list_issue_labels` with `includeGroups: true`: ten children under `Area`, `area:hub` included, and the ten type labels as the Type row already had them.
- `save_project` schema read from the MCP surface: `lead` present, no member field.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_no_real_identity_in_the_repository.py`: 7 passed, the one preflight check that runs on every diff and is installed on this machine (`orbiters-identity-scan` is not on this box).

The table is a dated snapshot, and the old bullet list was one too; the new sentence that the change opening or closing a project updates the table is what stops it going stale a third time, and it is a rule with no check behind it. The alternative was to drop the list and keep only the `list_projects` pointer, which I did not do because a reader without an MCP session (a reviewer on GitHub) still wants the names.

The four incomplete projects are named as incomplete rather than fixed. The MCP surface cannot add a member, so that is a UI action for whoever opens Linear next; nothing on the board was changed for it.

## Screenshots

Nothing a person can see changed: two Markdown files, no UI.

Linear: ORB-136.

### DIFF
--- a/.claude/skills/linear-ticket/SKILL.md
+++ b/.claude/skills/linear-ticket/SKILL.md
@@ -106,7 +106,7 @@ wrong.
 | `milestone` | the milestone id from `list_milestones(project)`, unless the issue genuinely belongs to no body of work. It is accepted and not echoed back: trust `list_milestones` progress, not the response. |
 | `title` | the observed problem, not the fix: "the invoice page shows the numbers but never the document", not "add a PDF preview". |
 | `description` | per the `linear-content` skill. Real newlines, never `\n` escapes. |
-| `addLabels` | exactly two: one from the `type` group (`fix`, `feature`, `refactor`, `chore`, `docs`, `test`, `ci`, `design`, `security`, `spike`) and one from `Area` (`area:web`, `area:api`, `area:core`, `area:mcp`, `area:infra`, `area:ci`, `area:website`, `area:brand`, `area:repo`). Both are groups: a second label from the same group is silently dropped. Use `addLabels`, never `labels`: `labels` replaces the whole set. |
+| `addLabels` | exactly two: one from the `type` group (`fix`, `feature`, `refactor`, `chore`, `docs`, `test`, `ci`, `design`, `security`, `spike`) and one from `Area` (`area:web`, `area:api`, `area:core`, `area:mcp`, `area:hub`, `area:infra`, `area:ci`, `area:website`, `area:brand`, `area:repo`). Both are groups: a second label from the same group is silently dropped. Use `addLabels`, never `labels`: `labels` replaces the whole set. |
 | `priority` | 1 Urgent, 2 High, 3 Medium, 4 Low. A field, never a label. |
 | `estimate` | the team's points. |
 | `assignee` | never omitted. `"me"` when you will do the work, the person who asked for it when they will. A card filed for later still gets one: an empty assignee reads as free to the other agent. |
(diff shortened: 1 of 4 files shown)


slavni96:
Independent review (read-only reviewer on the worktree, `git diff origin/main...HEAD`), seven findings, all applied in `aaf214da`:

1. **Should-fix, the skill still required a project.** `linear-ticket` § Filing said `project` is the project id, full stop, while the new tracker paragraph files a repository-wide issue with none. The row now says "or omitted when the issue is repository-wide and fits no open project", and the tracker's "What an issue carries" list says the same.
2. **Should-fix, `docs/adding-a-project.md` §9 presented `Monorepo hygiene v1` as the open home for repository-wide work.** Rewritten in the past tense, says it is closed, points at § Where things are for where such an issue goes now.
3. **Should-fix, "updates this table in the same PR" had no referent** when a project is opened or closed on the board alone. Now: whoever opens or closes a project adds or updates the row, in the PR that ships the release or in one of its own. §9 also asks for the row, and for the members to be set in the UI, when a project is created.
4. **Nit, "without a lead and without both members"** read literally excluded the four one-member projects. "or", here and in the same sentence of `AGENTS.md`.
5. **Nit, the four incomplete projects had no owner for the fix.** Filed as ORB-137 (assigned to me, a Linear UI action); the paragraph points at it.
6. **Nit, "neither open `Monorepo` project"** hard-coded that two are open. "no open `Monorepo` project".
7. **Nit, "for a day"** was a duration the board read did not measure. Dropped.

Nothing left as it was. The reviewer also checked the table against the mounted `list_projects` and `save_project` schemas and found the API claims, Markdown and line widths correct.

PR #48, opened 2026-09-10
fix(hub): a text field is white on every OS appearance, like the landing's

slavni96:
## What this changes

On a Mac whose appearance is set to Dark, every text field in the hub (the wizards' «Nome», «Cognome», the email, the textareas, the member area's edit page, the admin's forms) rendered with a grey-blue fill, while the landing's own signup fields stayed white. Ivan spotted it side by side on 2026-09-10 and asked for the hub to match the landing: white.

The hub has no dark theme, and never did. What it had was shadcn's `dark:bg-input/30` on Input and Textarea, with Tailwind v4 binding `dark:` to the OS `prefers-color-scheme` by default. Under `.site`, where `--input` is the full-strength ink, that is prussian blue at 30% over the paper.

Now a visitor's OS appearance changes nothing: `dark:` is bound to a `.dark` class nothing sets (`@custom-variant dark (&:is(.dark *))`, the line PigroCRM's tokens.css already carries), and Input and Textarea paint `bg-card`, the same `#ffffff` as `orbiters.css`'s `.signup input`, instead of letting the page ground show through. The compiled stylesheet carries no `prefers-color-scheme` block at all now; it carried exactly one.

## How I verified it

- `pnpm --filter hub exec vitest run`: 14 files, 57 tests passed. The two new ones in `tokens.test.ts` pin the variant and that no `.dark` class exists anywhere it could match. The first one failed before the fix on `@custom-variant`, then passed.
- `pnpm --filter hub lint` clean, `tsc --noEmit` clean, `pnpm --filter hub build` green.
- `grep -c prefers-color-scheme dist/assets/index-*.css`: 1 on `main`, 0 on this branch. The `.dark\:bg-input\/30` rule compiles to `:is(.dark *)`.
- Playwright (the repo's own package, headless Chromium, 1440x900) against `vite preview` of `main` and of this branch, with `colorScheme` emulated. Computed `background-color` of the «Nome» field on `/hub/freelance`:

| | `main` | this branch |
|---|---|---|
| dark appearance | `oklab(0.21 -0.02 -0.06 / 0.3)` (ink at 30%) | `rgb(255, 255, 255)` |
| light appearance | `rgba(0, 0, 0, 0)` (the card behind it) | `rgb(255, 255, 255)` |

I did not open the admin area in a browser: it needs a session, and the change is the two primitives it shares with the wizards.

- `bg-card` instead of `bg-transparent` is a visible change in light appearance only where a field sits on the paper rather than on a white card: the admin's forms. A white field on the paper is what the landing does, so I took it, but it is the one place a reviewer could prefer the ground showing through.
- The `dark:` classes stay in `button`, `badge`, `input` and `textarea`. Stripping them would be tidier and would rot the moment the next shadcn primitive is pasted in; the variant makes them inert either way, and the test guards the variant.
- `DECISIONS.md` gets a row saying neither product has a dark theme until one is designed. That is a rule, not a picture, so it belongs there.

## Screenshots

**1. The wizard's first screen under a dark OS appearance, `main` on the left, this branch on the right.** The pink outline on «Nome» is the autofocus ring, present in both frames.

Wizard first screen under a dark OS appearance, before and after

Under a light appearance the pair shows no visible difference on this screen: the fields sit on the wizard's white card, so a transparent field and a white one look the same there.

Linear: ORB-138.


slavni96:
**Independent review, applied in `840ac9c3`.** Eight findings, none blocking.

Changed:
1. The DECISIONS.md row and the tokens.css comment claimed neither product has a dark theme. The CRM has a `.dark` token block (its `tokens.test.ts` pins it) and no switch. Both now say so, and the CRM's reason for the same `@custom-variant` line is stated as the opposite of the hub's.
2. The row's last sentence made a white fill a rule for every Orbiters surface while the CRM's fields are still `bg-transparent`. Scoped to the hub and the landing; the CRM is named as not this row's.
3. The second test read only the four primitives' `className=` attributes. It now reads every `src/**/*.ts(x)` and `index.html` and fails on a bare `dark` inside any string literal, so a `.dark` added in Shell, AdminLayout or the HTML shell fails it.
4. Dropped the `not.toMatch(/@media[^{]*prefers-color-scheme/)` on the source stylesheet: the offending block was Tailwind's output, never in tokens.css, so it guarded nothing.

Left as is:
5. `&:is(.dark *)` rather than Tailwind's documented `&:where(.dark, .dark *)`: equivalent while nothing sets the class, and it matches the CRM's line.
6. `bg-card` side effects checked by the reviewer: disabled still replaces the fill, aria-invalid touches border and ring only, `.site` leaves `--card` white. A disabled field under `.site` painting 50% ink is pre-existing and not this PR's.
7. `__dirname` in vitest is fine and already used at line 14.
8. Commit subject at 74 characters; no repo rule caps it.

Reviewer also confirmed: 57/57 tests, eslint clean, a production build with zero `prefers-color-scheme` occurrences and the `dark:` utilities emitted under `:is(.dark *)`.

PR #49, opened 2026-09-10
feat(web): every column of the economic charts says its figure

slavni96:
## What this changes

The two economic charts of the dashboard labelled one segment per column, the tallest, and only when it filled more than a third of the plot. On Ivan's 2026 data «Andamento economico» left February and April mute, «Proiezione economica» labelled two months out of nine, and the «9.970,12 €» on August was the gold «Da incassare» segment alone, which a reader takes for the month's total. Ivan asked on 2026-09-10, with a screenshot: «fai sì che il valore economico si veda sempre dentro alle colonne».

Now every segment tall enough to hold the text says its own value inside. A column too short for a label says its value above itself, small and muted, rather than nowhere. A column that stacks more than one segment says its total above, so the figure on top of a month is the whole month. The plot keeps a strip of headroom for those figures.

The figure above a stacked column is the height of the bar as money, and it comes from the server: `CashMonth` gains `pila_andamento` (incassato + costi) and `pila_proiezione` (all four series), summed once in `AnalyticsService.cash_overview` beside the shares, because a sum is a figure and no figure is born in the browser (criterion 14, `no-browser-arithmetic.test.ts`). They are named for what they are, a bar's height with the costs inside, not a total an assistant on the MCP would read as an income.

## How I verified it

- Red then green. Core: `test_analytics_cash.py` asserts the two heights on a month with a paid invoice and a cost, on an empty month, and on a month whose only money is owed (in the projection's height, not the cash chart's); both failed with `AttributeError` on `origin/main`. Web: `MonthlyBars.test.tsx`, five tests (every tall segment labelled, the bar's height above a stacked column, nothing above a single tall segment, the value lifted above a short column and not repeated inside, nothing above an empty month); all five failed on `origin/main`.
- `pnpm --filter web test`: 1087 passed, 102 files. `tsc --noEmit`, eslint, `pnpm --filter web build`: clean. `no-browser-arithmetic.test.ts` passes: the chart still does no arithmetic on amounts.
- `test_analytics_api.py`, `test_dashboard_api.py`, `test_analytics_cash.py`, `test_dashboard_commercial.py`: 90 passed. ruff, ruff format, mypy: clean.
- `uv run pytest -q -n 3 --dist loadfile -m "not slow and not planner" projects/pigrocrm`: 4190 passed.
- After the independent review (recorded as a comment on this PR): the dashboard and criterion-14 test files 102 passed, the whole web suite 1089 passed; the page re-measured at 1440 and 1024: the tallest column ends on the plot's bottom edge, no label is clipped, none is shown where it would not fit.
- `pnpm --filter web generate:api` with the API up: the only change in `api-types.ts` is the two fields and their descriptions.
- Opened http://localhost:5173/app/?tab=economica against a local stack with seven imported invoices spread over the year (six paid, one owed), a September proforma and an August cost. Before: nine figures across the two charts. After: sixteen, one per column plus a total above the two stacked Augusts; February's «1.450,00 €» sits above its short bar.

- The stacked August total in the projection reads 18.104,03 €, which includes the 320,00 € cost stacked on top of the revenue, since that is what the column stacks. Whether costs belong in that stack is the open question Ivan raised about the projection («non sono convinto») and is not decided here.
- A segment shorter than 16px carries no label (the 320 € cost sliver above August): its figure is in the tooltip and in the table view, and the column's total above already contains it.
- Labels are 10px in a twelfth of the plot's width; at 1440px «18.104,03 €» fits with room to spare. Below 72px of column (about 1280px of viewport) they hide rather than clip, and the tooltip and the table carry the figures.

- What each chart stacks, the default «per competenza» reading, the series colours, the tooltips and the table view. The projection's content is a separate decision pending Ivan's answer.

## API changes

- `GET /api/analytics/panoramica`: every element of `cassa.mesi` gains `pila_andamento` and `pila_proiezione` (decimal strings, the stacked bar's height as money, costs inside). No status code changes. `pnpm --filter web generate:api` was run with the API up on `localhost:8000`; `projects/pigrocrm/apps/web/src/lib/api-types.ts` matches.

## MCP surface

- No tool added or renamed. `get_economic_dashboard` returns the same `CashMonth` rows, so its output carries the two new keys.

## Screenshots

**1. The economic tab, both charts: a figure on every column, the bar's height above a stacked month, the short February labelled above its bar**
Economic charts, before and after

Linear: ORB-139.

### DIFF
--- a/projects/pigrocrm/apps/web/src/features/dashboard/EconomicTab.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/dashboard/EconomicTab.test.tsx
@@ -3,6 +3,7 @@ import { render, screen, waitFor, within } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { EconomicTab } from './EconomicTab'
+import type { CashMonth } from './queries'
 import { api } from '@/lib/api'
 
 vi.mock('@/lib/api', async (importOriginal) => {
@@ -45,7 +46,16 @@ function failed(error: unknown, status: number) {
   return { error, response: new Response(null, { status }) } as never
 }
 
-function month(mese: number, incassato = '0.00', da_incassare = '0.00', bozze = '0.00', costi = '0.00') {
+/** Typed as the wire shape so a field added to `CashMonth` fails here rather than
+ *  rendering something nobody asserted on (ORB-139: the two `pila_*` heights). The
+ *  heights are illustrative, not sums: the chart prints them, it never checks them. */
+function month(
+  mese: number,
+  incassato = '0.00',
+  da_incassare = '0.00',
+  bozze = '0.00',
+  costi = '0.00',
+): CashMonth {
   return {
     anno: 2026,
     mese,
@@ -53,6 +63,8 @@ function month(mese: number, incassato = '0.00', da_incassare = '0.00', bozze = 
     da_incassare,
     bozze,
     costi,
+    pila_andamento: incassato,
+    pila_proiezione: incassato,
     quote_andamento: { incassato: incassato === '0.00' ? 0 : 1, costi: costi === '0.00' ? 0 : 0.1 },
     quote_proiezione: {
       incassato: incassato === '0.00' ? 0 : 0.6,
--- a/projects/pigrocrm/apps/web/src/features/dashboard/MonthlyBars.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/dashboard/MonthlyBars.test.tsx
@@ -3,7 +3,7 @@
  * value, the tallest, and only when it filled more than a third of the plot: most months
  * said nothing and the one figure shown read as the column's total when it was one
  * segment's. Now every segment tall enough to hold the text says its value, a column too
- * short for a label says it above, and a stacked column says its total above.
+ * short for a label says it above, and a stacked column says its height as money above.
  */
 import { render, screen, within } from '@testing-library/react'
 import { describe, expect, it } from 'vitest'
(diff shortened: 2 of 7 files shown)


slavni96:
**Independent review.** A fresh read-only reviewer went over `68e3136` with criterion 14, the chart's consumer and fixtures, and the live page measured at 1440, 1024 and 400px. Findings and what I did with each, in `bc3be040`:

1. Amounts truncated to an ellipsis below about 1280px. Fixed: the column is a container and the labels hide under 72px of width instead of clipping; the tooltip and the table keep the full figure. Re-measured: no clipped label at 1440, none shown at 1024.
2. A payload without the new fields printed «NaN €» above every stacked month, and the EconomicTab fixtures were such a payload. Fixed: nothing above when the height is missing, with a test; the fixture is typed as `CashMonth` so the compiler asks for every field.
3. The tallest stacked column overflowed the plot by 5 to 9px and dropped its baseline. Fixed: fixed 16px label slot, 24px headroom, `min-h-0`; re-measured, the column bottom sits exactly on the plot bottom.
4. `totale_*` added costs to revenue under a name that reads as an income, on the API and the MCP alike. Decided: the fields are `pila_andamento` and `pila_proiezione`, the height of the stacked bar as money, costs inside, with descriptions saying so. The printed figure is unchanged, it matches the bar a reader measures; whether costs belong in the stack at all is Ivan's open question about the projection, not this PR's.
5. Test hygiene: typed `month()`, an `empties()` helper, a test for a stack with a sliver segment, a test for a payload with no height, a core assertion where the two heights differ by exactly the projection-only series. Done.
6. `column-total` shared the columns' test-id prefix. Renamed `lifted-value`.
7. Contrast: informational, no change.
8. MCP: no test pins the month's key set; no change.

Verdict on `68e3136`: do not merge as is. Every fix it asked for is in, and the pair in the body is re-rendered from `bc3be040`.

PR #51, opened 2026-09-10
fix(core): an em dash in a causale no longer costs the invoice its XML

slavni96:
## What this changes

A causale typed with an em dash passed «Emetti», spent number 18 of 2026, and every XML export after it refused with «invoice.causale: FPR12 ammette in Causale solo caratteri latini di base o Latin-1». The parties have had a Latin pre-check since ORB-56; the document's own text had none, and U+2014 is outside FPR12's `String*LatinType` range, so the number was gone before anything said no.

Now typographic punctuation is spelling, not content. `latinise` maps the dashes, the curly quotes, the ellipsis, the bullet and the Unicode spaces to their plain Latin-1 equivalent, and one `_check_text` applies it before the width and the Latin range, for the writer and for every pre-check alike, so the two cannot disagree. A `Causale` stored as «Consulting services — August 2026» is written as «Consulting services - August 2026» and the file validates. What survives the mapping and is still outside the range (a Cyrillic letter, an emoji) is refused with the field named, and `issue` now runs that check on `Causale`, `Descrizione`, `UnitaMisura` and `RiferimentoNormativo` through `check_document_text_exportable`, before the counter moves. The stored text is not rewritten: the row and the PDF keep what was typed, only the FatturaPA file spells it. Invoice 18/2026 becomes exportable as it stands, without an annulment.

## How I verified it

- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_fatturapa.py -n 0`: 73 passed (66 before, 7 new: the mapping on `Causale` and `Descrizione`, an em dash in a party name through pre-check and writer alike, a Cyrillic causale refused by field, the pre-check and the writer returning identical details for causale, descrizione and unita_misura, and an ellipsis that crosses the 200-character bound only once spelled).
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_issue.py -n 0`: 38 passed (35 before, 3 new: a Cyrillic causale and an emoji description refused before the counter moves, with `_nothing_consumed`; the ORB-140 case itself, an em dash causale issued and then exported with a hyphen in the file). The three new tests failed before the fix.
- `uv run pytest -q projects/pigrocrm/packages/core/tests projects/pigrocrm/apps/api/tests projects/pigrocrm/apps/mcp/tests -n 0 -m "not slow"`: 4200 passed, 75 deselected. Serial on purpose: `-n 3 --dist loadfile` on this box hit the testcontainers reaper race with the other containers running here (943 setup errors, 0 failures, the same files pass serially).
- `uv run ruff check projects/pigrocrm`, `uv run ruff format --check projects/pigrocrm`, `uv run mypy`: clean.
- Reproduced the defect over MCP against the live data before writing anything: `export_invoice_xml` on 18/2026 answered the message above. Not re-run after: the live API runs what is deployed, so the export of 18/2026 is proven here by the issue-then-export test and by hand only after a release.

- The mapping is punctuation and spaces only. A euro sign in a PEC still refuses, as the ORB-56 test pins; a `€` in a causale refuses too, before the number. Mapping it to `EUR` would change an email address, so I left the line there and wrote it into `docs/design/DECISIONS.md`.
- `_check_widths` now measures the latinised party values, because an ellipsis is one character stored and three written. Nothing else about the party checks changed.
- `export_xml`'s promise that a re-export produces the same bytes holds: a value with a mapped character was refused before this change, so no exported file can contain one.

The web after «Emetti» still stays on the consumed proforma and shows the export error as if the emission had failed. That is ORB-134, linked from the card, and not this PR.

## Screenshots

Nothing a person can see changed: the fix is in the XML writer and the pre-check inside `packages/core`, with no UI, route or MCP tool touched. No pair to capture.

Linear: ORB-140.


slavni96:
## Independent review, applied in cd0e352e

Seven findings from a read-only reviewer on the diff, ranked as they came.

1. **UnitaMisura is `String10Type`, Basic Latin only, at most ten characters** (medium). Neither side applied that bound: `m²` (U+00B2, inside Latin-1) and a ten-character unit whose ellipsis spelled into an eleventh passed the pre-check and the writer and failed the XSD. Fixed: `_check_text` takes the writer's `pattern`, both sites pass `_UNITA_MISURA_MAX` and `_BASIC_LATIN_RE`, two parametrised cases pin it.
2. **Email and PECDestinatario had no width in the pre-check** (low, pre-existing). Fixed: both carry `_EMAIL_MAX` through `_check_latin`, one test.
3. **C0 control characters passed every pre-check and crashed the writer with a `ValueError`** (low, pre-existing). Fixed: `_check_text` runs `escape_xml` first and raises `ValidationFailed` naming the field; the existing test now asserts that type and that the pre-check answers the same.
4. **`_check_widths` duplicated the width rule with its own message** (low). Fixed: the three widths moved into `_check_latin` with the writer's wording; `_check_widths` keeps the two coded fields. No test pinned the old wording.
5. `__all__` gains `check_document_text_exportable` and `latinise`.
6. `_LineText` docstring reworded.
7. A test pins the determinism argument (every key above U+00FF, every spelling inside Latin-1), and `RiferimentoNormativo` gets its own pre-check test.

Also taken: the other invisible joiners and direction marks (U+200C to U+200F, U+2060) and the hyphen bullet join the spellings. Left as is: single angle quotes, which have no plain spelling that keeps their meaning.

Verified after: `test_invoice_fatturapa.py` 78 passed, `test_invoice_issue.py` and `test_invoice_artifacts_xml.py` 57 passed, ruff and mypy clean; the full serial core, api and mcp run is in progress and will be reported on the card.

PR #52, opened 2026-09-10
chore(ci): drop the identity guard and stop the heavy checks selecting paths that cannot change them

fiorelorenzo:
## What this changes

Two things, both about what a local `preflight` run costs.

**The repository-wide identity guard is gone.** `test_no_real_identity_in_the_repository.py` was written on 2026-09-09 to get the tree clean before we went public, and it kept running on every diff afterwards: `identity-shapes` and `identity-names` both matched `**`, and the file itself sat in PigroCRM's suite, so editing it selected `pigrocrm-py-test`, `pigrocrm-py-corpus` and `pigrocrm-py-planner` and cost minutes of PostgreSQL on a change that could not break any of them. That is what ORB-141 measured. Lorenzo's call on 2026-09-10: the check was a one-off for the publication and does not need to keep running. So the file is deleted and both checks leave `.github/preflight.json`; the name scanner stays outside the repository and is simply no longer wired to anything here. What the tree loses is written down in `docs/design/DECISIONS.md` rather than left to be discovered: a fiscal shape, a fixture company name on a shipped surface (ORB-97) or the real root slug in a deploy vhost is caught by review now, and by nothing else.

**Three heavy checks stopped selecting paths that cannot change their answer.** This is the rest of ORB-141, and it is what makes a run short rather than moving one file out of the way:

- `pigrocrm-py-corpus` and `pigrocrm-py-planner` ran on any change under `apps/api` or `apps/mcp`. The two marked files import `pigrocrm.core` and the local `corpus` helper and nothing else, so a router or a tool cannot change which plan PostgreSQL picks over core's schema. Twelve minutes, serial, for an answer that could not move. They select on core now, and the trunk's corpus job gets the same narrowing through a `pigrocrm_py_corpus` filter of its own, because that job sits twelve minutes in front of every preview deploy.
- `pigrocrm-images` selected a two-image docker build on `projects/pigrocrm/deploy/**`. The build reads one file from there, `deploy/nginx/spa.conf` (Dockerfile.web line 30); the host vhosts and `setup-server.sh` are installed on the server and never enter an image.
- `website-e2e` selected Playwright, serial on port 4173, on the vitest files under `projects/website/src/`. Playwright loads `projects/website/e2e/` and previews the build, which no unit test enters.

## How I verified it

`preflight --list --base HEAD`, one changed file at a time, which is the honest way to read a selection change:

- an edit under `projects/pigrocrm/apps/api/src/` selects `pigrocrm-py-lint`, `pigrocrm-py-types`, `pigrocrm-py-test` and `pigrocrm-e2e`, and no longer `pigrocrm-py-corpus` or `pigrocrm-py-planner`;
- an edit to `deploy/nginx/pigro.joinorbiters.conf` selects nothing at all, where it used to select `pigrocrm-images`;
- on this branch's own diff, `pigrocrm-images` and `website-e2e` are no longer selected, and no `identity-*` check exists to select.

The rest, run for real: `uv run ruff check projects/pigrocrm` and `ruff format --check` (495 files, clean), `uv run pytest --collect-only projects/pigrocrm` (4,273 tests collected, so the deletion leaves no import or fixture behind), `pnpm --filter website lint`, `pnpm --filter website test` (205 passed) and `pnpm --filter website exec tsc --noEmit`. Both edited manifests parse (`json.load`, `yaml.safe_load`, and `pigrocrm_py_corpus` is reachable as a `changes` output).

I did not run `pigrocrm-py-test`, `pigrocrm-py-corpus` or `pigrocrm-py-planner` in full. This diff selects them because it deletes a file under `packages/core/tests`, which is the last time that path will trigger them; the suite is otherwise untouched, collection is clean and nothing imports the deleted module. Twenty minutes of PostgreSQL would have proved nothing this diff can break, which is the same argument the PR itself makes.

The narrowing of `pigrocrm-py-corpus` is the one judgement call. It rests on the two `slow` files importing only `pigrocrm.core`, so if a corpus assertion ever reaches through the API layer it must come back to selecting `apps/**`. Two other things found on the way and worth knowing: `dorny/paths-filter` ORs its patterns under the default quantifier, so a `'!path'` line in an existing filter would have matched almost everything unless the step also set `predicate-quantifier: some-with-excludes`, which is why the CI side is a separate filter rather than an exclusion; and the `why` on `identity-shapes` claimed CI had an `identity` job asking the same question, which was never true.

## Screenshots

Nothing visible changed: this is a manifest, a workflow, a deleted test and three comments.

Linear: ORB-141.


fiorelorenzo:
Independent review (read-only, fresh agent, ranked by severity): four findings, no correctness defect. All four applied in `d93f599`.

1. **The new DECISIONS row did not say which row it supersedes** (P2), and this file's own header asks for exactly that. The 2026-09-10 footer attribution row still described `identity-shapes` as the gate that keeps a fixture company name off a public surface, so a reader consulting the live rule would have been told a gate exists that does not. The row now names it and names what pins the two footers instead: `landing-pages.test.ts` for the website, `Shell.test.tsx` for the hub.
2. **`pigrocrm_core` is not a name in this tree** (P3), in the manifest's `why` and again in the ci.yml comment: the distribution is `pigrocrm-core`, the import root `pigrocrm.core`, and the two marked files also import the local `corpus` helper beside them. Both are under `packages/**`, so the narrowing itself was right and only the sentence a future auditor reads was wrong. Fixed in both places.
3. **`website-image` deserved the same `unless` as `website-e2e`** (P3, called a judgement call). The reviewer checked it: `projects/website/Dockerfile` runs `pnpm --filter website build` and `vite.config.ts` pins the rollup inputs to the four HTML files, so a `src/**/*.test.ts` reaches neither `dist` nor the nginx layer, and this branch's own comment change was still selecting a docker build. Applied, and `preflight --list` no longer selects it here.

Nothing was left as it was. The reviewer confirmed all four narrowings from the code independently: which files carry `slow` and `planner` and what they import, that `Dockerfile.web` line 30 is the only reader of `deploy/`, that `playwright.config.ts` sets `testDir: './e2e'`, and that the new `pigrocrm_py_corpus` output is plumbed correctly with the `!cancelled()` and `all == 'true'` escape hatches intact, so a skipped corpus job is neither a false green nor a false red.

CI on `ddc70cc`: run 34531976979, `ci` green, `pigrocrm · python` 4m13s (the whole suite, on a runner, without the deleted file), `hub · python` 1m11s, the three web gates green, images and corpus skipped by their filters.

PR #53, opened 2026-09-10
fix(web): after Emetti the page lands on the fattura, and a consumed proforma says which

slavni96:
## What this changes

Issuing a proforma creates a new numbered row and freezes the proforma (spec 5). The web toasted «Documento emesso», rendered the proforma's own PDF, and stayed on a page that had just become a consumed proforma with no actions and no XML; the fattura was somewhere in the list, to be found by number. Ivan hit exactly this the evening ORB-140 shipped: «non ci sta il pulsante».

Now, on that page:

- After «Emetti» on a proforma the route navigates to the issued fattura, where «Scarica XML» and «Rigenera documenti» are. A draft fattura is issued in place and stays. The toast names the number: «Fattura 2026/18 emessa».
- The render the web asks for after emission targets the issued row, not the proforma, so the fattura is the document that gets printed.
- A consumed proforma carries one sentence: «Questa proforma è stata emessa come fattura 2026/18: PDF e XML FatturaPA sono lì», with the number as a link to the fattura's page. If the list answers nothing (a race with the emission just made, a network error) the sentence still says where to look.

The sentence needs the way back from a proforma to its fattura, which nothing stored: the fattura points at the proforma through `origine_proforma_id`, the proforma at nothing. So the invoice list gains `origine_proforma_id` as a filter, in `InvoiceListQuery`, the repository and `GET /api/invoices`.

## How I verified it

- `pnpm --filter web test`: 1092 passed in 103 files (1089 before). New: `ConsumedProformaNotice.test.tsx` pins the sentence, the link target and the exact query the page sends, plus the empty answer; `InvoiceActions.test.tsx` pins that from a proforma the render call carries the issued row's id, `onIssued` receives the issued row, and the toast reads «Fattura 2026/18 emessa». The existing «a failed render is not a failed emission» test still passes.
- `pnpm --filter web lint`, `pnpm --filter web build` (vite and `tsc --noEmit`): clean.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_service.py -n 0`: 55 passed (54 before). New: the list by `origine_proforma_id` answers the one fattura and not a sibling proforma's or a standalone one; an unknown id answers empty.
- `uv run pytest -q projects/pigrocrm/apps/api/tests/test_invoices_api.py -n 0`: 24 passed (23 before). New: create a proforma, confirm, issue over HTTP, then `GET /api/invoices?origine_proforma_id=` answers exactly the issued row; a malformed id is 422.
- `uv run ruff check`, `uv run ruff format --check`, `uv run mypy`: clean.

- `useProduceArtifacts(invoiceId)` now accepts a target id at `mutate` time and invalidates both. The one caller that passes one is «Emetti»; «Rigenera documenti» still passes nothing and renders the row it sits on.
- The navigation happens right after `issue` succeeds, while the render is still in flight against the issued row; the fattura's page then shows the PDF when its query is invalidated. If the render fails, the warning toast still fires, on the fattura's page.
- The API endpoint itself renders the artefacts inside `POST /issue` and answers an error when that fails after the number is spent, which is what Ivan saw as a failed emission. Not touched here: filed as ORB-143.

## API changes

`GET /api/invoices` gains the optional query parameter `origine_proforma_id` (UUID). No response shape changed. `projects/pigrocrm/apps/web/src/lib/api-types.ts` was regenerated from the app's own OpenAPI document (`create_app().openapi()` piped to `openapi-typescript`, the same generator `generate:api` runs against a live API); the diff is the one new line.

The MCP `list_invoices` tool builds its `InvoiceListQuery` explicitly and gains no `origine_proforma_id` argument: an agent reads the fattura's `origine_proforma_id` from `get_invoice`, and the reverse lookup is the page's need. Recorded on the card.

## Screenshots

Before and after of the consumed proforma's page, attached below once captured on the same throwaway data: the before shows the page with the state pill and nothing else, the after shows the sentence with the link to the fattura.

Linear: ORB-134.

pair-1-proforma-consumata

### DIFF
--- a/projects/pigrocrm/apps/api/src/pigrocrm_api/routers/invoices.py
+++ b/projects/pigrocrm/apps/api/src/pigrocrm_api/routers/invoices.py
@@ -71,6 +71,9 @@ def list_invoices(
     # (§6.2). The card links here and nowhere else, and the rows returned are counted by
     # the same predicate function the card's `COUNT` uses.
     scadute: Annotated[bool, Query()] = False,
+    # The fattura a consumed proforma was issued as: the proforma's page asks for the
+    # one row whose `origine_proforma_id` is its own id (ORB-134).
+    origine_proforma_id: Annotated[UUID | None, Query()] = None,
     limit: Annotated[int, Query(ge=1, le=200)] = 50,
     cursor: Annotated[UUID | None, Query()] = None,
 ) -> InvoicePage:
@@ -82,6 +85,7 @@ def list_invoices(
         anno=anno,
         stato_pagamento=stato_pagamento,
         scadute=scadute,
+        origine_proforma_id=origine_proforma_id,
         limit=limit,
         cursor=cursor,
     )
--- a/projects/pigrocrm/apps/api/tests/test_invoices_api.py
+++ b/projects/pigrocrm/apps/api/tests/test_invoices_api.py
@@ -361,6 +361,33 @@ def test_the_list_filters_and_paginates(
     assert filtered["items"] == []
 
 
+def test_the_list_answers_the_fattura_a_consumed_proforma_became(
+    logged_in: TestClient,
+    customer: dict[str, Any],
+    fiscal_profile: dict[str, Any],
+    emitter: dict[str, Any],
+) -> None:
+    """After «Emetti» on a proforma the numbered row is a different one (spec 5), and the
+    consumed proforma's page needs a way to it: `?origine_proforma_id=` answers exactly
+    that row (ORB-134)."""
+    proforma = logged_in.post(
+        "/api/invoices",
+        json={
+            "customer_id": customer["id"],
+            "tipo": "proforma",
+            "righe": [{"descrizione": "Consulenza", "prezzo_unitario": "100.00"}],
+        },
+    ).json()
+    assert logged_in.post(f"/api/invoices/{proforma['id']}/confirm").status_code == 200
+    issued = logged_in.post(f"/api/invoices/{proforma['id']}/issue", json={}).json()
+    assert issued["id"] != proforma["id"]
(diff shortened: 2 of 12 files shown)


slavni96:
## Independent review, applied in f60a6da4

Seven findings from a read-only reviewer on the diff.

1. **The render warning was lost after navigating** (medium, a regression this PR would have introduced). `onIssued` navigates, the bar unmounts, and TanStack Query drops the callbacks passed to `mutate()` once the observer has no listeners, so a failed render after «Emetti» on a proforma would have shown nothing on the fattura's page. Fixed: `mutateAsync` with a `catch`; a new test unmounts from `onIssued` the way the route does. I verified it fails on the callback form and passes on the promise form.
2. **On the fattura's page the render is still in flight and nothing shows it** (low). Left as is: a click on «Rigenera documenti» in that window renders twice, idempotently. A `mutationKey` and `useIsMutating` would be the fix if it ever bites; nothing in the web uses `mutationKey` yet.
3. **The notice's comment named a race that cannot happen** (low). Reworded: the fattura is written in the same transaction that consumes the proforma, so an empty answer is the request failing after retries.
4. **The in-place case was not pinned** (low). The existing draft test now asserts `onIssued` receives the same id, which is what the route's guard relies on.
5. **`origine_proforma_id` has a foreign key and no index** (low). Left as is: one page view per consumed proforma on a freelancer's register. Noted for a migration if the filter becomes a list facet.
6. Apostrophes in the fallback sentence written as characters instead of escapes.
7. Leaving the MCP `list_invoices` without the filter is defensible: the tool already omits `stato_pagamento` and `scadute`, and the commit message records the choice.

After: `pnpm --filter web test` 1093 passed, lint and `tsc --noEmit` clean.

PR #54, opened 2026-09-10
feat(hub): the admin area lists PigroCRM's spaces and who opened them

slavni96:
## What this changes

PigroCRM provisions a space per signup, and nothing read the registry back: the only way to know how many spaces exist and who opened them was `psql` on the server. Ivan asked for it in the hub's admin area («dammi modo di vedere quali istanze di pigro esistono e di chi sono»).

Now the sidebar has «Istanze Pigro», at `/hub/admin/pigro`: every space in the registry, newest first, its name linking to `https://pigro.joinorbiters.com/<slug>/app/`, the owner and the day it was created. When the address that opened a space belongs to somebody who filled in the hub's wizard, the row names them (`Ada Lovelace`, linking to their card) with the address underneath; otherwise the address stands alone. A header count says how many there are.

The hub never touches the CRM's database. PigroCRM gains `GET /api/tenants/`, answering the registry rows to a caller presenting `PIGROCRM_REGISTRY_TOKEN` as a bearer, and not existing at all (404) while the token is unset, so a self-hosted installation exposes nothing new. The hub's API relays it under the admin cookie at `GET /api/hub/pigro/istanze`, calling the CRM with `ORBITERS_PIGRO_API_URL` and `ORBITERS_PIGRO_REGISTRY_TOKEN` through the same HTTP seam the mail uses. Without the token the page says «Il registro di Pigro non è configurato: manca ORBITERS_PIGRO_REGISTRY_TOKEN.»; when the CRM answers anything but a list it says «Pigro non ha risposto (401).» or «Pigro ha risposto qualcosa che non è un elenco.». The reasoning is a new row in `docs/design/DECISIONS.md`.

## How I verified it

Test first on both sides, watched red, then green.

- `uv run pytest -q projects/pigrocrm/apps/api/tests/test_tenants_api.py projects/pigrocrm/packages/core/tests/test_compose_environment.py`: 87 passed. The two new tests cover the 404 without a token, the 401 on a missing or wrong bearer, and the list after a real signup with exactly `id`, `slug`, `owner_email`, `created_at`. The compose test is what makes `PIGROCRM_REGISTRY_TOKEN` reach the container.
- `uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests`: 178 passed. The four new API tests hand a fake through `get_http_call` and read the request that would have left (one GET, the bearer), the member match, the case-insensitive match, the 503 and the two 502 sentences; the 401 loop gained the route.
- `pnpm --filter hub lint`, `pnpm --filter hub test` (61 passed, three of them the new page and one the client), `pnpm --filter hub build` (tsc clean).
- `uv run ruff check`, `ruff format --check` and the root `uv run mypy` are clean for both projects.
- End to end on a throwaway Postgres: the hub API from this branch on 8144 with a stub CRM on 8145 answering three rows behind the token, an admin created with `orbiters createadmin`, Ada applied through the wizard route, then `/api/hub/pigro/istanze` answered the three spaces with Ada matched. The screenshot below is that state.

- The route lives in PigroCRM's public `tenants` router and reads `SettingsDep`, the same as its siblings; `registry_token` is not in `OVERRIDABLE_KEYS`, so `space_settings` cannot set it, and the check is `secrets.compare_digest`. Under a space prefix (`/<slug>/api/tenants/`) it answers the same registry, as the signup route does.
- The member match is by lowercased address in one query; the CRM writes owners lowercased, and the test still covers a mixed-case row.
- The hub's `Pigro.test.tsx` mocks `Link` from the router, which no other page test does yet, because the page renders a `Link` to a member's card outside a router.

- No MCP tool on either side lists the spaces: the ask was the admin page. Adding one is a line if wanted.
- The root installation's own space (`PIGROCRM_ROOT_SLUG`) is not in the registry and does not appear.
- Nothing in a space is counted: the registry knows who opened it, not what is inside.
- Going live takes two tags, `pigrocrm-v*` first so the route exists, then `hub-v*`, and the same token written by hand into both host `.env` files.

## Screenshots

**1. The admin area: the sidebar before, and the new «Istanze Pigro» page after**
The hub admin area, before and after: a new sidebar entry and the list of spaces with their owners

The before frame is `main`'s web app on the same API, on the «Iscrizioni» page, since the route did not exist there; the after frame is this branch. Same admin, same fixture data, 1440×900.

Linear: ORB-142


slavni96:
## Review applied (`dbc27e7`)

An independent read-only review of the diff found five things; all five are in the commit above, with a failing test first for the three that change behaviour.

1. **Medium, fixed.** A non-ASCII bearer made `GET /api/tenants/` a 500: Starlette decodes headers as latin-1 and `secrets.compare_digest` refuses a `str` with a non-ASCII character. The comparison is now on UTF-8 bytes; a test sends `Bearer t\xe9ken` and expects 401.
2. **Medium, fixed.** A CRM that is down (refused connection, DNS miss, timeout) reached the hub as an unhandled `URLError`, so the page showed the generic sentence and the log a traceback. `PigroRegistry.list_spaces` now wraps the seam call and raises `PigroUnavailable("Pigro non risponde.")`, a 502; a test makes the fake raise `OSError`.
3. **Low, fixed.** `urllib_call` sent `data=b""` on a GET, which urllib turns into `Content-Length: 0` plus a form content type. It now sends `body or None`; `test_http.py` asserts the request's `data` is `None` for a bodiless call and unchanged for a POST.
4. **Low, fixed.** The API test fixture built `Settings` without `pigro_api_url`, so a shell exporting `ORBITERS_PIGRO_API_URL` would have moved the exact-URL assertion. Declared explicitly.
5. **Low, fixed.** `Pigro.test.tsx` used `queryByRole` where a miss should fail on the lookup; now `getByRole`.

Checked by the reviewer and left as is: the token cannot be set from `space_settings` (`registry_token` is not in `OVERRIDABLE_KEYS`), the route under a space prefix answers the same registry with the same token, `GET /api/tenants` without the slash 307s to the checked route, error bodies relay a status number or a fixed sentence and never the CRM's body, the hub imports nothing from PigroCRM, and every product string is Italian while the rest is English.

**CI red on the first run** (run 34536678428, `pigrocrm · python`): `test_mcp_surface_coverage.py` requires every public service method to be an MCP tool or a declared exclusion, and `TenantService.list` was neither. It is now a declared exclusion, in the same commit: the registry is read by another product with a token, not by an agent acting inside one space.

PR #81, opened 2026-09-11
fix(web): the invoice list under Tutte shows a fattura once, not beside its proforma

slavni96:
## What this changes

Issuing a proforma creates the numbered row and freezes the proforma as «Consumata» (spec 5). Under «Tutte» the list showed both rows, 2026/18 and PROV-2026-0002, with the same customer, description, period and total, so every issued proforma read as two documents. Ivan on 2026-09-11: «se ho la fattura dovrei vedere solo quella e non la proforma».

Now a consumed proforma is treated as what it is, the antecedent of its fattura: with no state chosen the list shows the fattura alone. The «Consumata» chip still lists exactly them, and the consumed proforma's own page keeps naming its fattura (ORB-134), so nothing is unreachable. The customer and deal «Fatture» tabs, which have no chips, leave them out always.

The predicate is the server's, like `scadute`: `GET /api/invoices` gains `escludi_consumate` (boolean, default false), `InvoiceListQuery` and the repository apply `stato <> 'consumata'` on top of the other filters. No filtering in the browser, so the page and its cursor stay the API's.

## How I verified it

- `pnpm --filter web test`: 1095 passed in 103 files (1093 before). New in `routes/app/fatture/index.test.tsx`: under «Tutte» the request carries `escludi_consumate: true` and no `stato`; pressing «Consumata» sends `stato: consumata` and drops the flag; pressing «Emessa» drops it too; unpressing brings it back. New in `InvoicesTab.test.tsx`: the deal tab sends `{ deal_id, escludi_consumate: true }`.
- `pnpm --filter web lint` and `tsc --noEmit`: clean.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_service.py projects/pigrocrm/apps/api/tests/test_invoices_api.py -n 0`: 81 passed (79 before). Core: `escludi_consumate` drops the consumed proforma and nothing else, `stato=consumata` still answers it. API: confirm and issue over HTTP, then the three reads (`everything`, `escludi_consumate=true`, `stato=consumata`) answer as expected.
- `uv run ruff check`, `uv run ruff format --check`, `uv run mypy`: clean.

- The list page sends the flag only when `stato === null`. A chip for any state, «Consumata» included, is an explicit ask and gets exactly that state, so a person pressing «Bozza» does not have consumed proformas removed from a set they were never in anyway.
- The tabs send it unconditionally. A customer's page therefore never shows the consumed proforma; the fattura row carries the period and total, and the proforma is one click from the fattura through the list's «Consumata» chip if ever needed.

## API changes

`GET /api/invoices` gains the optional boolean query parameter `escludi_consumate`. No response shape changed. `projects/pigrocrm/apps/web/src/lib/api-types.ts` was regenerated from the app's own OpenAPI document (`create_app().openapi()` piped to `openapi-typescript`, the same generator `generate:api` runs against a live API). Beyond the one new parameter, the file catches up with the tenants routes ORB-142 (PR #54) added to the API without regenerating; `tsc --noEmit` is clean on the result and the web does not call those routes.

The MCP `list_invoices` tool builds its `InvoiceListQuery` explicitly and gains no argument: an agent that wants the register without the consumed proformas asks by `stato`, and one that reads a fattura sees `origine_proforma_id`. Recorded on the card.

## Screenshots

Before and after of `/app/fatture` under «Tutte» on the same throwaway data, attached below once captured: before, the fattura and its consumed proforma as two rows; after, the fattura alone, with the «Consumata» chip still there to ask for the proforma.

Linear: ORB-169.

pair-1-fatture-tutte


slavni96:
## Independent review, applied in 477408dc

Six findings from a read-only reviewer on the diff; nothing serious on substance.

1. **Tab test comment orphaned** (low): the new test was inserted between the ORB-98 comment and the test it documents. Fixed, each comment sits on its own test.
2. **`lambda` with a `noqa`** in the API test (low): now a nested `def`.
3. **The rule was not recorded** (low): a row in `docs/design/DECISIONS.md` now says that under «Tutte» a consumed proforma is not a row and the fattura stands for it.
4. **MCP asymmetry**: the web could ask for the register without consumed proformas and an agent could not. Taken: `list_invoices` gains `escludi_consumate` with the reason in its Italian docstring, passed through to `InvoiceListQuery`. The MCP surface tests pass unchanged, since no tool was added or renamed.
5. `stato=consumata&escludi_consumate=true` answers empty. Left as is: no caller sends both, and the pair is legal and empty.
6. Commit scope `fix(web)` for a change whose predicate lives in core: noted, the pushed commit stays.

The reviewer also regenerated `api-types.ts` from `create_app().openapi()` in its own scratch and found it byte-identical to the committed file, and confirmed the tenants routes are the same on `origin/main`, so the extra lines are main's stale types catching up. It checked every other consumer of the list (the overdue drill-through, the analytics revenue and pending figures, the calendar, the reminders): all pin `tipo=fattura` with an explicit state, so no figure ever counted a consumed proforma and nothing disagrees with the list now.

After: web 1095 passed, lint clean; API and MCP tests 542 passed; ruff and mypy clean.

PR #82, opened 2026-09-11
feat(api): the signup can ask the hub whether an email is a community member

slavni96:
## What this changes

Nothing told PigroCRM whether an address belongs to an Orbiters community member, so the signup at `/app/registrati` asked a member for a name the hub already holds (ORB-173, spec 2026-09-12 §6.3). Two routes now exist, one per product, and neither imports the other:

- **Hub**: `POST /api/hub/members/lookup` with `{"email": "..."}` answers `{"membro": true, "nome": "Ada", "cognome": "Lovelace"}` for a freelancer whose address matches case-insensitively (the `uq_freelancers_email_lower` rule), `{"membro": false, "nome": null, "cognome": null}` for anyone else, never an error for an unknown address. It is the mirror of the CRM's `GET /api/tenants/` (ORB-142): bearer `ORBITERS_PIGRO_REGISTRY_TOKEN`, 404 `Not Found` while the token is unset so nothing says there is a door, 401 with a wrong or missing bearer, `compare_digest` on utf-8 bytes. The query is `MemberService.lookup` in `orbiters_core`, so the MCP could reuse it.
- **CRM**: `POST /api/tenants/membro` with `{"email": "..."}`, no auth like the signup itself but throttled per client IP (five a minute, then 429 with `Retry-After: 60`), answers `{"membro", "nome", "cognome", "spazi"}`: the first three relayed from the hub through `pigrocrm.core.tenants.hub.lookup_member`, `spazi` an integer count of the spaces the registry holds for that `owner_email` (`TenantService.count_for_owner`). New setting `PIGROCRM_HUB_URL` (default `https://joinorbiters.com`), documented in `.env.example` and forwarded to the `api` container in `docker-compose.yml`; the bearer is the existing `PIGROCRM_REGISTRY_TOKEN`, which now travels in both directions. Plain `urllib`, five second timeout, redirects refused so the bearer never follows a `Location` elsewhere, body capped at 4096 bytes, and every failure (no token, no URL, refused connection, non-2xx, 3xx, malformed or oversized body) is `membro: false` without an exception and without a log line: the community is the fast lane of the signup, never a gate.

Both hops carry the address in a body and never in a query string, so no access log or proxy on either host keeps it.

Commits: `01c57f01` the hub, `8fb76ed5` the CRM, `a980ebde` two repository guards CI caught (compose forwarding, MCP exclusion), `d7708281` the independent review's six findings (see the PR comment). The hub half ships with a `hub-v*` tag, the CRM half with a `pigrocrm-v*` tag; each side is inert until the other is live and the token is set on both hosts.

## How I verified it

From the worktree root, serially, after the review commit:

```
$ uv run ruff check projects && uv run ruff format --check projects && uv run mypy && uv run pytest -q -n 0 projects/hub/apps/api/tests projects/pigrocrm/packages/core/tests/test_hub_lookup.py projects/pigrocrm/apps/api/tests/test_tenants_api.py projects/pigrocrm/packages/core/tests/test_compose_environment.py projects/pigrocrm/apps/mcp/tests/test_mcp_surface_coverage.py
All checks passed!
570 files already formatted
Success: no issues found in 291 source files
612 passed in 11.84s
```

Tests: hub `test_member_api.py` (no token → 404 whatever is presented; wrong, missing or latin-1 bearer → 401; a member applied as `Ada@Studio.it` found through `  ada@studio.IT `; unknown and non-address → `false`; no body or `?email=` → 422). CRM `test_hub_lookup.py` (member with the bearer and the address in the body, non-member, 401, `OSError`, `AssertionError` surfaces, garbled, wrong-shaped and oversized body, no token/URL/address never call, the real `urllib` transport against a closed loopback port, and a tiny loopback `http.server` that answers 302: exactly one POST reached it and `membro` stayed false). CRM `test_tenants_api.py` (`spazi` 0 before the signup and 1 after, the slug absent from the body, `lookup_member` monkeypatched; the real client against `http://127.0.0.1:9` with a token answers 200 `membro: false`; malformed, empty, missing body and `?email=` are 422; the sixth question in a minute is 429 with `Retry-After: 60` and another `X-Real-IP` has its own bucket). `test_architecture.py` passed on the earlier run and the new core module imports only the standard library and pydantic.

## API changes

- Hub, new: `POST /api/hub/members/lookup`, body `MemberLookupRequest {email: str}` → `MemberLookup {membro: bool, nome: str|null, cognome: str|null}`; 404 without `ORBITERS_PIGRO_REGISTRY_TOKEN`, 401 on a wrong bearer, 422 on a missing body. The hub's hand-written web client (`projects/hub/apps/web/src/lib/api.ts`) is unchanged: nothing in the SPA calls this.
- CRM, new: `POST /api/tenants/membro`, body `MemberQuestion {email: EmailStr}` → `MemberAnswer {membro, nome, cognome, spazi: int}`; 422 on a malformed or missing address, 429 past five a minute per client. `pnpm --filter web generate:api` was not run: no web code reads this route yet (that is ORB-176, the wizard, which this card blocks), and the generated `api-types.ts` will be refreshed there together with its first consumer.

- The two names in the CRM answer are readable by an unauthenticated caller who knows an address (throttled, five a minute per IP). Ivan asked for the member's name to be prefilled; the trade-off and the alternative (name only after the address is proven by the mail link) are recorded as a **Decision** comment on ORB-173 for him to keep or reverse.
- `pigrocrm_api/ratelimit.py` is a deliberate copy of the hub's limiter: the two products share no code by rule, and a dependency for eighty lines was not worth a `shared/` package yet.
- The hub's `email` is a plain `str`, not `EmailStr`: the spec says never an error for an unknown address, and a malformed one is just unknown. The CRM side validates (`EmailStr`, 422), so a non-address never reaches the hub.
- No row in `docs/design/DECISIONS.md`: this mirrors the 2026-09-10 cross-product decision (ORB-142) and creates no new rule.

## Screenshots

Nothing visible changed: two API routes, a setting, a client and a limiter, no UI on either side. The screen that will consume this is ORB-176.

Linear: ORB-173.


slavni96:
Independent review of the first three commits, applied in `d7708281`. Six findings, six changes, nothing left as it was:

1. **The address sat in the URL on both hops** (`?email=`), so uvicorn's access log and any proxy would have kept it. Both routes are now POST with `{"email": ...}` in the body: `POST /api/hub/members/lookup` and `POST /api/tenants/membro`. Responses unchanged; a query string is a 422 on both sides, and the tests assert it.
2. **The CRM answer listed the slugs an address owns to an unauthenticated caller.** `spazi` is now an integer count (`TenantService.count_for_owner` replaces `slugs_for_owner`; the MCP exclusion moved with it). The wizard only needs to know there is something to get back into; the mail carries the addresses. `nome` and `cognome` stay by product decision, recorded on ORB-173 with the trade-off and the alternative.
3. **`POST /api/tenants/membro` had no rate limit and the CRM had no limiter.** Added `pigrocrm_api/ratelimit.py`, a per-client token bucket with the hub's constants (five a minute, 429 with `Retry-After: 60`, `X-Real-IP` / last hop of `X-Forwarded-For`), called first thing in the route. Not imported from the hub: the two products share no code. Tested: the sixth request is 429, another client has its own bucket.
4. **`lookup_member` followed redirects with the bearer and swallowed every exception.** The opener now carries an `HTTPRedirectHandler` whose `redirect_request` returns `None`, so a 3xx stays a 3xx and reads as `membro: false`; a loopback `http.server` test proves exactly one POST leaves and nothing reaches the `Location`. The `except` is narrowed to `(OSError, http.client.HTTPException, ValueError)`, so the suite's network guard (`AssertionError`) surfaces, with a test for that too. The body is read to `MAX_BODY_BYTES + 1` (4096) and anything longer is not a member.
5. **API tests could aim at the real hub through a developer's `.env`.** `container_settings` now sets `hub_url="http://127.0.0.1:9"`.
6. **The route-ordering comment claimed `/membro` could be shadowed by `/{slug}/disponibile`.** They differ in segment count and cannot collide; the comment now only says the route sits with the other public signup routes.

Verification after the commit: ruff clean, format clean, mypy clean (291 files), `612 passed in 11.84s` on the hub API suite, `test_hub_lookup.py`, `test_tenants_api.py`, `test_compose_environment.py` and `test_mcp_surface_coverage.py`. The PR body is updated to the final shapes.

PR #84, opened 2026-09-11
feat(core): a space is born with its default stages, templates and categories

slavni96:
## What this changes

A space is born ready. `TenantService.provision` now seeds the default pipeline stages, the default templates and the default cost categories right after the first admin, and creates the emitter profile with the signup name as `ragione_sociale` and nothing fiscal. The defaults gain an editable «Offerta» template with five flat variables (`render/assets/template-offerta-default.md`), because the offer asset the renderer uses iterates a list no compilation form can fill. A new `ensure_defaults(session)` applies each seed only while its table is empty, so a default somebody deleted on purpose never comes back; `pigrocrm ensure-space-defaults` walks the registry and applies it to every space, and the API image's `CMD` runs it after the migrations so the eight spaces that exist today catch up at the first boot after the deploy, without a command run by hand. The root installation is not touched.

Why: read in production on 2026-09-12, all eight spaces had zero stages, zero templates and zero categories, and creating a deal answered «nessuno stato aperto configurato». The seeds were a button in Impostazioni or a CLI on the server. This is the first card of the PigroCRM v2 project; the design record (`docs/superpowers/specs/2026-09-12-onboarding-product-led-design.md`), the plan, the `DECISIONS.md` row and the project row in `docs/tracker.md` ride along.

## How I verified it

From the worktree root, serially:

- `uv run pytest -q -n 0 -m "not slow and not planner" projects/pigrocrm/packages/core/tests projects/pigrocrm/apps/api/tests projects/pigrocrm/apps/mcp/tests` → `1 failed, 4213 passed, 75 deselected in 288.80s`. The one failure was `test_time_report_pdf.py::test_the_seeded_template_exists_and_names_no_freelancer`, which asserted exactly two seeded templates; the last commit updates it to three and `uv run pytest -q -n 0 projects/pigrocrm/packages/core/tests/test_time_report_pdf.py` → `12 passed`.
- The new tests: `test_space_defaults.py` (3), the extended provisioning test and the CLI test in `test_tenants.py`, the two offer tests in `test_templates_service.py`; `test_tenants_api.py` unchanged and green (79 passed on the focused run).
- `uv run ruff check projects/pigrocrm && uv run ruff format --check projects/pigrocrm && uv run mypy` → clean, `no issues found in 290 source files`.
- `docker build -f projects/pigrocrm/Dockerfile.api .` → ok; `docker run --rm <image> sh -c "cd /app && uv run --no-sync python -m pigrocrm.core.cli --help"` lists `ensure-space-defaults` among the subcommands.

- The `CMD` has three steps now. `ensure_space_defaults` always returns 0 and reports an unreachable space on stderr (type only, never the text: a psycopg error can carry the URL with the password), so a furnishing problem can never keep the API down.
- «Only while the table is empty» rather than «always»: a space that deleted every stage gets them back (a pipeline with no open stage cannot take a deal), one that kept a custom template of its own and deleted the seeds does not.
- `ensure_defaults` is called inside `provision`'s try, so a seed failure undoes the space like a migration failure. No hook in `deps._tenant_session_factory` on purpose: ORB-170 is moving that registry into core.
- The seeded offer uses flat variables and `cliente.*`/`emittente.*`/`oggi` from the document scope; `render/assets/template-offer.md` is untouched.

## Screenshots

Nothing a person sees changes on this branch: the pages are the same, a new space simply has stages, templates and categories in them where it had none. No pair captured for that reason.

Linear: ORB-171.


slavni96:
Independent review applied in `af230498`. Findings and what I did:

1. The registry step of `pigrocrm ensure-space-defaults` sat outside the guard, so a registry that cannot be reached would have stopped the `CMD` before uvicorn. Now everything in the command is guarded: a registry failure is one line on stderr with the exception type and exit 0. Test: `test_the_cli_never_fails_the_boot_when_the_registry_is_unreachable`.
2. The CLI derived a space's database from the slug; it now reads `db_name` from the registry row.
3. A seed failure printed «non raggiungibile»; it prints «non arredato (Type)» now. Test: a ghost registry row whose database does not exist is skipped, the URL never appears, the other space is still furnished.
4. No test covered `_undo` when a seed fails mid-provisioning: `test_a_failing_seed_undoes_the_space_and_frees_the_name` monkeypatches `ensure_defaults` to raise.
5. The offer's placeholders were checked by regex only: a test now renders the seeded body through `render_template` with the five answers and the document scope.
6. `nome[:255]` was dead (`TenantSignup.nome` is capped at 200) and a blank name would have made an empty emitter header: `data.nome.strip() or data.slug`.
7. Mid-file `import re` moved to the top.

Left as is: the offer reaches only spaces whose templates table is empty, by design («seed only while empty»); the eight production spaces qualify, the root and a space that pressed the seed button do not. The spec §6.5 says so.

PR #85, opened 2026-09-12
feat(api): one enters PigroCRM with a link by mail, and the password stays as a second way

slavni96:
## What this changes

PigroCRM asked for a password the community never gave, sent no mail, and a person who forgot the password lost the space. Now one enters as in the community, with a link by mail. `POST /api/auth/link` takes an email and answers 202 whether the address is known or not; under a space's prefix it mails that space's user, at the root it mails one link per space the registry says the address owns, then the root itself; the mail leaves after the response. `POST /api/auth/entra` spends the link once and sets the same cookies `login` sets, at the same path. The first link entry of a user writes `users.email_verificata_il` and revokes every refresh token issued before it, which is what will make the session opened at signup (ORB-176) safe. The login page asks only for the email («Mandami il link»); «Hai una password? Accedi con la password» opens today's form, so the root, the spaces of the first days and the e2e admin get in as before, and nobody is asked for a password by default any more. `/app/entra` is the page the link lands on.

Resend sends the mail through `pigrocrm.core.mail`, the same seam as the hub's `mail.py` and not an import of it: `PIGROCRM_RESEND_API_KEY` and `PIGROCRM_MAIL_FROM` reach the container; without a key the two endpoints answer 503 with a sentence that says to use the password. Spec `projects/pigrocrm/docs/superpowers/specs/2026-09-12-onboarding-product-led-design.md` §6.1 and §6.2 (on `main` since #84), plan `docs/superpowers/plans/2026-09-12-orb-172-posta-e-login-via-link.md`, a `DECISIONS.md` row.

## How I verified it

From the worktree root, serially:

- `uv run pytest -q -n 0 projects/pigrocrm/apps/api/tests/test_auth_api.py projects/pigrocrm/apps/api/tests/test_tenants_api.py` → `39 passed` (five new link/entra tests, one root-to-space link test).
- `uv run pytest -q -n 0 projects/pigrocrm/packages/core/tests/test_magic_link.py projects/pigrocrm/packages/core/tests/test_auth_users.py projects/pigrocrm/packages/core/tests/test_migrations.py projects/pigrocrm/packages/core/tests/test_tenants.py` → `76 passed` (the migration test compares metadata and the provisioning test runs `alembic upgrade head` for real).
- `test_mail.py` (6) and `test_architecture.py` → `21 passed`; `test_compose_environment.py` and `test_mcp_surface_coverage.py` green with the new setting forwarded and the two service methods excluded.
- `uv run ruff check`, `ruff format --check`, `uv run mypy` → clean (`292 source files`).
- `pnpm --filter web lint`, `pnpm --filter web test` → `105 files, 1101 tests passed`; `pnpm --filter web build` (vite + `tsc --noEmit`) ok.
- The full Python suite of core, api and mcp is running serially as this opens; I will add its count here.
- Not run: the Playwright e2e (`:8000` is shared on this machine); the login helper and the two auth specs got the one extra click, and the `push` to `main` runs them.

## Migrations

`0034_magic_links`: `users.password_hash` becomes nullable, `users.email_verificata_il` (timestamptz, null) is added, `magic_link_tokens` is created (uuid id, `user_id` FK to `users` on delete cascade, `token_hash` unique, `expires_at`, `used_at`, timestamps). Safe on production tables: an `ALTER COLUMN ... DROP NOT NULL`, an `ADD COLUMN`, a `CREATE TABLE`. `test_migrations.py`'s `compare_metadata` is clean. The downgrade puts the NOT NULL back and fails if a row has no password, on purpose.

## API changes

- `POST /api/auth/link` body `{email}` → 202 `{ok: true}`; 503 without a sender.
- `POST /api/auth/entra` body `{t}` → 200 `UserRead` + the two session cookies; 401 for a spent, expired or unknown link.
- `apps/web/src/lib/api-types.ts` regenerated from the app's OpenAPI document (`openapi-typescript` on `create_app().openapi()`): only the two routes and their three schemas are added.

- The revocation on first link entry (`MagicLinkService.enter`): it consumes every still-valid refresh token of the user, including one on another device, once in the account's life. That is the guarantee ORB-176 relies on.
- `_owned_slugs` in `routers/auth.py` opens the registry with its own engine and treats a registry that cannot be reached as «no spaces»: the API test client has no registry database, and a self-hosted CRM may never have one.
- The 202 for an unknown address plus the mail in a background task: the response says nothing about whether the email exists. There is no rate limit on `link`, as there was none on `login`; ORB-100 covers the hub's side of the same question.
- `_origin` prefers `PIGROCRM_PUBLIC_URL` and falls back to the request's base URL; behind the nginx of the compose stack the header is what nginx forwards.

## Screenshots

**1. The login, email first**
The login before and after: email and password with «Accedi», then email alone with «Mandami il link» and «Hai una password? Accedi con la password»

Both frames from Vite dev servers with no API behind them (the page renders the same, the `me` and `root` calls fail quietly), 1440×900, `scale: css`; the before from a worktree on `origin/main`. The button is disabled until an email is typed. `/app/entra` has no before: it did not exist.

Linear: ORB-172.


slavni96:
Independent review applied in `555709fd`. Findings and what I did:

1. `entra.tsx` re-ran its effect on every render of `AuthProvider` (a fresh `enterWithLink` closure while the mutation is pending) and spent the token twice, reading its own second 401 as a dead link. Once per mount now, with a `useRef` guard, and a test that re-renders with new props and asserts one call.
2. `_origin` fell back to the request's Host when `PIGROCRM_PUBLIC_URL` was empty, which would have mailed a live token to whatever host the caller named. The link now uses the configured public origin only; without it `POST /api/auth/link` answers 503 with a sentence naming the variable. Test: 503 without a public origin; the mail carries the origin and never `testserver`. Under a prefix the space's settings already carry `/<slug>` in `public_url`, so nothing is appended there (a test caught the doubled slug).
3. `_owned_slugs` ran `ensure_tenants_database` (DDL) on an anonymous request and let a missing space database become a 500 for the one address that owns it. Now a plain read on the registry URL, `SQLAlchemyError` → no spaces; each owned space is opened in its own guarded helper and skipped on failure.
4. Timing difference between a known and an unknown address at the root: left as is. The hub's `request_link` has the same shape, the DB work is one indexed lookup plus an insert, and ORB-100 covers the rate-limit side of that class for both products.
5. Revocation on first entry consumes refresh tokens only; an access JWT issued before it lives out its `access_token_minutes`. Accepted and written in `enter`'s docstring.
6. A deactivated user's link was not spent and would have worked again on reactivation within its lifetime: spent first, then the `attivo` check, with the test extended.
7. Tests added: the link asked under a space's prefix, the earlier session dying at the first entry while the one `entra` issued still refreshes, the 503 without a public origin.
8. Style: `background.add_task(sender.send, mail)` without the pass-through; `LinkToken.t` capped at 128 characters. The `open_space` helper suggestion is left for ORB-176, which will need it too.

Also in this commit: the ORB-171 CLI test asserted `"password" not in stderr` and Alembic now logs this branch's migration title («nullable password…») beside the CLI's line; the test reads the CLI's own line only.

Full local suite before the review: `4249 passed, 1 failed` (that CLI test, for the reason above). CI on the new head is the record.

PR #87, opened 2026-09-12
feat(web): the Home of a new space sells the assistant and lists the four first steps

slavni96:
## What this changes

The Home of a new space showed two empty charts and said nothing about what to do first; the assistant the landing sells was a «Token» entry at the bottom of the sidebar. Now, above the dashboard of a space that has not been used yet, two panels read from the data and never stored: a wide card («Il CRM che lavora al posto tuo», «Collega l'assistente») that stays until the space has a personal access token, and «Primi passi», four items in the order the product needs them (the fiscal data, the first customer, the first deal or the first hours, the first offer), each a link and each done when the thing exists, with «2 di 4» and a «Nascondi» whose one preference lives in the browser. Six one-row reads (`/api/customers`, `/api/deals`, `/api/time-entries`, `/api/documents`, `/api/emitter`, `/api/tokens`); a read that fails counts as done so the panel never nags because a request broke. The root and any space in use see nothing. Spec `projects/pigrocrm/docs/superpowers/specs/2026-09-12-onboarding-product-led-design.md` §6.7.

## How I verified it

- `pnpm --filter web test` → `103 files, 1099 tests passed`, four of them new in `DashboardPage.test.tsx`: the empty space shows both panels with «0 di 4» and four «da fare»; a customer and an emitter with a VAT number make «2 di 4» and turn the two done steps from links into struck text; everything done plus a token shows neither panel; «Nascondi» removes the steps, keeps the card and writes `pigrocrm.primi-passi.nascosto`.
- `pnpm --filter web lint` clean; `pnpm --filter web build` (vite + `tsc --noEmit`) ok.
- The pair below was taken on two Vite dev servers (this branch and `origin/main`) with every `/api/*` answer mocked by Playwright to the same empty space, 1440×900, `scale: css`, sidebar cropped.

The button goes to `/app/token` (`CONNECT_ASSISTANT_TO` in `newSpace.ts`). ORB-170 is shipping the «Collega un agente» dialog in the sidebar; when the Home can open that dialog the constant becomes the call, one line. `AppShell.tsx` is untouched for the same reason. No new API, table or migration; no sample data (spec §7).

- The six extra requests on every Home load, cached 30 s. They are `limit=1` reads; the alternative, a server-side «attivazione» read, is in the spec's backlog (§7) and would move the cost, not remove it.
- «A failed read counts as done»: a space whose API is down sees the charts' own error and no panel, which is the honest state.

## Screenshots

**1. The Home of an empty space**
The Home before and after: two empty charts, then the assistant card and the four first steps above them

Linear: ORB-174.

### DIFF
--- a/projects/pigrocrm/apps/web/src/features/dashboard/DashboardPage.test.tsx
+++ b/projects/pigrocrm/apps/web/src/features/dashboard/DashboardPage.test.tsx
@@ -7,9 +7,9 @@
  * handed back as a new search object for the route to push into the URL.
  */
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
-import { fireEvent, render, screen } from '@testing-library/react'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { DashboardPage } from './DashboardPage'
 import type { DashboardSearch } from './search'
 import { api } from '@/lib/api'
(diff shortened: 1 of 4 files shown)


slavni96:
Independent review applied in the second commit. Findings and what I did:

1. «A failed read counts as done» was not what the code did: openapi-fetch returns `{error, response}` and never throws, so a 500 read as «da fare». Every read now throws through `toProblem` for any non-2xx except the emitter's 404, and a test with a 500 on `/api/customers` shows the step done.
2. Private keys sat outside every invalidation. The reads now key under `['customers', …]`, `['deals', …]`, `['time-entries', …]`, `['documents', …]`, `['emitter', …]` and `['tokens', userId, …]`, so creating the first customer or minting a token refreshes the panel on the next Home.
3. The hide key was one string for every space on the origin: it is per space prefix and per user now.
4. Roles: a readonly user sees neither panel (`useCanWrite`); a collaboratore sees the fiscal step as text with «Li imposta l'amministratore», not a link to a page they cannot open. Two tests.
5. The five step reads no longer run while the list is hidden (`enabled: !hidden`); only the token is read for the card. Test.
6. Icons are `aria-hidden` and each row carries «Fatto: » or «Da fare: » as screen-reader text; the tests read that text.
7. Position: the panels sit under the tab strip and above the tab body; the spec says «sopra le tab». Left as is: the tabs belong to the header, and a panel above the page title would read as part of the header.
8. `to` is a literal union and `CONNECT_ASSISTANT_TO` an `as const`; the `as never` cast is gone (the paths type-check); the check icon uses `text-foreground`; apostrophes are typographic throughout.
9. The absence test waits on the reads instead of a timer; `afterEach` clears `localStorage`.

Left as is: «the space has a token» is really «this user has a token» (`PatService.list` filters on the actor), so the card shows to each collaborator until they connect their own assistant. That is what the button offers them, so it reads right; the copy says «Collega Claude al tuo spazio», which is true of what they get.

PR #90, opened 2026-09-12
feat(web): «Get started» is its own page and menu entry, and the first login lands there once

slavni96:
## What this changes

ORB-174 put the assistant card and «Primi passi» on top of the Home's dashboard, where they competed with the charts and disappeared on their own. Ivan on PR 87: «spostiamo la card assistente e primi passi in una nuova voce del menu indipendente che si chiamerà get started, di default dopo il primo login rimanda su quella, le volte successive sulla home». Now «Get started» is a page of its own at `/app/get-started`, with a top-level entry right under Home for every role: the assistant card (until this user has a personal access token) and the four first steps, always, with their state and «Fatti tutti» at four of four. The Home no longer mounts the panels. After the first login the Home sends the person to Get started once (a `replace`, so Back does not bounce) when a step or the assistant is still to do, remembered in the browser per space and per user by the page itself; a space with nothing left to do (the root, any space in use) is never sent and remembers that too; the next times the Home stays the Home. The signup's landing, the login and the link entry all reach the Home, so that one redirect covers the three doors. The six one-row reads, the role gates and the error handling are ORB-174's, moved to `features/get-started/`.

## How I verified it

- `pnpm --filter web test` → `107 files, 1118 tests passed` after the review (1114 before it); new: `GetStartedPage.test.tsx` (six: the empty space with the card and four «da fare» and the visit remembered; two steps done; a 500 that counts as done; everything done shows «Fatti tutti» with four ticks and no card; the readonly sees text, no links; the collaboratore gets no link to the fiscal settings), `useFirstVisit.test.tsx` (six: once with `replace` when something is left to do; nothing read once decided; a space in use stays and remembers; nothing while logged out; per user; a storage that throws never redirects). `DashboardPage.test.tsx` loses the panel tests. The sidebar tests are unchanged and green.
- `pnpm --filter web lint` clean; `pnpm --filter web build` (vite + `tsc --noEmit`) ok, the route tree regenerated with `/app/get-started`.
- The frames below come from two Vite dev servers (this branch and `origin/main`) with every `/api/*` answer mocked by Playwright to the same empty space, 1440×900, `scale: css`. The first-login frame is the branch loaded with an empty `localStorage`: the Home redirected to `/app/get-started` on its own.

The «Collega l'assistente» button still goes to `/app/token` (`CONNECT_ASSISTANT_TO`) until ORB-170's dialog lands. No server-side record of the first login: the browser remembers, which is what «le volte successive» means on that device; a new device sees Get started once more. `AppShell.tsx` changes by one entry and the order of the top-level items, nothing else, to stay clear of ORB-170.

- The redirect lives in the Home route, not in the three doors: simpler, and it also catches a person who types `/app/` first. It is a `replace`, so the history holds Get started only.
- «Get started» is shown to a readonly user too, with the steps as text: the page is a menu entry now, and an empty menu entry would be stranger than a list they cannot act on.

## Screenshots

**1. The Home: the panels gone, «Get started» under Home in the sidebar**
The Home before and after: the assistant card and the steps above the charts, then the plain dashboard with a new «Get started» entry under Home

**2. The first login lands on Get started** (a new page: no before)
Get started: the assistant card and the four first steps, the sidebar entry selected

Linear: ORB-180.


slavni96:
Independent review applied in the third commit. Findings and what I did:

1. «First login» was «first time this browser sees the Home», for everybody: every existing user and the root would have been bounced once to a page with four ticks. The Home now reads the six answers only while the browser has not decided yet, redirects when something is left to do (a step or the assistant), and otherwise remembers the decision and stays. Tests for both branches, and that nothing is read once decided.
2. The two storage comments contradicted each other; `hasSeenGetStarted` keeps answering `true` when storage throws (no loop, the entry is one click away) and both comments now say so. Test with `Storage.prototype.getItem` throwing.
3. The redirect runs from an effect after the Home mounted: one dashboard request and a frame on the very first login. Left as is; a `beforeLoad` needs auth in the router context, which the app does not have yet.
4. Spec §6.7 rewritten for the page, and a `DECISIONS.md` row.
5. Tests: the sidebar tests name «Get started» in the column and in the rail, once, right under Home; `GetStartedPage` marks nothing while nobody is logged in.
6. The stale header line and the marking comment (a sidebar visit counts, on purpose) fixed.
7. The two feature commits stay two: the repository merges with a merge commit and keeps its history.

Web after the review: `107 files, 1120 tests`, lint and build clean.

PR #95, opened 2026-09-12
feat(website): PostHog enters through the cookie notice, after the same yes as the pixel

slavni96:
## What this changes

joinorbiters.com measured one thing, whether an ad brought a signup, and nothing else: no page read, no CTA clicked, no count of how many visitors follow «Entra e scaricala». Ivan asked on 2026-09-12 for PostHog on the CRM, the site and every public surface, and for «quanti scaricano la guida e come e cosa cliccano sul sito».

This is the first of five cards (the design for all of them is `docs/design/2026-09-12-posthog-analytics-design.md`, committed here):

- A new workspace package, `shared/analytics` (`@orbiters/analytics`), holds the one PostHog project every surface reports to: the public `phc_` key, the EU hosts, which hostnames are silent (localhost, the test runners) and which are internal (the previews), and the init policy the CRM and the hub will share on ORB-184 and ORB-185 (`initAnalytics`, `identifyUser`, `identifyGroup`, `capture`, `resetUser` on top of `posthog-js`).
- On the site, `consent.js` injects PostHog's loader beside the ChatGPT Ads pixel, from the same accepting branch. Before «Va bene» the browser makes no request to PostHog; a refusal stays final; privacy, termini and pitch carry neither tracker. Pageviews, autocapture (so every click on a CTA is counted with its text and destination), session replay with inputs masked, and one explicit event, `iscrizione_community`, captured after the API answered 201, next to the pixel's `registration_completed`.
- The notice reads «Un cookie di misurazione, per sapere se un annuncio funziona e come usi il sito. Dettagli.» instead of naming only the ad.
- `privacy.html` § Cookie e misurazione says what PostHog records behind the logins (id, email, nome, ruolo, spazio; replay with every field and text masked; legitimate interest; how to opt out) and that on the site it waits for the yes. The old sentence promising «nessuna analitica» inside the CRM would have become false the day ORB-184 ships.
- One `DECISIONS.md` row: a tracker on these surfaces is PostHog, configured from `shared/analytics`; on a public page it enters through `consent.js` and nowhere else; behind a login it identifies the person and the privacy page says so; a personal API key (`phx_`) goes in no file.

## How I verified it

- `pnpm --filter @orbiters/analytics lint && test && exec tsc --noEmit`: 9 tests, clean.
- `pnpm --filter website lint | test | exec tsc --noEmit | build`: 227 unit tests (was 220; the new ones: nothing fetched before a yes for either SDK, both loaders inside the accepting branch, the literals equal the shared exports, privacy and termini carry no key and no SDK URL, the policy names PostHog and its basis, localhost stays silent).
- `pnpm --filter website test:e2e`: 54 passed. The measured pages request nothing foreign on a first visit; after «Va bene» the pixel's host is requested and PostHog's is not, because the preview runs on localhost and PostHog is silent there by design, so CI never writes a visitor into the project production reports to.
- Live, on the built site served under `preview.joinorbiters.test` (a `--host-resolver-rules` mapping to the preview on 4180) in a real Chromium with automation signals hidden (PostHog drops events from `navigator.webdriver` and `HeadlessChrome`, which cost me four runs to find): `array.js` loads, `__loaded` is true, `sessionRecordingStarted()` is true, `POST /e/`, `/i/v0/e/` and `/s/` answer 200, and the PostHog API lists `$pageview`, `live_check` and `$web_vitals` for that distinct id, `$host: preview.joinorbiters.test:4180`, lib 1.430.2.

- `consent.js` carries the PostHog key and hosts as literals, a second copy of `shared/analytics/posthog.ts`. The file runs without a bundler and the DOM test runs it as text, so importing was not an option; `pixel.test.ts` imports the package and compares the three values, which is the same idiom `path-map-plugin.test.ts` uses for nginx.conf.
- `internal_or_test_user_hostname` is not passed at init in either place. Live, with `defaults: '2026-08-30'`, the SDK's own default for that option (`/^(localhost|127\.0\.0\.1)$/`) won over mine; the explicit `setInternalOrTestUser()` call, queued on the stub and made in `browser.ts`, is what the option would have done. The comment in `browser.ts` says so with the version.
- `posthog-js` is pinned to 1.429.5, not 1.430.2: pnpm's release-age rule refused yesterday's build and I did not want to add an exclusion for it. The site loads whatever `array.js` PostHog serves (1.430.2 today) regardless.
- `orbiters.js` sits at 9206 bytes under its 9216-byte ceiling; I shortened two comments rather than raise the ceiling. `?.` is used once, the file already relied on ES2019.
- CI: `shared/analytics/**` was added next to `shared/brand/**` in every web filter of `ci.yml` and `preflight.json`, plus an `analytics-unit` preflight check. The package has no CI job of its own; the two consumers' gates will exercise it once ORB-184 and ORB-185 import it.

The CRM and the hub do not import the package yet (ORB-184, ORB-185). The CRM's MCP server is ORB-186, after ORB-170. The Postgres sources and the dashboards are ORB-187. The hub's own MCP is not instrumented at all: its only user is Ivan.

## Screenshots

**1. The notice on the landing, at 390px**
The cookie notice, before and after

**2. privacy.html § Cookie e misurazione, at 1440px**
The privacy page's cookie section, before and after

Linear: ORB-183.


slavni96:
**Review applied** (independent read-only reviewer, findings ranked by severity), commit `73f02ed5`:

- **Privacy visible date** stayed «9 settembre» while the attribute moved: fixed, and the test now pins `<time datetime="2026-09-12">12 settembre 2026</time>`.
- **Site session replay undisclosed**: the site paragraph now says PostHog «registra la sessione con i campi mascherati».
- **Text masking over-promised for the hub**: the sentence reads «ogni campo è mascherato, e nel CRM anche ogni testo», which is what `browser.ts` does (`maskText` is the CRM's option).
- **IP not named for PostHog** while it was for OpenAI: one clause added.
- **`iscrizione_community` pinned only by grep**: two driven tests in `orbiters.test.ts` (once after the 201; not after a 422; a throwing stub never reaches the person).
- **Preview-internal rule asserted as text**: `__consent.internal(hostname)` is exposed and called in the test.
- **`__SV` comment wrong** (the SDK recognises the stub by `_i`, not `__SV`): corrected, `__SV` kept for fidelity with the official snippet.
- **"the SDK's default won" claim** not reproducible in the 1.429.5 merge code: reworded to the observation (option had no effect live on array.js 1.430.2; config showed the default; events arrived unmarked). The explicit call stands either way.
- `maskTextSelector` omitted rather than `null`; notice copy «Cookie di misurazione» (two cookies now); page comments say three pages; two over-long lines rewrapped; design sentence about vitest aligned with the code; e2e waits one second instead of `networkidle` towards an unreachable host.

Left as is: the `?.` in `orbiters.js`, which ships unchanged under the module target and keeps the file under its 9 KB ceiling; the redundant `capture_pageview`/`autocapture` in `browser.ts`, kept explicit with the test comment corrected; the optional jsdom test that evals the real `array.js` against the stub, which the live check covered.

Unit 229 + 9, e2e 54, lint and tsc clean after the changes.

PR #96, opened 2026-09-12
feat(mcp): every tool call an agent makes is reported to PostHog

slavni96:
## What this changes

`projects/pigrocrm/apps/mcp` is the surface the onboarding spec sells first («collega il tuo assistente»), and nothing recorded whether an agent used it: which tools it calls, how often, how long they take, which fail. Ivan asked for PostHog on every surface on 2026-09-12; the design is `docs/design/2026-09-12-posthog-analytics-design.md` § "The CRM's MCP server".

Now, when `PIGROCRM_POSTHOG_KEY` holds the project key, PostHog's own adapter (`posthog.mcp.instrument`, SDK 7.53.0, which knows `mcp` 2.x and `MCPServer`) wraps the server at the end of `build_server` and emits `$mcp_tool_call`, `$mcp_tools_list`, `$mcp_initialize` and `$mcp_exception` with the tool name, the parameters, the response, the duration and the error flag. Each event carries the actor's user id as `distinct_id`, `ruolo` and `via` as properties, and the installation's slug (or `root`) as the `spazio` group: the same identity the web uses on ORB-184, so a customer created from Claude and one created from the UI are one person in one space. An actor without an id (the fixtures' `mcp` admin) stays anonymous.

- `pigrocrm_mcp/analytics.py` (new): `install`, `build_client`, `identity_for`, `space_group`. Empty key: no client, `install` answers `False`, nothing changes.
- `pigrocrm_mcp/server.py`: one call, last, once every tool and resource is registered.
- `Settings.posthog_key` / `posthog_host` in core's config (values only; core imports nothing from PostHog), forwarded by `docker-compose.yml` with an empty default and documented in `.env.example`.
- `posthog==7.53.0` as a dependency of `apps/mcp` only.

The adapter's intent capture (`context`) is **off**: on, it injects a required `context` argument into every tool's input schema, which `test_mcp_schema.py` and the surface tests pin and every connected agent has already learned.

## How I verified it

- `uv run pytest -q projects/pigrocrm/apps/mcp/tests/test_analytics_instrumentation.py`: 4 passed. Without a key `install` answers `False`; the group is `root` or the slug; the identity is the actor's id with the space; with a key and an observed client, `describe_schema` through an in-process `mcp.Client` produces one `$mcp_tool_call` with `distinct_id` = the user's id, `$mcp_tool_name` = `describe_schema`, `$groups` = `{spazio: studio}`, `$mcp_is_error` = false, and `describe_schema`'s input schema has no `context` property.
- `test_mcp_schema.py` + `test_mcp_surface_coverage.py` + the new file: 465 passed. `test_architecture.py`: 15 passed (core's allowed imports unchanged).
- `uv run mypy`: success, 297 files. `ruff check` and `ruff format --check`: clean.
- The full `apps/mcp` suite was running under `-n 4` when this opened; its count is in the first comment below.
- Not verified live: a real tool call from Claude against production reaching PostHog, because the key is not set on the server yet. Setting `PIGROCRM_POSTHOG_KEY` in the host `.env` and one `docker compose run ... python -m pigrocrm_mcp` is the check, and it is written on ORB-186 as the closing step.

- `identity_for` reads `actor_provider()` on every call rather than once. For the stdio process the provider answers the same actor forever; for ORB-170's HTTP transport (PR #93, open) it is the per-request actor, which is the point. The two branches touch `build_server` at different lines; whichever merges second has a one-line conflict at most.
- The test observes `Client.capture` by monkeypatching it on a real `Posthog(sync_mode=True)` client, and normalises its positional `event` argument. If the SDK moves the event name to a keyword, the recorder already reads both.
- `posthog_host` is a setting although only EU is used, so a self-hosted installation with its own PostHog can point elsewhere without a code change.

The hub's MCP server is not instrumented: its only user is Ivan. No intent capture, no missing-capability feed, no `send_feedback` tool (the adapter offers all three): each changes what an agent sees, and the surface is the contract. Core has no PostHog dependency.

## Screenshots

Nothing a person can see changed: this is a server process with no UI. The section is kept on purpose.

Linear: ORB-186.

### DIFF
--- a/projects/pigrocrm/.env.example
+++ b/projects/pigrocrm/.env.example
@@ -37,6 +37,15 @@ PIGROCRM_HUB_URL=https://joinorbiters.com
 PIGROCRM_RESEND_API_KEY=
 PIGROCRM_MAIL_FROM=PigroCRM <ciao@joinorbiters.com>
 
+# --- Product analytics (ORB-186) -----------------------------------------------------
+# The MCP server reports every tool call to PostHog when this holds the project key: the
+# public `phc_...` of the one project every Orbiters surface writes to, the same value
+# `shared/analytics/posthog.ts` carries for the browsers. It is a project key, not a
+# personal one: it can only write events. Empty, the installation measures nothing,
+# which is the right default for a self-hosted CRM. The host is Cloud EU.
+PIGROCRM_POSTHOG_KEY=
+PIGROCRM_POSTHOG_HOST=https://eu.i.posthog.com
+
 # Why this line exists and is `false` ONLY here:
 # Vite's dev proxy (apps/web/vite.config.ts) serves the dev API over plain
 # http://localhost, without TLS. Chrome and Firefox treat "localhost" as a secure
--- a/projects/pigrocrm/apps/mcp/pyproject.toml
+++ b/projects/pigrocrm/apps/mcp/pyproject.toml
@@ -2,7 +2,11 @@
 name = "pigrocrm-mcp"
 version = "0.1.0"
 requires-python = ">=3.13,<3.14"
-dependencies = ["pigrocrm-core", "mcp==2.0.0"]
+dependencies = [
+    "pigrocrm-core",
+    "mcp==2.0.0",
+    "posthog==7.53.0",
+]
 
 [tool.uv.sources]
 pigrocrm-core = { workspace = true }
--- a/projects/pigrocrm/apps/mcp/src/pigrocrm_mcp/analytics.py
+++ b/projects/pigrocrm/apps/mcp/src/pigrocrm_mcp/analytics.py
@@ -1,0 +1,104 @@
+"""PostHog on the MCP surface: every tool call an agent makes, counted.
+
+The web reports pageviews and the activation events (ORB-184); the MCP is the surface
+the onboarding spec sells first, and until this module nothing recorded whether an
+agent used it. PostHog's own adapter wraps the `MCPServer` and emits `$mcp_tool_call`,
+`$mcp_tools_list`, `$mcp_initialize` and `$mcp_exception` with the tool name, the
+parameters, the response, the duration and the error flag; this module decides three
+things around it.
+
+- **Off by default.** An empty `PIGROCRM_POSTHOG_KEY` builds no client and changes
+  nothing: a self-hosted CRM measures nothing unless its operator says so, and the test
+  suite never sends an event. The key is the public project key
+  (`shared/analytics/posthog.ts` carries the same one for the browsers), read from the
+  environment rather than imported because this process runs in the API image.
+- **The same identity as the web.** The actor's user id is the `distinct_id` and the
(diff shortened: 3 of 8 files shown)


slavni96:
Full `apps/mcp` suite with the instrumentation in place, run twice: under `-n 4 --dist loadfile` it reports 558 passed and 223 errors, every error a fixture-setup collision between workers on this machine (a known race, unrelated to this change: the same errors appear on `origin/main`); run sequentially (`-p no:xdist`) it reports **781 passed in 32.35s**.

slavni96:
**Review applied** (independent read-only reviewer, findings ranked by severity), commits `e8685064` (merge of main, ORB-170) and `418e2a53`:

- **Behind main by ORB-170, compose anchor**: merged `origin/main`; the two variables now sit in the shared `x-api-environment` anchor that both the `api` and the `mcp` service read (`test_compose_environment.py` green), and `http.py`'s `_build` passes `space=slug` so the group is the space, not the installation's own slug.
- **Concurrency canary breaks with a key exported**: reproduced (`barrier.broken`). Cause: the canary puts a ten-party `threading.Barrier` inside `actor_provider` and counts on one read per call from the tool's thread; the identity callback adds a read on the event loop before dispatch, and that read blocked the loop. The tool path is unchanged (the wrapper awaits nothing that serialises; the other eight concurrency tests pass with the key). The canary now builds its server with an empty key and says why; rerun with `PIGROCRM_POSTHOG_KEY` exported: 1 passed.
- **Exception autocapture on by default**: off (`enable_exception_autocapture=False`). A `DomainError` `_guard` raises on purpose is guidance, not a defect; `$mcp_tool_call` keeps `$mcp_is_error` and the message. New test: the agent reads the same sentence for a not-found with the wrapper on and off, one `$mcp_tool_call` with `$mcp_is_error: true`, no `$exception`. The docstring and `.env.example` no longer name a `$mcp_exception` event that does not exist.
- **Flushing at exit, one client per space**: `build_client` now keeps one client per process (keyed by key and host); `analytics.shutdown()` flushes it in the stdio `main`'s `finally` and at the end of the HTTP lifespan, after the spaces stop, in a worker thread since it blocks on the network.
- **Tests**: the no-key test inspects the server `build_server` returned (`get_server_tracking_data(mcp._lowlevel_server) is None`) and an autouse fixture drops a key exported in the shell; a test for the one-client-per-process rule and a double `shutdown()`.
- **mypy on the test**: the `type: ignore` moved with a `_settings()` helper; the test files type-check.
- **Small things**: `identity_for` is typed `Callable[[object, object], UserIdentity | None]`; `disabled=False` dropped; the idempotence note says a second `install` with another client is ignored.

Left as is: the `$identify` per tool call under `stateless_http=True` (each request is its own session; acceptable, and the SDK warns once). Counts after the changes: analytics tests 6 passed; `test_http_transport.py` 20 passed; `test_compose_environment.py` green; `ruff`, `mypy` clean.

PR #97, opened 2026-09-12
docs(repo): the runbook says how PostHog's warehouse reads two databases

slavni96:
## What this changes

PostHog now reads two of this repository's databases (the hub's `orbiters`, and the CRM's `pigrocrm_tenants` registry) through an SSH tunnel to the server, so that signups, applications, members, guide downloads and spaces can be joined to the events the surfaces send. That arrangement existed only on the server and in PostHog's UI. `docs/adding-a-project.md` § 7 now records it: the restricted `posthog` SSH user and its `Match` block, the `posthog_ro` roles and what they may `SELECT`, the two sources and their prefixes, and the one switch that is not documented by PostHog's API: «Require TLS through tunnel» must be off (`ssh_tunnel.require_tls: {"enabled": false}`), or every attempt answers that the database does not support encryption. It also says what a new project with a database has to add.

Design: `docs/design/2026-09-12-posthog-analytics-design.md`.

## How I verified it

The paragraph describes what is on the server as of 2026-09-12 10:00 UTC, done under ORB-187: `id posthog` exists with `/usr/sbin/nologin`; `sshd -t` accepted the `Match` block and `ssh` reloaded; a shell attempt as `posthog` answers «This account is currently not available» (exit 1) while a forward to 55435 and 55432 answers Postgres's `N` to an SSLRequest; both PostHog sources report `Completed` for every selected table. No code changed; `git diff --stat` is one file.

The server address and the two host ports are written in the runbook. They were already there (the port table above the new paragraph and the vhosts), and the address is the one `ssh orbiters` resolves to. No password, no key.

## Screenshots

Nothing a person can see changed: documentation only. The section is kept on purpose.

Linear: ORB-187.

### DIFF
--- a/docs/adding-a-project.md
+++ b/docs/adding-a-project.md
@@ -287,23 +287,33 @@ are. The copy in `/etc/nginx/sites-available/` has certbot's port-443 block on t
 it: **edit that one in place**, with `nginx -t` before the reload. Overwriting it from
 the repository takes TLS away on the spot.
 
-### The tables behind the events: PostHog's warehouse reads two databases
+### What PostHog's warehouse reads
 
-Since 2026-09-12 (ORB-187) PostHog's data warehouse reads the hub's database
+Since 2026-09-12 (ORB-187) PostHog's data warehouse reads the hub's production database
 (`orbiters` on 55435: `signups`, `freelancers`, `companies`, `guide_downloads`,
-`member_logins`, `comments`) and the CRM's registry (`pigrocrm_tenants` on 55432:
-`tenants`), so the events the surfaces send can be joined to the rows behind them. The
-design is `docs/design/2026-09-12-posthog-analytics-design.md`. Both Postgres answer on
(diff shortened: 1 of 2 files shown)


slavni96:
**Review applied** (independent reviewer, docs-only PR), commit `35a877cf`: the key pair was generated for this purpose and its private half is in PostHog's source configuration, so both the runbook and the design doc now say that instead of «a key PostHog generates»; the preview databases are named as not connected; the `authorized_keys` options moved into a code block; the drop-in notes that a `Match` block runs to the end of its file and gives `sshd -T -C user=posthog` as the check; nologin, not `ForceCommand`, is credited with the «not available» message; and a «Keeping it working» list covers a new table's GRANT, a new port, and rotating the key or a role password. Confirmed by the reviewer against `man sshd`/`sshd_config`: `restrict` then `port-forwarding`, comma-separated `permitopen`, space-separated `PermitOpen`, and `ForceCommand` not limiting a `-N` forward. Left as is: the heading style follows the reviewer's suggestion; nothing else.

PR #99, opened 2026-09-12
feat(web): report pageviews, sessions and the funnel to PostHog

slavni96:
## What this changes

Nobody could see how a space is used: the activation metric of the onboarding spec (first customer, activated within seven days) had no events behind it, and the CRM sent nothing to PostHog at all. Now the SPA initialises the shared PostHog project from `@orbiters/analytics` once in `main.tsx` with `maskText: true`, so pageviews, autocapture and session replay arrive with every text and every input masked: a recording shows where a person clicks and stops, never an invoice amount or a customer's name.

The funnel is read off the API client rather than written into the features. `main.tsx` registers one `openapi-fetch` middleware from `lib/analytics.ts` on the client in `lib/api.ts`, with a closed table of nine successful calls to named events: `POST /api/tenants/` is `spazio_creato`, `POST /api/auth/entra` is `entrato_con_link`, `POST /api/customers` is `cliente_creato`, `POST /api/deals` is `deal_creato`, `POST /api/documents` is `documento_creato`, `POST /api/time-entries` is `ore_registrate`, `POST /api/invoices` is `fattura_emessa`, `POST /api/tokens` is `assistente_collegato`, `PUT /api/emitter` is `profilo_emittente_salvato`. The space's prefix (`/<slug>/api/...`) and any trailing slash are removed before the lookup; a 4xx or 5xx counts nothing; nothing outside the table emits.

`AuthProvider` calls `identifyUser(id, { email, nome, ruolo })` and `identifyGroup('spazio', slug)` (`root` for the installation with no prefix) when `me` answers a person, and `resetUser()` on logout, in the `finally`, before `window.location.assign`. The web image copies `shared/analytics/package.json` before `pnpm install` and the directory before the build, next to the two `shared/brand` lines, or the workspace link cannot resolve inside the image.

**Adjacent:** ORB-183 (the site) is merged and already rewrote the privacy page for this; ORB-185 (the hub) is the same package on the other SPA, in its own worktree; ORB-186 (the MCP server) is server-side and its own card.

## How I verified it

From the repository root:

- `pnpm install --frozen-lockfile`: `Already up to date`, exit 0 (the lockfile gained the one `link:../../../../shared/analytics` entry).
- `pnpm --filter web lint`: `eslint .`, no output, exit 0.
- `pnpm --filter web test`: `Test Files 109 passed (109)`, `Tests 1153 passed (1153)`. New: `src/lib/analytics.test.ts` (23 tests: each of the nine rows emits exactly its event on a 2xx; 400/401/409/422/500/503 emit nothing; a read, an update, `POST /api/documents/from-template`, `POST /api/invoices/import`, `POST /api/emitter` and the session endpoints emit nothing; the prefix is stripped under `/studio` and another slug is not; trailing slashes are ignored; one run through a real `createClient` with a stubbed `fetch` proves the request reaches the middleware as `openapi-fetch` builds it; `identifySession` names `root` and `studio`). `auth.test.tsx` gained four: identify and group once when the session appears and not again after the thirty-second poll answers the same person; nobody identified on a 401; `resetUser` before `assign('/app/login')` on logout; the same when the server answered 500.
- `pnpm --filter web build`: `vite build` built in 788 ms and `tsc --noEmit` passed, exit 0.
- `pnpm --filter @orbiters/analytics test`: `Test Files 2 passed (2)`, `Tests 9 passed (9)`.

Not verified: the image build (`docker compose build web`) was not run here; the two `COPY` lines mirror the `shared/brand` pair exactly, and the trunk's image job will prove it. Not verified either: an event arriving in PostHog, which needs the preview deploy (the hostname rule makes localhost silent by design); that check is written on the card after the deploy, as the design's Verification section says.

- The identify effect depends on `[userId, email, nome, ruolo]` rather than on `[user?.id]` alone. `user` is a new object on every thirty-second poll, so the object cannot be the key; the four primitives are, and `react-hooks/exhaustive-deps` stays clean without a disable. The consequence is that a role change re-identifies the person with the new `ruolo`, which is what a person property should do. The test proves the poll alone does not re-identify.
- Under a space, a bare `/api/customers` reaching the space's client would still match the table, because there is no prefix to strip. The client's `baseUrl` makes that request impossible, so I did not add a rule for it; another slug's prefix is left in place and matches nothing (tested).
- `identifySession` sends email and name, as Ivan decided in the design (`What was decided, and by whom`). If that decision moves, the one place to change is `lib/analytics.ts`.
- The middleware is registered from `main.tsx`, not inside `lib/api.ts`, so every test that stubs `fetch` around the client keeps seeing a plain typed client.

- No feature file mentions analytics: the table in `lib/analytics.ts` is the whole contract, as the design says.
- The MCP server is ORB-186 (server-side, `posthog` Python SDK, after ORB-170), not this card.
- The privacy page and `docs/design/DECISIONS.md` were rewritten by ORB-183 (#95) and already describe what the CRM records.
- `spazio_creato` carries no group: the space does not exist until the API answers, and PostHog computes the funnel per person from the `spazio_creato` timestamp (design, «The CRM»). Confirmed by the review as design-level, not a gap here.
- `POST /api/documents/from-template` and `POST /api/invoices/import` are not events: the table names `POST /api/documents` and `POST /api/invoices` and nothing else, and widening it is a design change rather than an implementation one.

## Screenshots

Nothing a person can see changed: PostHog has no UI inside the CRM, there is no banner (the person is signed in and the privacy page says what is recorded), and the masking only affects the recording PostHog stores. The section is kept on purpose.

Linear: ORB-184.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -167,6 +167,9 @@ importers:
       '@dnd-kit/core':
         specifier: 6.3.1
         version: 6.3.1(react-dom@19.2.8(react@19.2.8))(react@19.2.8)
+      '@orbiters/analytics':
+        specifier: workspace:*
+        version: link:../../../../shared/analytics
       '@orbiters/brand':
         specifier: workspace:*
         version: link:../../../../shared/brand
--- a/projects/pigrocrm/Dockerfile.web
+++ b/projects/pigrocrm/Dockerfile.web
@@ -14,9 +14,11 @@ WORKDIR /app
 COPY package.json pnpm-workspace.yaml pnpm-lock.yaml ./
 COPY projects/pigrocrm/apps/web/package.json projects/pigrocrm/apps/web/
 COPY shared/brand/package.json shared/brand/
+COPY shared/analytics/package.json shared/analytics/
 RUN pnpm install --frozen-lockfile --filter web...
 
 COPY shared/brand ./shared/brand
+COPY shared/analytics ./shared/analytics
 COPY projects/pigrocrm/apps/web ./projects/pigrocrm/apps/web
 RUN pnpm --filter web build
 
--- a/projects/pigrocrm/apps/web/package.json
+++ b/projects/pigrocrm/apps/web/package.json
@@ -16,6 +16,7 @@
   },
   "dependencies": {
     "@dnd-kit/core": "6.3.1",
+    "@orbiters/analytics": "workspace:*",
     "@orbiters/brand": "workspace:*",
     "@tailwindcss/vite": "4.3.3",
     "@tanstack/react-query": "5.101.4",
--- a/projects/pigrocrm/apps/web/src/lib/analytics.test.ts
+++ b/projects/pigrocrm/apps/web/src/lib/analytics.test.ts
@@ -1,0 +1,143 @@
+import createClient from 'openapi-fetch'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+vi.mock('@orbiters/analytics/browser', () => ({
+  capture: vi.fn(),
+  identifyGroup: vi.fn(),
+  identifyUser: vi.fn(),
+  resetUser: vi.fn(),
+}))
+
+import { capture, identifyGroup, identifyUser } from '@orbiters/analytics/browser'
+import type { paths } from './api-types'
+import { analyticsMiddleware, eventFor, identifySession, installAnalyticsMiddleware } from './analytics'
+
+// Absolute, because a `Request` needs a base to resolve against and the middleware
+// reads the pathname off `request.url`, the way it does in a browser where
+// `openapi-fetch` resolved `/<slug>/api/...` against the page's origin.
+const ORIGIN = 'http://pigrocrm.test'
+
+/** One call through the middleware alone: the request as the client would have built
+ *  it, the response as the server answered. */
+async function answered(method: string, path: string, status: number, prefix = '') {
+  const middleware = analyticsMiddleware(prefix)
+  await middleware.onResponse?.({
+    request: new Request(`${ORIGIN}${path}`, { method }),
+    response: new Response(null, { status }),
+  } as never)
+}
+
+beforeEach(() => {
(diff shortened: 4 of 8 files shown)


slavni96:
**Review applied** in `76860453` (the first commit is `c02806fd` after the amend).

1. **Subject over 72 characters.** Amended the first commit to `feat(web): report pageviews, sessions and the funnel to PostHog` (63), body kept, force-pushed with lease, PR title updated to the same.
2. **The person was never forgotten when the session vanished without a logout.** `AuthProvider` keeps the last identified id in a `useRef`; on the non-null to null transition it calls `forgetSession()` once, before the router shows the login page. An anonymous visitor's first `null` does not reset, so the login pageview still merges into the person they become. `forgetSession` sits next to `identifySession` in `lib/analytics.ts`. New test: `me` answers ADMIN, the next poll answers 401, `resetUser` exactly once and not again after another minute; the visitor test now also asserts no reset.
3. **"Answers the same person as a new object" was false.** The mock is `mockImplementation(() => Promise.resolve(ok({ ...ADMIN })))`, a fresh object per poll; the comment in `auth.tsx` now says the effect keys are the properties PostHog receives (a role change re-identifies, a new object does not).
4. **Logout `finally`.** The forgetting is in a `try { } catch { }` with the reason (leaving matters more than forgetting), so `queryClient.clear()` and `window.location.assign` always run.
5. **`installAnalyticsMiddleware` wrapper.** Dropped; `main.tsx` calls `api.use(analyticsMiddleware())` and the tests call `client.use(...)` the same way.

**Left as is:** `spazio_creato` carries no group. The space does not exist until the API answers, and PostHog computes the funnel per person from that timestamp; design-level, now written under "What did not change, on purpose" in the body.

**Verified before pushing, from the repository root:** `pnpm --filter web lint` clean, exit 0. `pnpm --filter web test`: Test Files 109 passed (109), Tests 1155 passed (1155). `pnpm --filter web build`: vite built in 648 ms, `tsc --noEmit` passed, exit 0.

PR #100, opened 2026-09-12
feat(hub): the wizards, the member area and the admin report to PostHog

slavni96:
## What this changes

The hub recorded nothing about who starts a wizard, where they leave it, who finishes, and who downloads the guide, so «where do they leave» (ORB-122) and the guide's reach were guesses. Now the hub SPA initialises the one shared PostHog project (`@orbiters/analytics/browser`, design `docs/design/2026-09-12-posthog-analytics-design.md` § The hub) once in `main.tsx`, with inputs masked in a recording and text visible, and reports:

- `wizard_iniziato` once per mount of a wizard, with `tipo: 'freelance' | 'azienda'` and `perk` when the URL carries `?perk=...` (`?perk=guida` is how the site's guide section arrives). It goes out from the engine's first step call, so it always precedes `wizard_passo { passo: 0 }`.
- `wizard_passo` with `passo` (0-based) and `passi` (the total) every time the step on screen changes: the first step on mount, «Avanti», «Indietro», «Modifica» from the review, the review itself as `passo === passi`, and the jump back to the step a 422 names. A funnel takes the first in-order occurrence of each step; the design says so since the review.
- `wizard_completato` only after the API accepted the submission, in the line before the page navigates to `/grazie`. A refused submission reports nothing completed.
- `guida_scaricata` on the click of «Scarica la guida» in the member area; the link and the download are as they were.
- Identity: the member as `identifyUser(id, { email, nome })` in the `/io` guard, once the profile is known and once only, however often the profile is refetched; the admin as `identifyUser(id, { email, nome, ruolo: 'admin' })` in the admin frame; `resetUser()` in both logout hooks before `window.location.assign`.

The engine (`wizard/Wizard.tsx`) gained one optional, memoised `onStep(index, total)` prop; the event names live in `lib/analytics.ts`, which the two pages, the guard and the admin frame call. `projects/hub/Dockerfile.web` copies `shared/analytics` beside `shared/brand` in both places, or `pnpm install --filter hub...` cannot resolve the workspace link. `pnpm-lock.yaml` carries the new link. posthog-js ships in a chunk of its own (`vite.config.ts`), so a release of the hub does not make every browser fetch it again.

One shared file changed, after the review: `shared/analytics/browser.ts` wraps `posthog.init` in a `try`, because both SPAs call `initAnalytics` before their first render and a blank page is worse than no analytics. The CRM gets the same protection.

## How I verified it

From the repository root, at the review commit:

- `pnpm install --frozen-lockfile`: `Already up to date`.
- `pnpm --filter hub lint`: clean, no output after `eslint .`.
- `pnpm --filter hub test`: `Test Files 21 passed (21)`, `Tests 102 passed (102)` (87 on `origin/main`; the 15 new ones cover the events above, the order of the first two, identify once across a refetch, reset on logout, the perk cap).
- `pnpm --filter hub build`: vite built `posthog-Cupzi_mQ.js 280.67 kB │ gzip 92.77 kB` and `index-Blbynt1-.js 452.43 kB │ gzip 138.53 kB` with no chunk warning; `tsc --noEmit` exit 0.
- `pnpm --filter @orbiters/analytics test`: `Test Files 2 passed (2)`, `Tests 10 passed (10)`.
- `cd projects/hub && POSTGRES_PASSWORD=build-only-never-used ORBITERS_DATA_DIR=./build-only-never-mounted docker compose -p orbiters-orb185 build web` (first commit): `web Built`, exit 0, image 77 MB, removed afterwards.

Not verified: an event arriving in PostHog's activity view. Analytics is silent on `localhost` by design, so the live check is the preview after the merge, written on the card as the design's verification section asks.

- `wizard_passo` fires for the first step on mount as well, so `wizard_iniziato` and `wizard_passo { passo: 0 }` arrive together, in that order. I chose uniformity (every step reached has a `wizard_passo`, a breakdown by `passo` needs no special case) over saving one event; it could have gone the other way.
- `perk` is sent as it arrived (trimmed, capped at 200 characters), the way `readUtm` keeps a UTM, while the banner's `readPerk` recognises only `guida` on top of the same parser. A funnel by perk shows the value that was in the URL.
- The chunk split uses `build.rolldownOptions.output.codeSplitting`, Vite 8's rolldown option; `advancedChunks` is its deprecated alias in rolldown 1.2.3.
- The two «Esci» tests each print one `Not implemented: navigation to another Document` line: jsdom cannot navigate and says so from its own console, which no spy in a test can reach (checked against jsdom 30's `virtual-console.js` and Vitest's environment). I preferred that line to a navigation seam in production code that only tests would use.

The hub's MCP server is not instrumented: its only user is Ivan (design § The hub). The site's own pages and the privacy page are ORB-183, already merged. No `maskText` here: the design masks inputs only in the hub. No event on the admin login form itself: the admin is identified once the frame knows the session, which is the same moment. posthog-js is imported as the shared package ships it, not lazily: a lazy import would lose the mount events.

## Screenshots

Nothing a person can see changed: every line here is an event or an identify call. The section is kept on purpose.

Linear: ORB-185.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -73,6 +73,9 @@ importers:
 
   projects/hub/apps/web:
     dependencies:
+      '@orbiters/analytics':
+        specifier: workspace:*
+        version: link:../../../../shared/analytics
       '@orbiters/brand':
         specifier: workspace:*
         version: link:../../../../shared/brand
--- a/projects/hub/Dockerfile.web
+++ b/projects/hub/Dockerfile.web
@@ -4,13 +4,15 @@ RUN corepack enable && corepack prepare pnpm@11.6.0 --activate
 WORKDIR /app
 
 # The build context is the repository root: the pnpm workspace root, the single
-# lockfile and `shared/brand` live above this project.
+# lockfile and the two `shared/` packages the hub imports live above this project.
 COPY package.json pnpm-workspace.yaml pnpm-lock.yaml ./
 COPY projects/hub/apps/web/package.json projects/hub/apps/web/
 COPY shared/brand/package.json shared/brand/
+COPY shared/analytics/package.json shared/analytics/
 RUN pnpm install --frozen-lockfile --filter hub...
 
 COPY shared/brand ./shared/brand
+COPY shared/analytics ./shared/analytics
 COPY projects/hub/apps/web ./projects/hub/apps/web
 RUN pnpm --filter hub build
 
--- a/projects/hub/apps/web/package.json
+++ b/projects/hub/apps/web/package.json
@@ -13,6 +13,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@orbiters/analytics": "workspace:*",
     "@orbiters/brand": "workspace:*",
     "@tailwindcss/vite": "4.3.3",
     "@tanstack/react-query": "5.101.4",
--- a/projects/hub/apps/web/src/lib/analytics.test.ts
+++ b/projects/hub/apps/web/src/lib/analytics.test.ts
@@ -1,0 +1,13 @@
+import { describe, expect, it } from 'vitest'
+import { readPerkParam } from './analytics'
+
+describe('readPerkParam', () => {
+  it('reads the perk as it arrived, and nothing for a blank or absent one', () => {
+    expect(readPerkParam('?perk=guida')).toBe('guida')
+    expect(readPerkParam('?utm_source=linkedin&perk=guida')).toBe('guida')
+    // Unlike the banner, the funnel keeps a value it does not recognise: it happened.
+    expect(readPerkParam('?perk=crm')).toBe('crm')
+    expect(readPerkParam('?perk=%20')).toBeNull()
+    expect(readPerkParam('')).toBeNull()
+  })
+})
--- a/projects/hub/apps/web/src/lib/analytics.ts
+++ b/projects/hub/apps/web/src/lib/analytics.ts
@@ -1,0 +1,70 @@
+import { capture, identifyUser } from '@orbiters/analytics/browser'
+import { useCallback, useEffect, useMemo, useRef } from 'react'
+import type { Admin, MemberProfile } from './api'
+
+/**
+ * What the hub tells PostHog beyond pageviews and autocapture (ORB-185, design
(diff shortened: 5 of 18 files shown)


slavni96:
**Review applied** in `34976945`, six findings, nothing left as it was.

1. **Order of the first two events.** `wizard_iniziato` now goes out from the first `onStep` call, behind the same ref, and the separate effect is gone: a child's effects commit before its parent's, so the old shape put `wizard_passo { passo: 0 }` first. The test asserts the first two captures in that order.
2. **The once-per-mount guard is exercised.** The `wizard_iniziato` test mounts under `StrictMode`, so React 19's simulated remount is what the ref answers. Two new tests refetch the same member and the same admin (`invalidateQueries` + `waitFor`) and see one identify each.
3. **One perk parser.** `readPerk` is `readPerkParam(search) === 'guida' ? 'guida' : null`; a test feeds 201 characters and gets 200 back.
4. **posthog-js in its own chunk**, via `build.rolldownOptions.output.codeSplitting` (`advancedChunks` is the deprecated alias in rolldown 1.2.3). `pnpm --filter hub build` now prints `posthog-Cupzi_mQ.js 280.67 kB │ gzip 92.77 kB` and `index-Blbynt1-.js 452.43 kB │ gzip 138.53 kB`; the 500 kB warning is gone and the app's chunk is back to the size it had on `origin/main` (450.83 kB).
5. **A throwing SDK cannot blank the SPA.** In `shared/analytics/browser.ts` the `posthog.init` and `setInternalOrTestUser` block is inside a `try`; on a throw `initAnalytics` answers `false` and every wrapper stays a no-op. `browser.test.ts` has the case. This is the one edit to a shared file in this PR; the CRM calls the same function and gets the same protection.
6. **The design doc** (§ The hub) says `wizard_passo` fires for every step shown, backward moves included, so a funnel takes the first in-order occurrence and a «last passo per person» reading is not «where they left».

Counts, from the repository root at `34976945`: `pnpm install --frozen-lockfile` already up to date; `pnpm --filter hub lint` clean; `pnpm --filter hub test` 21 files, **102 tests** passed (99 before); `pnpm --filter hub build` vite built, `tsc --noEmit` exit 0; `pnpm --filter @orbiters/analytics test` 2 files, **10 tests** passed (9 before).

PR #106, opened 2026-09-14
feat(brand): the name is drawn in a face of its own, as outlines

fiorelorenzo:
## What this changes

ORB-193 renamed the platform to Rebase and left the word itself undrawn: «Orbiters» was plain text in `--font-sans` at weight 500, so «rebase» would have inherited a body step as its logotype. Lorenzo asked for the UI and the four tiles to stay exactly as they are, for the word to be set in a face of its own, and for the result to be usable as a picture, «una cosa semplice che poi possiamo usare in immagini, copertina di linkedin ecc».

I rendered six OFL candidates on the real palette at the two sizes the name is actually read at, 18px in the header chip and 64px on the deck (the sheet is attached to ORB-197), and the pick is Space Grotesk 700 at -0.035em: it keeps one grotesque voice with Outfit, while its flat terminals and single-storey `a` hold the square system where Outfit's round geometry reads soft at header size.

It ships as outlines, not as a second webfont. `shared/brand` gains `wordmark.svg` (the word, 2981x714) and `lockup.svg` (the four tiles at cap height, then the word, 3943x714), plus `wordmark-paper.svg` and `lockup-paper.svg` for the dark bands the landing already alternates. All four are generated by `shared/brand/tools/build-wordmark.py`, which fetches the variable font from google/fonts, refuses it unless it checksums to `FONT_SHA256`, instances the weight, shapes the word through HarfBuzz so the kerning is the font's own rather than an average letter-spacing, and reads its colours out of `palette.css`. The source font is not committed: nothing serves it, and the artefacts are the four files.

So no page loads a display face to read the name, and the one string a visitor uses to tell they are on the right site cannot arrive late and reflow. The tile is half the cap height and the gap three quarters of the tile, so a consumer sets one width and the pair holds; none of the four carries a `width` or `height`, because the same file is the 18px chip and a 1584px LinkedIn cover.

## How I verified it

- `preflight`: 8 checks selected by the diff, all green in 66s (`website-lint`, `website-unit`, `website-build`, `website-types`, `hub-web-unit`, `hub-web-build`, `website-e2e`, `website-image`).
- `pnpm vitest run src/landing-style.test.ts` in `projects/website`: 21 passed, 4 of them new. I checked they bite: repainting the gold tile `#ffd700` in `lockup.svg` fails the tile-order test with `expected [ '#011936', '#ffd700', …(2) ] to deeply equal [ '#011936', '#f9dc5c', …(2) ]`.
- `uv run ruff check` and `uv run ruff format --check` on the generator: clean.
- I ran the generator twice from a clean state and got byte-identical SVGs, and rendered all four in headless Chrome at 18px, 64px and 110px, on paper and on Prussian Blue (see Screenshots).

Three things, in order of how much I would argue about them.

The generator is the only Python in the repository that no gate runs: `.github/preflight.json` lints `projects/pigrocrm` and `projects/hub`, and `uv run mypy` names each project's `src` plus `projects/hub/tools`, so `shared/brand/tools` is covered by neither. Its dependencies are PEP 723 inline exactly because it should not be in the workspace resolution for a script that runs by hand when the face changes, which also means mypy cannot see fontTools or uharfbuzz to check it. I ran ruff on it by hand. Adding a `shared/brand/tools` ruff check to preflight costs one entry and would fire only when that path changes: I would rather propose that than install it, since it lands on Ivan's pushes too.

The paper cuts are two more files rather than a CSS override. A surface that uses the asset as an `` or a `background-image` cannot recolour it, which is most of what these are for, so the alternative was inlining the SVG everywhere it appears on a dark band.

The tests live in `projects/website/src/landing-style.test.ts` although the files they assert are in `shared/brand`. That follows what is already there: that suite imports `@orbiters/brand/mark` and holds the website's `.glyph` to `BRAND_TILES`, and the app holds its own copy in `BrandMark.test.tsx`. The lockup is the fourth surface that draws the mark and the first that cannot use a `var()`, so its hexes are literal and pinned to `palette.css` there.

The four tiles, at any scale, on any surface, including the social profiles: the mark survives the rename untouched, which is what Lorenzo asked for. `--font-sans` stays Outfit everywhere, and no page or component is wired to the new assets in this PR. Every «Orbiters» a person still reads on the site, the hub, the CRM, the deck, the guide and the README is ORB-194, which is Ivan's and which this unblocks. Code identifiers (`orbiters_core`, `ORBITERS_*`, the `/orbiters` route, `@orbiters/brand` itself) stay as they are, the same boundary ORB-194 draws.

## Screenshots

Attached, `pr-wordmark-fit.png`: the live chip today in Outfit 500 against the asset at the same 18px, then `lockup.svg` at 64px and `wordmark.svg` at 110px, then `lockup-paper.svg` on Prussian Blue, where the two ink tiles are the ground by the mark's own rule (`mark.ts`). No page changed in this PR, so there is no before-and-after of a surface to pair: the rendered assets are what there is to look at, and the surfaces move under ORB-194.

Linear: ORB-197.

pr-wordmark-fit


fiorelorenzo:
**Review applied**, twelve findings, all read, in `05fc542`.

Changed the code, seven of them:

- **The lockup viewBox was the word's, not the union**, and the reviewer demonstrated the consequence: set `WORD` to a word with no ascender and the top tiles land above the box while the script prints success. It only fitted for «rebase» because Space Grotesk's cap height equals the ink height of its `b`. The box is now the union of the word and a mark standing on the baseline; `WORD = "orca"` renders 714 tall with every tile inside, where before it was 524 with `y="-190"`.
- **`FONT_URL` tracked `google/fonts@main`** while `FONT_SHA256` did not move with it, so the next Space Grotesk revision would have left no way to fetch the bytes these SVGs were drawn from. Pinned to `2861cb7`, which is the commit that last touched the blob and checksums identically.
- The fetch takes `timeout=30`: it was the one path in the script that could hang rather than fail.
- `width` and `height` are taken from the rounded extremes, not from the rounded difference, which removes a class of off-by-a-tenth viewBox even though the committed output never hit it.
- The tiles carry `shape-rendering="crispEdges"`, on the group and not the root, the way `orbiters-logo.svg` already does it on its own field of tiles.
- `--check` redraws into memory and fails on any file that is not what the script draws. That is what the README's "never hand-edit an SVG" was missing, short of the lockfile-and-preflight machinery `build_guide_pdf.py` carries.
- The dead `# noqa: S310` is gone: the root `ruff.toml` selects no `S` rules, so it was suppressing nothing.

Two tests added, both checked against a mutation: the geometry the README sells (tile at half the cap height, gap at three quarters of the tile, mark on the baseline, which halving the gap fails with 875 against 962.5), and that all four files carry the same word, which is what a partial regeneration breaks.

Two documentation defects fixed: the README claimed the internal `aria-label` and `<title>` carry the name, which holds only when the SVG is inlined, so an `` consumer is now told to write its own `alt`; and it called literal hex "the one place in the system where that is right", which `orbiters-logo.svg` has been disproving since before this branch. The same wrong claim in the test comment is gone.

Left as it was, one: the reviewer suggested widening the lockup's gap toward the header chip's proportion (19.2px beside a 12px mark, against 0.375 of the cap height here). I documented it instead of matching it. The chip's spacing is two CSS margins around a text node, and this asset is what replaces that text node under ORB-194, so matching a number that is about to stop existing would fix the wrong side. The generator says so where the gap is computed.

`preflight` again after the changes: 8 checks, all green in 46s.

PR #107, opened 2026-09-14
feat(website): a link to any page shares with the mark, not with somebody's face

slavni96:
## What this changes

Every page's head named a title and a description and no image, so a client that builds a link preview fell back to the first large picture in the markup. On the landing that is the voices section, and its first face is Ivan's: he pasted the site's link and got his own photograph next to the bare domain.

The site now has one share card, 1200x630, the four tiles on the ink ground with the wordmark and the claim. It is drawn by `projects/website/scripts/share-card.mjs`, which reads the palette out of `shared/brand/palette.css` and the typeface out of `shared/brand/fonts`, and it is committed as a static PNG, as the issue asked. All six pages carry `og:site_name`, `og:image` with its type, dimensions and alt, and `twitter:card` set to `summary_large_image`.

Two details worth knowing:

- The card lives at `src/public/assets/share-card.png`, a directory deeper than a public file usually would, because `deploy/nginx.conf` serves the six pages and `/assets/` and 404s everything else. A card at the document root would have answered 200 in preview and 404 in production. `.gitignore`'s screenshot rule gets the matching exception.
- `pitch.html` had no Open Graph at all. It is `noindex`, but that is about crawling, not about sharing, and a deck exists to be sent by link, so it gets the same head as the rest.

## How I verified it

- `pnpm --filter website test`: 251 passed in 13 files (243 before). The new `src/share-card.test.ts` walks all six pages, not the four `landing-pages.test.ts` covers: each has the five tags with the exact values, an alt long enough to be a sentence, and a title and description to go with the image. It also checks the PNG's own IHDR bytes say 1200x630, that its size is far below the 600 kB above which some clients give up, that `route('/assets/share-card.png')` is a path the map serves, and that the generator contains no hex besides white.
- `pnpm --filter website lint`, `pnpm --filter website build`, `tsc --noEmit`: clean, and `dist/assets/share-card.png` is in the build output.
- `pnpm --filter website test:e2e`: 54 passed.
- Over HTTP against `vite preview`: `/assets/share-card.png` answers 200 as `image/png` at 29750 bytes; each of `/`, `/pigrocrm`, `/orbiters`, `/pitch`, `/privacy`, `/termini` answers 200 and carries the absolute `og:image`; `/share-card.png` at the root answers 404, which is what production would have done to a card placed there.

- The image URL is absolute on `https://letsrebase.com`, the domain main moved to today (ORB-193). It is not derived from anything, so if the public name moves again these six values move with it. A canonical address and `og:url` are ORB-111 and are deliberately not here.
- The wordmark says «Orbiters», which is what every page still prints, while the domain already says letsrebase.com. The copy's own rename is ORB-194, named by `website-v0.10.0`'s tag message; when it lands, the card is one word in the generator and `pnpm --filter website build:share-card`.
- While writing the head comments I found a trap: `landing-pages.test.ts` refuses the string «SLA» anywhere in `termini.html` with no word boundary, so a comment naming a certain chat application fails a test about subscriptions. I reworded my comment rather than loosen somebody else's assertion, and left a note in the new test.

No per-page image and no build-time compositing, as the issue decided: one static asset, one generator run by hand. The hub's own pages are not touched.

## Screenshots

The card itself is the change, and it is in the diff as `projects/website/src/public/assets/share-card.png`; it is attached below as well. A before-and-after pair of the real preview cannot be captured from this repository: the «before» is the preview LinkedIn built from the live site, which Ivan reported with a screenshot, and the «after» is what a client will build once this is deployed, which is a third party's render of the tags. The nearest honest evidence is the served file and the six heads, both checked over HTTP above; the preview itself is worth re-pasting the link to see once the tag ships.

Linear: ORB-112.

share-card-attach


slavni96:
## Independent review, applied in 3550365

Twelve findings from a read-only reviewer; no blocker, and four of them changed the code.

1. **The card's name now carries a number** (medium). The reviewer pointed out that LinkedIn, Facebook and WhatsApp cache a scraped card per URL for days, so a redraw under the same name keeps sharing as the old card, and the rename of the copy to Rebase (ORB-194) is already due. It is `share-card-1.png`, and the number lives in the generator, the test and the six heads together.
2. **The path assertion proved nothing** (medium). `route('/assets/…')` answers `{kind: 'file'}` for any path with a dot in it, so it would have passed for a card that exists nowhere. The unit test now reads the `location /assets/` block out of `deploy/nginx.conf`, and a new e2e case asks the real server for the card and checks it comes back `image/png`.
3. **The `meta()` regex could bridge two tags** (medium-low): `[\s\S]*?` crossed a `>`, so a tag with no `content` of its own would have captured the next one's. Bounded to `[^>]*`.
4. **The mark was restated with nothing asserting it** (low). `shared/brand/mark.ts` exists so the surfaces that draw the four tiles cannot drift, and this is the fourth. The test now ties the generator's tile order to `BRAND_TILES`, with the ink-to-white substitution the dark ground needs.
5. **The generator was not linted** (low): eslint's glob here is `{ts,js}` and the file was `.mjs`, so 83 rules were passing it by. It is `scripts/share-card.js`, `type: module` keeps it ESM, and eslint now reports those 83 rules on it.
6. It takes an output path, so the committed PNG can be compared with what it draws today without being overwritten, and it prints what drew the file: `share-card-1.png: 1200x630, chromium 151.0.7922.34 on darwin`.
7. **X reads `twitter:image:alt`**, not the Open Graph one, so both are declared; the alt is pinned to the string rather than to a length.
8. The `.gitignore` exception is narrowed to this project's directory.
9. The head comment no longer claims a face fallback on the three pages that carry no image at all.

Left as they were, with the reason:

- **`noindex` stays on the pitch.** The reviewer notes X's crawler honours it and cards on noindex pages are a recurring report, unlike LinkedIn, WhatsApp, Facebook and iMessage. The fix if it fails is to scope the directive to `googlebot`, and the check is X's own card validator once this is live: worth doing then rather than guessing now.
- **The redraw is versioned rather than documented**, so no re-scrape step is needed on the platforms.

The reviewer also confirmed, by reading rather than assuming, that `letsrebase.com` is the right host (`deploy/letsrebase.conf`, `server_name letsrebase.com`), that the card reaches production (`publicDir` to `dist/assets`, `.dockerignore` does not exclude `public/`, the image build carries it, preflight's `website-image` is selected by this diff), and one thing worth repeating here: **preview cannot prove this change**, because `deploy/preview.letsrebase.conf` sets `X-Robots-Tag: noindex, nofollow` and a `Disallow: /` robots.txt, so no real client will build a card from the preview host. The proof is the served file plus re-pasting the link once production has it.

After: 253 unit tests in 13 files, 55 e2e (54 before), lint, build and `tsc --noEmit` clean.

PR #111, opened 2026-09-15
feat(hub): the CV is asked for and no longer demanded

slavni96:
## What this changes

«Il tuo CV» was the one step of the freelancer wizard nobody could pass without a PDF to hand: «Avanti» answered «Serve il CV, in PDF.» and the form ended there. The member area then asked for one again before it would save anything on a card that had none, so the same wall stood one step later. Somebody who wanted into the community and had no CV ready left with nothing.

The CV is now optional where it is asked:

- The wizard step carries «(facoltativo)» and its hint says the file can arrive later from the area. A file that *is* attached is checked exactly as before, size and type, and the body posts no `cv` part when there is none.
- `POST /api/hub/freelancers` takes `cv` as optional, and `FreelancerService.apply` takes no bytes and stores the card without them. **An application that omits the CV never clears a CV already stored**: somebody refreshing their answers from the wizard is not somebody deleting their file.
- The member area's edit form is optional in both directions, with the hint as the difference: «Solo se vuoi sostituirlo» when we hold one, «Non ne abbiamo ancora uno» when we do not.
- The thanks page stops saying «Abbiamo il tuo profilo e il tuo CV», which it could no longer say for everybody, and the line below it points at the area where the file arrives.

Nothing new had to be modelled: a card without a CV was already a state this code knew, because an admin can draft one from a signup (ORB-155). `completa` is false, `cv_of` answers `NotFound`, the admin list says «da completare», and `replace_cv` exists to fill it in.

## How I verified it

- `pnpm --filter hub test`: 103 passed in 21 files. The wizard's own step test now pins that a missing CV validates and that the step is optional; a new test walks the whole wizard without a file and asserts the posted `FormData` has no `cv` part at all. The two member-area tests that demanded a CV now assert the opposite, including that «Salva» on a card with no CV sends the `PATCH` and no `PUT`.
- `uv run pytest -q projects/hub`: 215 passed. New in core: an application with no CV is stored, is not `completa` and has nothing to download; a second application without a CV keeps the bytes already stored; and a card the wizard left without one is completed from the member area, with the same «CV caricato dalla persona» in its thread. New in the API tests: a body with no `cv` part answers 201 and the row has no CV, while a `cv` that is not a PDF is still a 422 naming the field.
- `pnpm --filter hub lint`, `tsc --noEmit`, `uv run ruff check projects/hub`, `uv run ruff format --check`, `uv run mypy`: clean.
- In a real browser, both worktrees side by side on the dev server: at «Il tuo CV», pressing «Avanti» with nothing attached answers «Serve il CV, in PDF.» on `main` and moves to «Quanto costa una tua giornata?» (5 di 8) on this branch. That is the pair below.

- **The member area was not in the ask**, and I changed it anyway: leaving it as it was would have meant a person who skipped the CV at signup could not save a change to their rate until they produced one. I would rather that decision be seen than buried, and it is the reason the two `Modifica` tests are inverted rather than deleted.
- `apply()` keeps its positional signature (`cv`, `cv_filename`, `cv_mime`), now defaulted, so every existing caller and test is untouched. The MCP surface has no tool that applies on somebody's behalf, so it needed no change.
- The website's landing still says «Entri con il tuo CV e la tua tariffa». It is a different project and a copy decision, and it is now slightly ahead of what the form demands. Worth a look when the copy is next opened.

## API changes

`POST /api/hub/freelancers`: the `cv` part becomes optional. No response shape changed, and a non-PDF is still a 422 naming the field. The hub's web client is hand-written (`projects/hub/apps/web/src/lib/api.ts`) and already left the part out when no file was chosen, so it needed no change.

The rule that a CV must be a PDF under 5 MB, decided on the bytes rather than on the declared type. The admin's view of a card, which already distinguished «da completare». The retention question, which the README still lists as open.

## Screenshots

Two pairs, same data, same viewport, `main` on the left and this branch on the right: the CV step itself, and what «Avanti» does with nothing attached.

pair-1-cv-step

pair-2-cv-skip

Linear: ORB-201.


slavni96:
## Independent review, applied in 6c4cdbb

Seven findings. Two were real defects, and the first one was mine.

1. **An attached empty file was swallowed as «no CV»** (high). `content or None` in the router plus a falsy `if cv` in `apply` meant a zero-byte upload answered 201 and stored nothing, where the old code refused it with «serve il CV, in PDF». Reachable from the browser: the file field accepts any pick and the client posts any `File`. Fixed by telling the two cases apart where the difference is known: `cv is None` is "nobody attached a file", `cv == b""` is a broken upload and a refusal. The wizard now says it before the round trip («Questo file è vuoto: riprova con il PDF.»), and there is a test at the API and at `apply`. The reviewer probed the installed FastAPI and confirmed the other three shapes (no part, empty filename, `cv` as an empty string field) all arrive as `None`, so they stay «no CV».
2. **The welcome mailing told an incomplete card it was complete** (medium). `orbiters welcome` picks its voice by `compilata_da`, so a person who filled the wizard and skipped the CV got «La tua scheda è completa: sei X, e le aziende possono trovarti.» That is false, and it asks for nothing. The voice now reads `completa` through `FreelancerRead`, the same predicate the admin list and the member area read, and says «manca solo il CV [...] lo carichi dalla tua area». New test in `test_mail.py`.
3. The route comment described a mechanism that does not exist (an empty part «read as the same thing a few lines below»); it now describes what FastAPI actually does, and the two `cv.filename or "" if cv is not None else ""` expressions are gone with the shape that made them necessary.
4. Four comments said a card can only be incomplete because an admin drafted it from a signup. They name the wizard's skipped step too, in `Area.tsx`, `lists.tsx` and twice in `schemas.py`, and `_is_complete` stops calling itself «a card the wizard would have accepted», which is no longer true and is the point of keeping the fuller bar.
5. `apply` writes the four CV columns once rather than twice, and the `len(cv or b"")` that only existed to satisfy the type checker is gone.
6. The empty-part cases are pinned at both layers, as above.

**One suggestion I did not take as given.** The reviewer read the ternary in the member test as pointless duplication and proposed `mockResolvedValue(answer(200, WITHOUT_CV))`. That hands the same `Response` to every call, and a body can be read once: the GET consumed it and the PATCH failed for a reason that has nothing to do with the CV, which the suite caught immediately. It is `mockImplementation(async () => answer(...))`, a fresh response per call, with the reason written beside it.

**Scope.** The reviewer agrees the member-area change belongs in this PR rather than in its own: `editSteps(false)` was reachable only from an admin draft before, and the moment the wizard stops demanding a PDF it becomes the landing spot for everybody who skipped the step, so leaving it required would ship a branch where this PR's own promise is false.

After: `uv run pytest -q projects/hub` 218 passed (215 before), `pnpm --filter hub test` 103 passed, lint, `tsc --noEmit`, ruff and mypy clean.

Still open, and named in the PR body: the Linear card. The connector has been down for this whole session, so the id still has to go on this PR before it merges.

PR #113, opened 2026-09-15
feat(hub): the LinkedIn step takes the name, or the address as it was pasted

slavni96:
## What this changes

«Il tuo profilo LinkedIn» took only `https://…linkedin.com/…`, and stored what it took exactly as typed. What a person has to hand is `linkedin.com/in/ada-lovelace/?utm_source=share`, from a phone's address bar or LinkedIn's share sheet: the step answered «Serve l'indirizzo https di un profilo su linkedin.com, oppure niente.», and a share link that did carry `https` reached the admin list with its tracking on.

Now:

- The field shows **`linkedin.com/in/`** in front and asks for the name after it (placeholder `mario-rossi`). An address pasted whole is reduced to its name as it lands: `https://it.linkedin.com/in/ada-lovelace/?utm_source=share` becomes `ada-lovelace`. An address typed by hand, key by key, is left as typed and accepted too.
- Under the field, **«Apri il tuo profilo LinkedIn»** opens `https://www.linkedin.com/in/me/`, which takes a signed-in person to their own profile (checked: it lands on `/in/<name>/?isSelfProfile=true`), to copy from.
- **One stored shape**: `https://www.linkedin.com/in/<name>`, whatever arrived: no scheme, `http`, `it.` or `www.`, a trailing slash, a query, a fragment, a sub-page like `/details/experience/`. Another page on LinkedIn (a company, an old `/pub/` address) is kept as given, over `https`. Somewhere that is not LinkedIn is still refused.

The shape is decided in one place, `normalise_linkedin` in `orbiters_core.schemas`, which the signup, the wizard, the member area and the MCP all go through. The web has the same reading in `apps/web/src/lib/linkedin.ts`, so the review screen shows what will be stored and the body already carries it.

## How I verified it

- `uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests`: 244 passed. New or changed: eleven pasted shapes of one profile each stored as `https://www.linkedin.com/in/ada`; a company page and a `/pub/` address kept as given over https; the refusals now include `linkedin.com.evil.com/in/ada`, `evil.com/linkedin.com/in/ada`, `ftp://`, a bare `ada` and a newline after a scheme-less address; `MemberUpdate` stores a pasted `linkedin.com/in/ada/?trk=share` as the profile; `POST /api/hub/freelancers` with `linkedin.com/in/ada-lovelace/?utm_source=share` answers 201 and the row holds the profile, while `evil.com/linkedin.com/in/ada` is a 422 naming `linkedin_url`. (The first local run of the suite errored on the testcontainers Postgres starting up; the rerun was clean, nothing changed in between.)
- `pnpm --filter hub test`: 142 passed in 22 files. `linkedin.test.ts` pins every shape above on the web side, and that typing an address one key at a time is never cut down. The wizard test pastes the share link into the real step, sees `ada-lovelace` behind `linkedin.com/in/`, the link's `href`, the profile on the review screen and `linkedin_url` in the posted `FormData`. The member area shows a stored profile as `ada` and saves it back as the profile.
- `pnpm --filter hub lint`, `tsc --noEmit`, `uv run ruff check projects/hub`, `uv run ruff format --check projects/hub`, `uv run mypy`: clean.
- In a real browser, `main` and this branch on the dev server side by side (the pairs below). Taking them found a bug the tests had not: typing the address by hand reduced the field to «a» at `linkedin.com/in/a`, and the rest landed after it. Fixed in the second commit, with the test that would have caught it.

- **The independent review found a hole already on `main`**: a backslash in the host stored a link to another site. Fixed here, with everything else it raised; the PR comment lists each finding and what I did.

- **This reverses a written rule.** `schemas.py` said the address is «kept exactly as it was typed … because it is a person's own address for themselves». The new `DECISIONS.md` row says why the profile wins: one person, one stored string, and a sub-page or share tracking is lost on purpose.
- **The API refuses a bare name, the web completes it.** An agent passing `ada` through the MCP is more likely a mistake than a profile, so only the form, which shows the prefix, turns a name into an address.
- **Paste is told from typing by size**: a change that adds more than one character at once is reduced, a keystroke is not. An autofill counts as a paste, which is what we want; a single-character paste is not reduced, and it would not be a profile anyway.
- Rows already stored keep their old shapes; nothing rewrites them. They pass through the normaliser the next time the person saves from the member area.

## API changes

`POST /api/orbiters/signups`, `POST /api/hub/freelancers`, `PATCH /api/hub/me`: `linkedin_url` now accepts a LinkedIn address with no scheme or with `http`, and stores the normalised form. Still a 422 naming the field for anything not on LinkedIn. No response shape changed. The hub's hand-written client (`apps/web/src/lib/api.ts`) sends `linkedinProfile(...)` instead of the trimmed text.

The step's place in the wizard: how many screens the form has is ORB-118 and ORB-120. The website's landing signup, which keeps its own `https` check: it is another project, and now stricter than the API behind it. The MCP tool's signature.

## Screenshots

Two pairs, same input (`linkedin.com/in/ada-lovelace/?utm_source=share`, pasted in one go), 1440×900, `main` on the left and this branch on the right: the step right after the paste, and what «Avanti» does with it.

pair-1

pair-2

Linear: ORB-203.


slavni96:
## Independent review, and what I did with it

A fresh read-only reviewer went through the diff and probed both normalisers with about 60 inputs. Applied in `193b473d`.

**High, fixed.** `https://evil.com\.linkedin.com/company/x` passed the host check: `urlsplit` reads the host as `evil.com\.linkedin.com`, and a browser reads `\` as `/`, so the admin list's link opens evil.com. This was already true on `main`, through the public signup. A backslash is now refused on the raw value. A netloc with credentials (`user:pw@`) or a port is refused instead of stored. Another page is rebuilt from the host.

**Medium, fixed.** The `\u` escapes in `linkedin.ts` had become literal control characters, so git and GitHub treated the file as binary. Rewritten in ASCII.

**Low, fixed:**
- **Length.** The web now checks the stored length (300), so the step says it and the submit does not. The API's raw bound is 2000, because a share link carries its tracking; the stored value is still checked against the column.
- **The name.** It is decoded and must be letters, digits, `-` or `_`. So `j%C3%BCrgen` and `jürgen` are one stored address. `/in/`, `/in/..`, `/in/%2e%2e`, a space, `<` and a zero-width space are refused.
- **Paste detection.** It is now "more than one keystroke", not "longer by more than one", so a select-all and paste over a longer value is reduced too.
- **Screen readers.** The `linkedin.com/in/` prefix is the input's `aria-describedby`.
- **Ports.** A scheme may not contain a dot, so `linkedin.com:443/...` is refused by the port rule rather than read as a scheme.

**Beyond the review.** Web and API had to agree, so I ran both normalisers over the same 43 inputs. Three differences came out, all where the browser's `URL` forgives what `urlsplit` does not: `//www.linkedin.com/...`, `https:www.linkedin.com/...` and `https:/www...`. The web now requires the host written after `//` to be the host it parsed, and reads the path as written (`URL` resolves `/in/%2e%2e` to `/`). The rerun showed no differences. Those cases are in the tests on both sides.

**Left as is.** A stored value the web refuses blocks saving in the member area until the field is cleared (finding 6). With the backslash closed, the only such rows would be ones written before this PR in a shape it now refuses; the step names the field, so the person can see what to change.

Gates after the fixes: `pytest` hub 244 passed, `pnpm --filter hub test` 142 passed, lint, `tsc`, ruff, mypy clean.

PR #127, opened 2026-09-16
docs(repo): a PR ends with a video of the feature in use, beside the pairs

slavni96:
## What this changes

A PR here ended with before-and-after pairs, and that was all a reviewer got: a wizard, a menu that opens, a proforma that becomes an invoice, read from two still frames and the diff. Now the section that closes every body is «Screenshots and video»: the pairs stay for anything a person could see, and anything a person could **do** is shown working, in one `.mp4` of the feature in use, recorded from the branch's own worktree on the same data and viewport as the after frame.

- `docs/pr-screenshots/record.mjs`, beside `compose.py`, records it: a steps module that does what a person would do (`page.getByRole(...).click()`, `pause()`), Chromium from the `@playwright/test` the e2e tests already install, Playwright's own `recordVideo`, a cursor drawn on the page so each click is visible where it lands, and `ffmpeg` for the H.264 `.mp4` a PR body plays inline. It prints the length and the size, and refuses an output that is not `.mp4`.
- `docs/pr-screenshots/README.md` says what needs a video (a button that does something, a wizard, a dialog, a state that follows an action), what does not (one label, a colour: the pairs and a line saying why), how to record per app, and how to attach and verify it with the same `gh --attach` as the pairs.
- `.github/PULL_REQUEST_TEMPLATE.md` and the `pr-creation` skill ask for it; the section is renamed «Screenshots and video» in both.

Left alone, on purpose: no before video (the pairs already show what moved), no recording in CI, no GIF (an `.mp4` is smaller and sharper, and `gh` 2.99.0 uploads it).

## How I verified it

- `pnpm --filter website build && pnpm --filter website preview` from this worktree, then `node docs/pr-screenshots/record.mjs steps.mjs demo-1-landing.mp4 --url http://localhost:4173/ --viewport 1440x900`: `10.4 s, 1112 kB`. `ffprobe`: `h264`, `1440×900`, `yuv420p`, `25/1`. Three frames pulled with `ffmpeg` and read: the consent banner dismissed by the first step, the «Come funziona» band, the cursor on «Scopri PigroCRM», the PigroCRM page at the end.
- The usage paths: no arguments, no `--url`, an output not ending in `.mp4` each exit 2 with the reason on stderr; `node --check` clean.
- The upload is this PR's own closing section: the player below is the `gh pr create --attach ./demo-1-landing.mp4` of that file, and the verify block of the README run on this body (the count of `user-attachments`, no local path left).
- No test touches the template, the skill or `docs/`; CI runs nothing for this diff.

- The section header changes from «Screenshots» to «Screenshots and video». Older PRs keep the old header; nothing greps for it (checked: no workflow, test or script names it).
- Playwright is resolved with `createRequire` from `projects/website`, then `projects/pigrocrm/apps/web`, then `shared/brand`, the three packages that declare `@playwright/test`. If those move, the list at the top of `record.mjs` moves with them; the error names all three and says to run `pnpm install`.
- The cursor is a DOM node injected with `addInitScript`, `aria-hidden` and without a role, so a `getByRole` in the steps never matches it. It is drawn on the page under test, so it is in the video and in nothing else.

## Screenshots and video

No pair: nothing in a product changed, the diff is a script, a README, a template and a skill. The video below is the script's own output, the landing walked from `pnpm --filter website preview` on 4173 in this worktree at 1440×900, as the README describes.

**Video: the landing, from the consent banner to the PigroCRM page**

https://github.com/user-attachments/assets/5108ac0e-2da4-413f-8d42-97e3de9a360f

Linear: REB-220.


slavni96:
**Independent review, applied in ca553ac6.** Nothing at must-fix. Findings and what happened to each:

1. *Cursor vanishes after an in-page navigation* (should fix): fixed. The last position goes through `sessionStorage`; the init script places the dot from it on the next document, and a `mousemove` re-attaches the node if a page rewrote its root (nit 6 folded in). Verified on the landing → `/pigrocrm` flow: the dot is on the second page before the mouse moves.
2. *Temp dir leaks and the partial webm is never named on failure* (should fix): fixed. The run ends with `record.mjs: the flow failed: <message>. The partial recording is at <webm>`, exit 1; the dir is removed when nothing was recorded. `--viewport 0x900` is refused by the regex. Measured: a step on a missing button leaves one dir with the partial webm; a happy run leaves none.
3. *README overstates what the Playwright MCP cannot do* (should fix): reworded. Video is the server's `--save-video` launch option, the session's instance is not started with it, and its output would land in the server's directory.
4. *`locale: 'it-IT'` undocumented* (nit): dropped. The recording uses the browser's default locale, as the MCP's screenshots do; the README says so.
5. *`renameSync` across volumes* (nit): `copyFileSync`, the temp dir is removed afterwards as before.
6. *Cloned dot with no listeners* (nit): covered by 1.
7. *95-column line in the skill* (nit): rewrapped.
8. *Hand list of packages declaring `@playwright/test`* (nit): left as is, with the error naming all three.

Found on this PR while verifying, added in the same commit: `gh --attach` rewrites a video reference into a plain link, and GitHub plays the asset only as a bare URL in a paragraph of its own. The README now carries the second edit (read the body, `sed` the link away, `gh pr edit --body-file`), the skill points at it, and the body of this PR is the result: the player under «Screenshots and video» came from exactly that.

Also made explicit in README and skill, per Ivan's remark: pairs are before **and** after; the video is of the **after** only.

PR #128, opened 2026-09-16
docs(repo): the PR video recorder runs under Seatbelt on macOS

slavni96:
## What this changes

`docs/pr-screenshots/record.mjs` drove Chromium, a steps module written for the occasion and ffmpeg as the user, with every file and every network destination the user has: a steps module could write anywhere, a page under test could call home while recorded. Ivan asked for the runs that produce the screenshots and the video to be confined; bubblewrap is Linux-only and this is a Mac, so it is Seatbelt.

Now, on macOS, the script writes a Seatbelt profile for the run and re-executes itself under `sandbox-exec -f`; the inner process records as before. The profile: everything allowed, then writes denied except the output file's directory, the user's temp and cache directories (`DARWIN_USER_TEMP_DIR`, `DARWIN_USER_CACHE_DIR`, where Playwright's browser profile and the raw `.webm` go) and `/dev`; then the network denied except outbound to `localhost`, inbound and bind on `localhost`, and unix sockets (Playwright's pipes to Chromium). Reads are not confined. The first line of a run says what the profile allows. `--no-sandbox` opts out and says so; a platform without `/usr/bin/sandbox-exec` says so and records unconfined. The README's «Recording the video» section says what is confined, what is not, and when to reach for `--no-sandbox`.

Left alone, on purpose: no bubblewrap profile for Linux (nobody records on Linux here, and Chromium inside `bwrap` needs its own sandbox off, which deserves its own measurement); no `(deny default)` profile (Chromium wants dozens of mach services and IOKit reads, and a list of them breaks on the next macOS).

## How I verified it

On the website served from this worktree (`pnpm --filter website preview`, 4173), with a probe steps module that writes to `$HOME` and `fetch`es `https://example.com/` from the page:

- Default run: first line `record.mjs: under Seatbelt: writes only to <out dir>, the temp and cache directories and /dev; network only to localhost`; probe `home write EPERM; external fetch blocked:Failed to fetch`; video `5.5 s, 667 kB`; no file in `$HOME`.
- Same profile from node: `fetch('https://example.com/')` fails `EPERM`, `fetch('http://localhost:4173/')` answers 200.
- `--no-sandbox`: `probe: home write allowed; external fetch reached`, so the flag does what it says, and the sandbox was what blocked the rest.
- An output path in a directory that does not exist yet: created and allowed, the video lands there.
- A failing step under the sandbox: the error and the partial `.webm` path come through, exit 1; a bad viewport exits 2 before any re-exec; no `pr-video-*` temp dir left after a happy run.
- The first profile I tried, `(allow network* (local ip "localhost:*"))`, let the page reach example.com: an outbound socket matched the `local` filter. The shipped profile allows outbound on `remote ip` only, and that is the measurement in the commit.
- No test touches `docs/`; the `ci` job is the only check for this diff.

- `sandbox-exec` is deprecated by Apple and still present on macOS 26.6 (measured). If a future macOS drops it, the script says «no Seatbelt on this platform» and records unconfined rather than failing: the video still gets made, and the line says the run was open. A reader who wants the opposite (refuse to record unconfined) has a one-line change in `runInSandbox`.
- Paths in the profile go through `realpathSync`: Seatbelt compares canonical paths and `/var` is a symlink to `/private/var`. The output directory is created before that call so it can be resolved.
- Reads are deliberately not confined: a steps module still reads the repository, `~/.ssh`, everything. Confining reads means enumerating what Chromium needs, which is the deny-default profile this PR does not write.

## Screenshots and video

No pair: nothing in a product changed. The video below is the landing walkthrough recorded by this branch's script **under the sandbox**, on the website served from this worktree at 1440×900: the same recording as PR #127's, now confined.

**Video: the landing, recorded under Seatbelt**

https://github.com/user-attachments/assets/ec69004c-bca6-4d6f-bf37-db7283fc9385

Linear: REB-225.


slavni96:
**Independent review, applied in the second commit.** 7 findings, all applied:

1. *`$TMPDIR` not in the writable list* (must fix): fixed. `os.tmpdir()` is allowed beside `DARWIN_USER_TEMP_DIR` and `DARWIN_USER_CACHE_DIR`, deduped after `realpath`. Measured with `TMPDIR` set to a directory whose name has a space and a double quote: the probe still reads `home write EPERM; external fetch blocked`, the video comes out, nothing is left in that directory.
2. *Ctrl-C leaks the profile directory* (should fix): fixed. The outer process ignores `SIGINT`/`SIGTERM` while `spawnSync` runs, removes the profile in `finally`, and re-raises the signal on itself when the inner died by one. Measured with `kill -INT` on the process group four seconds in: no profile directory left, no Chromium left.
3. *`getconf` fallback can throw in the outer* (should fix): fixed. Directories that do not exist are filtered out before `realpathSync`; the cache fallback is gone.
4. *`execArgv` dropped on re-exec* (nit): fixed.
5. *Output directory created only on the sandboxed path* (nit): fixed, created at the top level for every path.
6. *A stray `PR_VIDEO_SANDBOXED` in the shell skips confinement* (nit): the marker has to be exactly `1`.
7. *`network-bind` on `localhost:*` buys nothing* (nit): removed; the probe and the landing recording pass without it.

The video in the body is re-recorded under the final profile (second attach, then the README's second edit).

PR #130, opened 2026-09-16
fix(pigrocrm): the CRM host tells crawlers to stay out, in the vhost and in the SPA head

fiorelorenzo:
## What this changes

The CRM's production vhost (`projects/pigrocrm/deploy/nginx/pigro.letsrebase.conf`) sends no `X-Robots-Tag`, serves no `robots.txt`, and the SPA head (`projects/pigrocrm/apps/web/index.html`) carries no robots meta -- so the root redirect, which answers with a real company's name in its `Location`, is one hop from a search result. This mirrors the two lines the preview vhost already carries onto the production vhost (a server-level `X-Robots-Tag` and a `robots.txt` location; no location block declares its own `add_header`, so none needed the header repeated) and adds the same `noindex` meta the hub's SPA head already has. A new vitest asserts all three stay in place.

## How I verified it

- `pnpm exec vitest run src/test/robots.test.ts` from `apps/web`: 3 passed. Confirmed it is a real regression test by stashing the two source changes and re-running it first (3 failed with the exact missing-header/missing-meta diffs), then restoring them (3 passed again).
- `pnpm --filter web test`: 112 files, 1177 tests passed.
- `pnpm --filter web build && pnpm --filter web exec tsc --noEmit`: clean, no type errors.
- `pnpm --filter web lint` (eslint on the new test file): clean.
- `docker compose build web` (pigrocrm-images, the web image only): built in 44s. Note this image only bakes in `deploy/nginx/spa.conf` (the in-container SPA server), not the host vhost this PR changes, so it does not exercise the new lines; it only proves the SPA build and image still assemble.
- Skipped: `pigrocrm-py-*` (no Python file touched), `pigrocrm-e2e` (the full Playwright suite; nothing in it reads the response headers or `robots.txt`, and it is a serial, ports-and-Postgres-heavy suite this change does not touch), the other projects' checks (not on this diff's `preflight --list` output).
- Could not verify: `nginx -t` against the edited vhost -- no `nginx` binary on this box. The syntax mirrors the preview vhost's already-installed, already-working block verbatim (same `add_header` line, same `location = /robots.txt` body), so I am not asking a reviewer to trust an untested shape, only an untested *file*.

`projects/pigrocrm` is Ivan's project; this touches two lines in its vhost and one in its SPA head. The vhost in this repo is not what's running: per `docs/adding-a-project.md` §7, the file on the server is edited in place (`certbot --nginx` has already rewritten it once, its 443 block sits on top of this one), so this PR only prepares the diff the orchestrator applies to the live file after merge -- landing this PR does not by itself change what `pigro.letsrebase.com` sends until that edit happens.

## Screenshots and video

Nothing a person sees: a response header, a `robots.txt` body, and a `<meta>` tag with no visual rendering. No pair or video to capture.

Linear: REB-107.


fiorelorenzo:
Independent review (severity-ranked) and what I did with each finding.

1. **CI/preflight never ran the guard when its own subject changed** (medium). `ci.yml`'s `pigrocrm_web` filter and `preflight.json`'s `pigrocrm-web-unit` `when` list only covered `apps/web/**`, not `deploy/nginx/pigro.letsrebase.conf`, so a commit that broke the header or the `robots.txt` location without touching `apps/web` would skip the new vitest on both tiers. Applied: added the vhost path to both.

2. **The X-Robots-Tag assertion matched anywhere in the file, not the server level** (low-medium). nginx does not inherit `add_header` into a location block that declares its own, so moving the header into `location = /robots.txt` would leave the root redirect (the actual point of REB-107) with no header, while the old regex still passed. Applied: sliced the assertion to the text before the first `location` block, plus a count check that only one `add_header` directive exists (matched at line start so a comment mentioning the name is not itself counted). Verified live: mutated the vhost to move the header into the location block, the test failed (2/3); restored, it passed (3/3) again.

3. **No note that the two new stanzas go into the certbot-rewritten copy, not over it** (low). The file's own head comment already said certbot rewrites the file in place; extended it to say explicitly that the header and the `robots.txt` location belong inside the installed copy's existing 443 block, matching the wording `projects/website/deploy/letsrebase.conf` already uses for the same situation.

All three applied in commit 0ede2149. `pnpm --filter web test` (1177 tests), `lint`, `build`, and `tsc --noEmit` all still green; `preflight --list` now shows the vhost path under `pigrocrm-web-unit`'s matched paths.

PR #131, opened 2026-09-16
feat(website): robots.txt answers 200, naming the sitemap and disallowing /pitch

fiorelorenzo:
## What this changes

`https://letsrebase.com/robots.txt` answered 404: the website's own nginx maps four
exact paths and `/assets/`, then 404s everything else, this file included. It now
answers 200 as `text/plain`, generated at build time from `src/path-map-plugin.ts`'s
`PAGES` map and a new `NOINDEX` list (today just `/pitch`, which already carries
`<meta name="robots" content="noindex">`):

```
User-agent: *
Disallow: /pitch
Sitemap: https://letsrebase.com/sitemap.xml
```

`Disallow: /pitch` is deliberate, not an oversight: the page already opts out of
indexing with its own meta tag, and saying the same thing in robots.txt costs nothing
and covers a crawler that ignores meta robots but reads this file. No other page is
disallowed.

`deploy/nginx.conf` gains `location = /robots.txt { try_files /robots.txt =404; }`,
the same shape as every other page, so `path-map-plugin.test.ts` (which parses
`nginx.conf` and compares it with the plugin's own map) keeps the two in sync. The dev
and preview servers answer the same generated content through the existing
`pathMapPlugin` middleware; the build gets the file from a new `writeBundle` hook on
that same plugin, since there is no HTML input in `vite.config.ts` to build it from.

The sitemap this file names does not exist yet: REB-110, next in this wave, builds it
from the same `PAGES` map right after this merges.

## How I verified it

- `pnpm --filter website test` — 263 passed (14 in `path-map-plugin.test.ts`,
  including three new ones: the nginx location exists, `route('/robots.txt')`
  generates the exact expected content, and the plugin's `writeBundle` hook actually
  writes the file to a temp directory).
- `pnpm --filter website exec tsc --noEmit` — clean.
- `pnpm --filter website build` then `cat dist/robots.txt` — the file above, byte for
  byte.
- `pnpm --filter website dev` (scratch port, since :5173 collides with nothing here)
  and `pnpm --filter website preview` (:4173) — `curl -D - http://localhost:<port>/robots.txt`
  both answered `200`, `Content-Type: text/plain; charset=utf-8`, the body above.
- `pnpm --filter website test:e2e` — 56 passed against `vite preview` on :4173.
- `WEBSITE_WEB_PORT=127.0.0.1:8099 docker compose build` (the `website-image`
  preflight check) — built, then ran the image directly and curled it:
  `docker run -d --rm -p 127.0.0.1:8199:80 website-web:latest`,
  `curl http://127.0.0.1:8199/robots.txt` answered `200` with the same body from
  nginx itself, and `/nonexistent` still answers `404`.
- Skipped: the full Python suite and the CRM/hub gates, since this diff touches only
  `projects/website`; `preflight --list` selected `website-lint`, `website-unit`,
  `website-build`, `website-types`, `website-e2e` and `website-image`, all run above.

`NOINDEX` is a small explicit list beside `PAGES` (today one entry, `/pitch`) rather
than something that parses `pitch.html`'s own `<meta name="robots">` at build time.
The card allowed either; I picked the explicit list because a Vite plugin parsing its
own sibling HTML file's meta tags is more moving parts for the same one-entry answer,
and `NOINDEX` is exactly what REB-110's sitemap exclusion needs next, so keeping it
here rather than duplicating `/pitch` by hand in that plugin too is the point of
having it now instead of waiting for REB-110 to invent it.

## Screenshots and video

Nothing visible changed: this is a file nobody browses to, and its only observers are
crawlers and this PR's own verification commands.

Linear: REB-109.


fiorelorenzo:
Independent review (agent, read-only) on the diff in commit 6ffefff4. Findings and what I did with each:

1. **`writeBundle` comment stated the opposite of rolldown's contract** (it runs only at the end of a real `bundle.write()`, not "whether or not write is false"). Fixed the comment; verified against `rolldown@1.2.3`'s own type doc comments.
2. **Nothing in the suite proved a real `vite build` writes `dist/robots.txt`**: the preview server's own middleware answers the same content from memory regardless of what is on disk, so an HTTP request against it can't tell the two apart. Added an e2e test that reads `dist/robots.txt` directly (`e2e/site.spec.ts`), alongside the existing HTTP check.
3. **`Disallow: /pitch` worked against its own goal**: blocking the page from crawling stops a crawler from ever reaching its `noindex` meta tag, and per Google's own docs a blocked-but-linked page can still be indexed with no snippet, worse than the noindex outcome it already has. Removed the `Disallow` line entirely; robots.txt now disallows nothing a visitor can reach, one of the two shapes the card allowed. `NOINDEX` is gone from this PR since nothing here reads it anymore.
4. Three lower-severity findings applied: `writeBundle` now throws instead of silently skipping the write when there is no output directory; `SITE_HOST` is now also the source for `share-card.test.ts`'s `CARD_URL`, so a rebrand that misses a head fails that test; the `writeBundle` unit test cleans up its temp directory.
5. One doc finding applied: one sentence in `README.md` naming `robots.txt`/`GENERATED_PATHS` as a third kind of build output.

Findings I left as they were: none, all applicable findings were addressed (the two priority-2 items and four of five priority-3 items; the fifth, giving `NOINDEX` a tuple shape like `GENERATED_PATHS`, is moot since `NOINDEX` no longer exists in this PR).

PR #132, opened 2026-09-16
fix(hub): the wizard review list validates as a dl under axe

fiorelorenzo:
## What this changes

`/hub/aziende`'s review screen («Tutto giusto?») reported a `serious` axe `definition-list`
violation at both 390x844 and 1440x900: "dl element has direct children that are not
allowed: div > button". Each review row is a `<div>` wrapping `<dt>`, `<dd>` and a
"Modifica" `<button>` as siblings; axe flattens a wrapping div that carries no explicit
role and only allows `dt`/`dd` among its children, so the button failed regardless of
ORB-89 (the missing `<h1>`) or ORB-93 (the 390px wrap), both unrelated to this markup.

"Modifica" now lives inside the `<dd>`, as a second flex child next to the summary text,
instead of as a third sibling of `dt`/`dd`. axe never looks inside a `dd` (it only
flattens `div` wrappers), so the row validates. The row's spacing and wrap are unchanged:
the button keeps the exact classes it had, the outer row keeps `dt`/`dd` as its only two
children, and the maths that produced the old gaps (4px mobile via `gap-1` + `mt-1`, 16px
desktop via `gap-4`) is reproduced one level deeper. Behaviour is untouched: "Modifica"
still jumps back to its step.

## How I verified it

- `pnpm --filter hub build` — vite build + `tsc --noEmit`, clean.
- `pnpm --filter hub lint` — clean.
- `pnpm --filter hub test` — 162 tests, 25 files, all passed.
- Reproduced first: drove a second worktree on `origin/main` (`071a78a5`) through the
  five-step company wizard to the review screen with a small browser script (draft
  seeded via the wizard's own `localStorage` key once filled), injected axe-core 4.13.0
  and ran `axe.run(document, { resultTypes: ['violations'] })`. It reported the
  `definition-list` violation, `impact: "serious"`, at both 1440x900 and 390x844,
  confirming the reproduction the card describes.
- Same flow against this branch's build (`vite preview`, port 4174 after a sibling
  agent needed 4173 for the website's `strictPort` preview): `axe.run` returns
  `violations: []` at both 1440x900 and 390x844. No other rule regressed (the
  `resultTypes: ['violations']` run was empty, not just filtered to this rule).
- `preflight --list` on this diff selects `hub-web-lint`, `hub-web-unit`, `hub-web-build`;
  all three run above. Nothing else in the tree is touched, so the rest stays skipped
  (no Python suite, no Playwright, no other project's build).

The gap between the summary text and "Modifica" used to come from the outer row's own
`gap-1`/`gap-4` plus the button's `mt-1`. Nesting the button inside `dd` reproduces the
same two numbers one level down (`dd`'s own `flex flex-col gap-1` / `sm:flex-row
sm:gap-4`, button keeps `mt-1 ... sm:mt-0`) rather than changing the layout model, so a
long summary that wraps to two lines lays out exactly as before (`dd` still owns
`sm:min-w-0 sm:flex-1`, now on both the wrapper and the inner `span`).

## Screenshots and video

**1. Review row, before and after (1440x900)**

Review row, before and after

Pixel-identical: text position, wrap point and the "Modifica" button's position (boxed)
are unchanged; only its DOM position moved from a sibling of `dd` to a child of it.

No video: nothing a person can do here changed. The click on "Modifica" still does the
same thing it did before; the fix is markup only, verified by axe and by the frame pair.

Linear: REB-96.


fiorelorenzo:
Independent review (reviewer agent), verdict: correct, no blocker or major issue.

Findings and what I did:
- **minor**, no regression guard for the axe fix: nothing pinned the button-inside-dd shape beyond a comment, so a later cleanup could reintroduce the violation silently. Applied: added an assertion in `Wizard.test.tsx` that every review row's element children are `dt`/`dd` only, which fails on the pre-fix markup and passes on this one (`7c54f0af`).
- **nit**, the comment named the wrong/imprecise axe machinery: it read as if `definition-list` itself rejects a button, when it is specifically `only-dlitems` flattening the row's role-less wrapping div and rejecting any non-dt/dd child (a button, a span, a stray text node). Applied: reworded the comment to name the check and its actual condition (`7c54f0af`).
- **nit**, pre-existing: every "Modifica" button shares the accessible name "Modifica" with nothing distinguishing rows for a screen reader (`Wizard.test.tsx` already indexes them positionally). Left as is: unrelated to this card's dt/dd fix and out of the stated non-goal of restructuring the review screen beyond it; reporting it to the orchestrator to file as its own card.

PR #133, opened 2026-09-16
test(repo): a script proves the non-public hosts stay unindexable

fiorelorenzo:
## What this changes

Nothing ran a request against `preview.letsrebase.com`, `preview.pigro.letsrebase.com`
or `pigro.letsrebase.com` before this: their unindexability rested entirely on
hand-edited vhosts that certbot also rewrites in place, with nothing checking that the
header, the basic auth (before REB-103) or the robots.txt location survived an edit.

`projects/website/deploy/check-unindexable.sh` asserts, for each of the three hosts,
following no redirect: the answer carries `X-Robots-Tag` with an unscoped `noindex`, and
`/robots.txt` answers `200` with `User-agent: *` paired with `Disallow: /`. Both preview
names are open since REB-103 (no basic auth), so the script authenticates nothing and
holds no credential; both curl calls put `-q` first so a stray `~/.curlrc` cannot
reintroduce one, or silently turn on `-L` (two of the three hosts answer `/` with a 302,
and a followed redirect would let `-D -` judge the wrong hop). `CHECK_UNINDEXABLE_HOSTS`
and `CHECK_UNINDEXABLE_SCHEME` point the same checks at a local responder for the
negative-path test, and are refused outright when `PREFLIGHT=1`, so a variable left set
in a shell cannot silently turn a real run into a check of localhost that still reports
green; an empty host list fails loudly instead of exiting 0 having checked nothing.

`pigro.letsrebase.com` is added per the card's **Adjacent** note: it is the CRM's
production host and only carried the header once REB-107's vhost (PR #130, same wave)
was installed there. It landed on the host while this PR was in review; all three hosts
now pass live.

Wired into `.github/preflight.json` as `check-unindexable`, naming the three vhosts it
actually checks one by one rather than globbing their directories (the globs the card
suggested also matched `nginx.conf`, `spa.conf` and `orbiters-proxy.conf`, which the
script reads nothing from, and `letsrebase.conf`, production, which stays indexable and
is never asserted on) and selecting its own path too, so editing the checker runs the
checker. It reaches live hosts, so a GitHub-hosted runner has nothing of its own to
prove and this never runs as a `pull_request` or `push` job. `docs/adding-a-project.md`
§7 gets one sentence next to "A preview name is open" naming the script and the card.

## How I verified it

- **Run today, against the live hosts** (`sh projects/website/deploy/check-unindexable.sh`):

  ```
  check-unindexable: https -> preview.letsrebase.com preview.pigro.letsrebase.com pigro.letsrebase.com
  PASS preview.letsrebase.com: X-Robots-Tag carries noindex on /
  PASS preview.letsrebase.com: robots.txt disallows everything for every crawler
  PASS preview.pigro.letsrebase.com: X-Robots-Tag carries noindex on /
  PASS preview.pigro.letsrebase.com: robots.txt disallows everything for every crawler
  PASS pigro.letsrebase.com: X-Robots-Tag carries noindex on /
  PASS pigro.letsrebase.com: robots.txt disallows everything for every crawler
  ```

  `exit=0`. When the PR first opened, `pigro.letsrebase.com` failed (REB-107's vhost was
  not installed yet); that run and the reason were reported honestly in this section
  rather than hidden. REB-107 has since landed on the host, so all three pass today.
- `PREFLIGHT=1 CHECK_UNINDEXABLE_HOSTS=localhost:8000 sh ...`: refuses to run
  (`exit=2`, "must not be set under preflight"), proving the override-safety guard added
  in review.
- `CHECK_UNINDEXABLE_HOSTS=" " sh ...`: fails loudly (`FAIL: no hosts to check`,
  `exit=1`) instead of passing on zero checks.
- **Negative path**, a responder that omits the header, via the same overrides:

  ```
  $ python3 -m http.server 8093 --bind 127.0.0.1 &
  $ CHECK_UNINDEXABLE_SCHEME=http CHECK_UNINDEXABLE_HOSTS=127.0.0.1:8093 \
      sh projects/website/deploy/check-unindexable.sh
  FAIL 127.0.0.1:8093: no unscoped X-Robots-Tag: noindex on http://127.0.0.1:8093/
  FAIL 127.0.0.1:8093: robots.txt at http://127.0.0.1:8093/robots.txt answered 404 or did not disallow / for *
  ```

  `exit=1`. As a control, a second tiny responder that does send
  `X-Robots-Tag: noindex, nofollow, noarchive` and a `/robots.txt` with
  `User-agent: *` / `Disallow: /` passes both assertions and exits `0`.
- `shellcheck -s sh projects/website/deploy/check-unindexable.sh`: no findings.
- `python3 -c "import json; json.load(open('.github/preflight.json'))"`: valid.
- `preflight check-unindexable`: `PASS check-unindexable 0.8s` (run through preflight
  itself, not just the bare script).
- `preflight --list` on the final diff: `check-unindexable` runs, selected by its own
  path; `website-e2e` and `website-image` now correctly skip (excluded via `unless`),
  where before the fix they ran a full Playwright suite and a docker build over a shell
  script neither reads. `website-lint`, `website-unit`, `website-build`, `website-types`
  still run because the script's path matches their existing `projects/website/**`
  globs; all four are cheap and green: `pnpm --filter website lint` (clean),
  `pnpm --filter website test` (260 tests, 13 files), `pnpm --filter website exec tsc
  --noEmit` (clean), `pnpm --filter website build` (452ms).

`check_host`'s two failure branches (`could not reach .../`, `could not reach
.../robots.txt`) each `return` before falling into the pass/fail check below them, so a
host that is entirely unreachable reports one `FAIL` line, not two contradictory ones.

An independent review flagged that shipping this check red-by-design (before REB-107
landed) could block `git push` for anyone with `preflight --install-hook` installed, on
a diff touching one of the three vhost files. It resolved itself when REB-107 deployed
mid-review; noted here since the general shape (a new check that starts red on a shared
box) is worth remembering next time one is added ahead of the fix it depends on.

## Screenshots and video

Nothing visible: this is a shell script, a preflight entry and one sentence of
documentation. No page, no pixel, no interaction changed.

Linear: REB-108.


fiorelorenzo:
Independent review (reviewer agent). Two major findings, both applied in fa989f48; two minor findings applied; two minor and two nit findings declined, reasons below.

Applied:
- **major**, silent false-green: `CHECK_UNINDEXABLE_HOSTS`/`CHECK_UNINDEXABLE_SCHEME` are inherited from any environment, including preflight's own, so a variable left set anywhere would have quietly redirected a real run at localhost while still reporting PASS; a whitespace-only host list also ran zero checks and exited 0. Applied: the script now refuses to run under `PREFLIGHT=1` with either override set, echoes which scheme/hosts it used, and fails on an empty host list.
- **minor**, curl without `-q` first lets `~/.curlrc` reintroduce a credential or `-L`; with `-L`, `/` on two of the three hosts is a 302 and `-D -` would judge the wrong hop. Applied: `-q` is now the first argument on both calls.
- **minor**, both assertions accepted a directive scoped to one crawler (`X-Robots-Tag: googlebot: noindex`, a `Disallow: /` under a narrower `User-agent` group while a later group allows everything). Applied: the header match now rejects a scoped value, and robots.txt requires `User-agent: *` paired with `Disallow: /`.
- **minor**, over-broad `when` globs pulled in nginx.conf/spa.conf/orbiters-proxy.conf (the script reads none of them) and letsrebase.conf (production, which stays indexable and is never asserted on), and the script did not select itself while its own path dragged in Playwright and a docker build via the broad `projects/website/**` globs on website-e2e/website-image. Applied: `preflight.json` now names the three vhosts the script actually checks one by one, adds the script's own path to its `when`, and excludes that path from website-e2e and website-image.

Declined:
- **major**, "the check is red on arrival and blocks git push via the pre-push hook": correct as written when I opened the PR, and I did report `pigro.letsrebase.com` failing honestly rather than hide it. Overtaken by events: REB-107 landed on that host while this review was running, and the check now passes all three hosts live (confirmed again after the fixes above). The general point (a check that ships red-by-design can block an unrelated push through the installed hook) is real and worth keeping in mind for the next one, but there is nothing left to change here today.
- **minor**, add a negative assertion that `letsrebase.com` (production) stays indexable: a different check than this card asked for (REB-108 is about the non-public hosts), reporting it to the orchestrator as a possible follow-up rather than growing this PR's scope.
- **nit**, `set -euf` to disable globbing on the host-list loop: applied as part of the override-safety fix above (line 33 already reads `set -euf`).
- **nit**, the doc's example commands said `bash` where the script's shebang and preflight both use `sh`: fixed in the same commit.

PR #134, opened 2026-09-16
feat(website): sitemap.xml lists the indexable pages, generated from PAGES

fiorelorenzo:
## What this changes

`https://letsrebase.com/sitemap.xml` answered 404 and no page referenced a sitemap,
so discovery was link-following only. It now answers 200, generated at build time from
`src/path-map-plugin.ts`'s `PAGES` map (the same one `deploy/nginx.conf` and the dev
and preview servers already share):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url><loc>https://letsrebase.com/</loc></url>
  <url><loc>https://letsrebase.com/pigrocrm</loc></url>
  <url><loc>https://letsrebase.com/community</loc></url>
  <url><loc>https://letsrebase.com/privacy</loc></url>
  <url><loc>https://letsrebase.com/termini</loc></url>
</urlset>
```

`/pitch` is excluded: it already carries `<meta name="robots" content="noindex">`
(`src/pitch.html:8`), and a sitemap listing a page you have told crawlers not to index
is a contradiction. The exclusion is a new `NOINDEX` list beside `PAGES` (today one
entry) rather than a second hand-written list, and this PR adds a test that holds
`NOINDEX` against every page's own head, so the two cannot drift silently the way
REB-109's review flagged as a risk for the list this PR reintroduces.

No `lastmod`, `changefreq` or `priority`: none of them would be true, and a build
timestamp on every page every deploy tells a crawler nothing.

Same mechanism as REB-109's robots.txt, which already names this file:
`deploy/nginx.conf` gains its own `location = /sitemap.xml`, and the plugin's
`writeBundle` hook (added in REB-109) writes both generated files into `dist/`
together, since neither has an HTML input in `vite.config.ts`.

## How I verified it

- `pnpm --filter website lint` and `pnpm --filter website exec tsc --noEmit` — clean.
- `pnpm --filter website test` — 269 passed, including: the nginx location exists,
  `route('/sitemap.xml')` lists exactly `Object.keys(PAGES)` minus `NOINDEX` (a page
  added to `PAGES` or `/pitch` appearing in the output both fail this, by set
  equality, not just membership), no `lastmod`/`changefreq`/`priority` anywhere in the
  output, the `writeBundle` hook writes `sitemap.xml` to a temp directory, and
  `NOINDEX` matches exactly the pages whose own head carries the noindex meta tag.
- `pnpm --filter website build` then `cat dist/sitemap.xml` — the five URLs above,
  `/pitch` absent.
- `pnpm --filter website preview` (:4173): `curl http://localhost:4173/sitemap.xml`
  answered 200, `Content-Type: application/xml; charset=utf-8`, the same body.
- `pnpm --filter website test:e2e` — 58 passed (56 before this wave plus the two REB-109
  added), including a new test that reads the real `dist/sitemap.xml` the build wrote
  and checks all five URLs are present and `/pitch` is not, for the same reason
  REB-109's robots.txt test reads the disk file rather than trusting an HTTP response:
  the preview server's middleware answers from memory regardless of what landed on
  disk.
- `WEBSITE_WEB_PORT=127.0.0.1:8099 docker compose build` (`website-image` preflight),
  then ran the image and curled it directly: `/sitemap.xml` answered 200 with the same
  five URLs from nginx itself. One thing worth naming rather than fixing: nginx serves
  it as `Content-Type: text/xml` (its own default MIME mapping for `.xml`), while the
  dev and preview servers answer `application/xml; charset=utf-8`; both are accepted
  by every sitemap consumer I'm aware of (Google, Bing), and it is the same kind of
  environment difference robots.txt already has (`text/plain` in production versus
  `text/plain; charset=utf-8` in dev/preview) that nobody has asked to unify.
- Skipped: the full Python suite and the CRM/hub gates, since this diff touches only
  `projects/website`; `preflight --list` selected the same six website checks as
  REB-109, all run above.

Nothing new beyond what REB-109's review already covered for the shared mechanism
(`writeBundle`, `GENERATED_PATHS`). The one thing specific to this card: `NOINDEX` is
a plain array (`['/pitch'] as const`), matching `GENERATED_PATHS`'s shape rather than
the `Record` shape `PAGES`/`REDIRECTS`/`ELSEWHERE` use, since its value is never read,
only its membership.

## Screenshots and video

Nothing visible changed: this is a file crawlers read, not visitors.

Linear: REB-110.


fiorelorenzo:
Independent review (agent, read-only) on the diff in commit a519108d. Overall verdict: correct, no blocking findings. What I did with each:

1. **Sitemap Content-Type diverged from what nginx actually serves**: the middleware answered `application/xml`, nginx's `mime.types` serves the built file as `text/xml`. Changed both the middleware and the tests that pin it to `text/xml; charset=utf-8`, verified against the docker image (`Content-Type: text/xml`).
2. **`writeBundle`'s failure message still named only robots.txt** after this PR added `/sitemap.xml` to `GENERATED_PATHS`. Reworded to name the list, not one file.
3. **The nginx comment and the README paragraph both still read as if robots.txt were the only generated file.** Updated both to name `sitemap.xml` too.
4. **A test comment overclaimed what the sorted-array assertion catches** (it does not, and is not meant to, catch a page quietly removed from `PAGES`; the nginx-parity test does that). Corrected the comment.
5. **The NOINDEX-drift test matched one exact spelling** of the noindex meta tag. Loosened it to a regex tolerant of attribute order and line wrapping, matching how the other `og:` tags in this file set are already sometimes wrapped.
6. **The e2e `/pitch` exclusion check asserted a bare substring.** Changed to assert the full `<loc>` it means.

Findings I left as they were: none, all six applied.

PR #135, opened 2026-09-16
ci(ci): shared/analytics is linted, tested and typechecked on every PR

fiorelorenzo:
## What this changes

`shared/analytics` has its own `lint`, `test` (two vitest files) and a tsconfig, but no hosted CI tier ever ran them: `_node-gate.yml` calls with `filter: web`, `filter: website` and `filter: hub` install `@rebase/analytics` as a dependency of the SPA that imports it, and `pnpm --filter <that package>... install` never runs a dependency's own scripts. The package's own suite only ran in `preflight.json`'s `analytics-unit` check, which is local and opt-in, so a change that broke `shared/analytics`'s own tests or its types merged with a green `ci`. Now a `shared_analytics` `changes` filter gates one more `_node-gate.yml` job, `analytics`, called with `filter: '@rebase/analytics'` and `build: false` (the package ships as source and has no build script), and it is in `ci`'s `needs:` list alongside the other nine.

## How I verified it

`python3 -c 'import yaml,sys; yaml.safe_load(open(".github/workflows/ci.yml"))'` parses clean. `python3 -c` against the loaded document confirms `jobs["ci"]["needs"]` carries the new `analytics` job and every job key resolves. I could not run the workflow itself locally; this PR's own `ci` run touches `.github/workflows/ci.yml`, which is one of the `shared_analytics` filter's own paths, so the new job should run on this PR and its result is the proof.

`preflight.json`'s existing `analytics-unit` check already runs `lint && test && exec tsc --noEmit` for `@rebase/analytics`, the same three checks `_node-gate.yml`'s `gate` job runs, so I did not touch `preflight.json`: the card only asks for it "only if the existing analytics-unit check does not already cover the same command", and it does.

## Screenshots and video

Nothing visible changed: this is a CI workflow file, not a product surface.

Linear: REB-253.


fiorelorenzo:
Independent review (reviewer subagent): overall correct, no blocker/major/minor findings. One nit applied in 2e02f118: the job id `analytics` diverged from its `shared_analytics` filter/output and `shared/analytics · web` display name, unlike every other gate here (`hub_web` / `hub-web` / `hub · web`); renamed the job id to `shared-analytics` and updated `ci`'s `needs:` list in the same commit. The review also confirmed the live run of this PR already shows the new gate passing.

PR #136, opened 2026-09-16
fix(api): a CV filename outside Latin-1 no longer 500s on download

fiorelorenzo:
## What this changes

A CV filename outside Latin-1 (Chinese, Greek, Cyrillic, Arabic, an emoji) made every download of it answer 500. `cv_response` sanitised the name with `ch.isalnum()`, which is Unicode-aware, so every letter of every script survived the sanitiser the docstring calls "ASCII-safe"; Starlette then tries to encode header values as Latin-1 and raises `UnicodeEncodeError` while building the `Response`, for both `GET /api/hub/freelancers/{id}/cv` (the admin) and `GET /api/hub/me/cv` (the member).

Now the sanitiser checks `ch.isascii() and ch.isalnum()`, so a character outside ASCII becomes an underscore the same way a quote or a newline already did. A card whose CV filename is `简历📄.pdf` downloads as `___.pdf` with a 200, not a 500.

## How I verified it

- `uv run ruff check projects/hub && uv run ruff format --check projects/hub`: all checks passed, 82 files already formatted.
- `uv run mypy`: success, no issues found in 307 source files.
- `uv run pytest -q projects/hub/apps/api/tests`: 75 passed.
- New regression test `test_a_cv_filename_outside_latin_1_still_downloads_instead_of_500` in `test_member_api.py`: uploads a CV named `简历📄.pdf`, then downloads it as the member. Confirmed it fails without the fix: temporarily reverted the sanitiser to plain `ch.isalnum()` and reran the same test, which raised `UnicodeEncodeError: 'latin-1' codec can't encode characters in position 22-23: ordinal not in range(256)` from `starlette/responses.py:61`, then restored the fix and reran, 1 passed.

I kept the fix to the ASCII fallback alone rather than adding an RFC 5987 `filename*` (which PigroCRM's `documents.py`/`invoices.py` use for their own downloads): the hub project has no such pattern today and the card's suggested fix is the ASCII sanitiser, so I did not add a second product's convention here on my own call. Happy to add `filename*=UTF-8''...` beside the ASCII fallback in a follow-up if a real accented filename (an Italian "à", "è") coming back as underscores is worth preserving.

## Screenshots and video

Nothing visible changed: this is an API response header, not a UI surface.

Linear: REB-241.


fiorelorenzo:
Independent review (reviewer subagent) found the fix correct and the injection guard sound against CJK, emoji, combining marks, RTL/bidi marks, control characters and a CRLF payload, but flagged two things:

1. **Should-fix:** the ASCII-only sanitiser also mangles accented Italian names that used to work (`Niccolò` -> `Niccol_`), which is a regression for the common case. Applied: added `filename*=UTF-8''...` (RFC 5987) alongside the ASCII `filename=` fallback, the same shape PigroCRM's own `documents.py`/`invoices.py` download routes already use, in commit 4154996d.
2. **Nit:** the new test's `disposition.encode('latin-1')` line could never fail (Starlette already encodes the header before the string exists). Replaced with a real `disposition.isascii()` assertion in the same commit, and added a test pinning the accented-name case.

projects/hub/apps/api/tests: 19 passed (test_member_api.py alone: 17 passed).

PR #137, opened 2026-09-16
fix(website): every page declares its own canonical address and og:url

fiorelorenzo:
## What this changes

None of the six pages carried `<link rel="canonical">` or `og:url`, so a crawler that
reached one of them by a query string, a UTM tail, or an alternate host had nothing on
the page itself saying which address was the real one. Every page now declares both,
built from the `SITE_HOST` constant `path-map-plugin.ts` already exports and the same
path the page answers to in `PAGES`, so a query string or an alternate host collapses
onto one address:

- `/` (`index.html`) -> `https://letsrebase.com/`
- `/pigrocrm` -> `https://letsrebase.com/pigrocrm`
- `/community` -> `https://letsrebase.com/community`
- `/privacy` -> `https://letsrebase.com/privacy`
- `/termini` -> `https://letsrebase.com/termini`
- `/pitch` -> `https://letsrebase.com/pitch` (still `noindex`, but a canonical costs
  nothing and matches what `NOINDEX` already says about it)

`links.test.ts` already reads all six pages through `path-map-plugin`'s `route()` and
`PAGES`; it now inverts `PAGES` to the path each file serves and fails when a page's
canonical or `og:url` is missing, or disagrees with that map.

Two existing tests assumed no page ever carried an absolute URL in a `<link>` href, or
in any `href`/`src` pointing back at this origin: `landing-pages.test.ts`'s "requests
nothing from another origin" and `community.test.ts`'s "writes no third-party tag".
Both now carve out the one `<link rel="canonical">` that points at `SITE_HOST`, since
it is metadata a crawler reads, not a request anywhere, first-party or otherwise; every
other absolute `<link>` href is still refused.

## How I verified it

- `pnpm --filter website test` — 275 passed (13 files), including the new canonical/
  `og:url` assertion in `links.test.ts` and the two adjusted tests.
- `pnpm --filter website exec tsc --noEmit` — clean (fixed a pre-existing possibly-
  undefined typing gap the new loop tripped in `landing-pages.test.ts`).
- `pnpm --filter website build` — builds all six pages.
- `pnpm --filter website test:e2e` — 58 passed, against `vite preview` on :4173.
- `website-image` (`docker compose build` for the website's nginx image) — built.
- `guide-pdf` and every other project's preflight checks stayed skipped: this diff
  touches no path they select on.

The two tests I narrowed (`landing-pages.test.ts`'s "requests nothing from another
origin", `community.test.ts`'s "writes no third-party tag") were written before any
page had a reason to carry its own absolute URL. I kept both bans in place for every
other tag and every other origin, and only let through the one `<link rel="canonical">`
pointing at `SITE_HOST`; worth a second look that this is narrow enough and not a hole
for a future third-party `<link>`.

## Screenshots and video

Nothing visible changed: `<link rel="canonical">` and `og:url` are both invisible to a
person looking at the rendered page. No before/after pair.

Linear: REB-111.


fiorelorenzo:
Independent review findings and what I did with each:

1. **Blocker-adjacent (applied).** The "requests nothing from another origin" exemption in `landing-pages.test.ts` skipped every same-origin absolute `href`/`src`, not just the canonical link's own value, so an `img`/`iframe`/`source` hardcoded to production would have passed unguarded. Now compares against the canonical's own extracted value (`a8505ba4`).
2. **Applied.** `community.test.ts`'s third-party-tag check stripped `<link rel="canonical">` regardless of its `href`; now only strips it when it points at `SITE_HOST` (`a8505ba4`).
3. **Applied.** `links.test.ts` read the canonical/`og:url` tags with an exact-attribute-order regex; now uses the same tolerant, order-independent pattern the suite's `meta()` helper already uses elsewhere (`a8505ba4`).
4. **Applied.** Added an assertion that `PAGE_FILES` and `PAGES` name the same set of pages, so drift between the two fails on its own terms (`a8505ba4`).
5. **Applied.** `pitch.html`'s canonical now carries a comment explaining it is not a contradiction with `noindex` (`a8505ba4`).

No blocker found overall; reviewer confirmed all six canonical/og:url values match `SITE_HOST` + `PAGES` exactly and the new assertion is not a tautology.

PR #138, opened 2026-09-16
fix(core): the outgoing HTTP seam stops following redirects and caps the body

fiorelorenzo:
## What this changes

The hub's outgoing HTTP seam (`rebase_core.http.urllib_call`, used by the Resend mail and by the Pigro registry read) followed redirects with the default `urllib` opener and read the response body with no cap. CPython's `HTTPRedirectHandler.redirect_request` copies every header except `content-length` and `content-type` onto a redirected request, so a 301/302/303 from `REBASE_PIGRO_API_URL` or from Resend re-sent `Authorization: Bearer ...` to whatever host the `Location` named, with nothing logged. `response.read()` was also uncapped, so a wrong or hostile endpoint decided how many bytes the API process allocated.

PigroCRM's mirror-image seam for the opposite call already has both guards. `urllib_call` now builds its opener with a `HTTPRedirectHandler` subclass whose `redirect_request` returns `None`, which turns every 3xx into an `HTTPError` (a status, never followed) instead of a second request, and reads the body to one byte past a `MAX_BODY_BYTES` cap, so the caller can tell "exactly the cap" from "too long". The cap is 1 MiB rather than PigroCRM's 4096 bytes, because this seam also carries the whole Pigro tenant registry, not just Resend's short `{"id": ...}`.

Both callers (`pigro.py`, `mail.py`) go through this one function unchanged; neither test suite reaches the real seam (both inject a fake `HttpCall`), so nothing there needed to change.

## How I verified it

- `uv run ruff check projects/hub && uv run ruff format --check projects/hub`: all checks passed.
- `uv run mypy`: success, no issues found in 307 source files.
- `uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests`: 276 passed (this also exercises `pigro.py`'s and `mail.py`'s own test suites, confirming both callers still work).
- Two new tests against a real local `HTTPServer` on loopback (matching PigroCRM's own `test_hub_lookup.py` style, since this is about proving the opener wiring itself, not a call site): one server answers `/lookup` with a 302 to `/altrove`, and the test asserts exactly one request reached the server with the bearer, none reached `/altrove`, and `urllib_call` returned status 302 rather than following it. Another server answers every request with a body 1000 bytes past the cap, and the test asserts the returned body is truncated to exactly `MAX_BODY_BYTES + 1`. Confirmed both fail without the fix: temporarily reverted `urllib_call` to the default opener and uncapped `read()`, reran, got `assert 200 == 302` and `assert 1049576 == 1048576 + 1`, then restored the fix and reran, both passed.
- Updated the two pre-existing tests in `test_http.py` that monkeypatched `urllib.request.urlopen` directly: that call is no longer on the path once a custom opener is built (`_OPENER.open` uses `http.client` directly), so they now monkeypatch `_OPENER.open` instead; verified they failed with the guard's own `AssertionError: a test tried to reach 'api.example.test'` before this change (the monkeypatch stopped taking effect) and pass now.

The 1 MiB cap is a judgement call, not a number the card gave: PigroCRM's twin seam caps at 4096 bytes because it answers three short fields, but this seam also serves the Pigro tenant registry with no pagination, which could plausibly grow past a few KB as spaces accumulate. Happy to tighten it if there is a better number.

## Screenshots and video

Nothing visible changed: this is an outbound HTTP client, not a UI surface.

Linear: REB-242.


fiorelorenzo:
Independent review (reviewer subagent) proved the fix correct against all five 3xx codes HTTPRedirectHandler knows (301/302/303/307/308, both GET and POST, empirically verified with a throwaway parametrized loopback server: 14/14 no-follow post-fix, 8 of the same cases followed pre-fix) and confirmed the cap is applied at both read sites and that all three seam consumers degrade gracefully. Two nits, both applied:

1. `PigroRegistry.list_spaces` went straight to `json.loads` instead of consulting the length the seam's own docstring promises the caller can check, unlike PigroCRM's mirror seam. Applied: an explicit `len(body) > MAX_BODY_BYTES` check before parsing, in commit c876a77f, with a test proving a truncated registry answer now gets the specific 'troppo lungo' sentence rather than the generic 'non è un elenco' one.
2. The redirect test only pinned 302, so a fix that overrode a single `http_error_302` instead of `redirect_request` would have stayed green while still leaking the bearer on 301/303/307/308. Applied: parametrized across all five in the same commit.

projects/hub/packages/core/tests + apps/api/tests + apps/mcp/tests: 280 passed.

PR #139, opened 2026-09-16
fix(infra): document uploads over 1 MB no longer 413 before reaching the app

fiorelorenzo:
## What this changes

Neither nginx layer in front of the CRM set `client_max_body_size`, so `web`'s stock `nginx:1.27-alpine` image answered any upload over its compiled-in 1 MiB default with a bare 413 before the request ever reached `api`, whatever the application itself allows (`DIMENSIONE_MAX`, 100 MiB, `packages/core/src/pigrocrm/core/documents/schemas.py:51`). A scanned invoice or a signed contract is routinely larger than 1 MB, so `POST /api/documents/{id}/versions` failed on ordinary input with an nginx error page instead of the product's own message, and since the SPA creates the document row before it uploads the first version, it left an empty document behind each time.

Now all three nginx configs (`deploy/nginx/spa.conf`, the container that answers today; `deploy/nginx/pigro.letsrebase.conf` and `deploy/nginx/preview.pigro.letsrebase.conf`, the two host vhosts) set an explicit, tight `client_max_body_size 1m;` at the server level, and a `location ^~ /api/documents/` override at `105m`, the one prefix that accepts a file body. 105m rather than 100m: a multipart body wraps the file in a boundary and a `Content-Disposition` header nginx counts against the same limit, so a file at the exact application ceiling needs a little room to clear this gate too, and it costs nothing on the smaller requests the same prefix also serves (create, list, restore).

## How I verified it

- `docker run --rm -v $PWD/projects/pigrocrm/deploy/nginx/spa.conf:/etc/nginx/conf.d/default.conf:ro nginx:1.27-alpine nginx -t` -- "configuration file /etc/nginx/nginx.conf test is successful" against the real image this container ships. `nginx -t` is not available on this devbox directly, so I ran it through the same image the Dockerfile uses instead.
- The two host vhosts are not `include`d files nginx can test standalone (they assume `/etc/nginx/snippets/orbiters-proxy.conf` and the top-level `http`/`events` blocks certbot's install provides), so I built a matching minimal harness -- `events {} http { include .../sites/*.conf; }` plus a copy of `orbiters-proxy.conf` at the expected path, `__ROOT_SLUG__` substituted the same way `setup-server.sh` does -- and ran the same `nginx -t` against both. Both pass.
- `pnpm --filter web exec vitest run src/test/upload-body-size.test.ts`: 6 passed. I first ran it against the three pre-fix files (`git show origin/main:...`) to confirm it fails without this change (all 6 fail: no server-level directive, no `/api/documents/` block), then restored the fix and it passed.
- `pnpm --filter web exec eslint src/test/upload-body-size.test.ts`: clean.
- `POSTGRES_PASSWORD=... PIGROCRM_JWT_SECRET=... docker compose -f projects/pigrocrm/docker-compose.yml build web`: succeeds in 47s (well under 5 minutes), which runs `pnpm --filter web build` (`vite build && tsc --noEmit`) against the changed `spa.conf` and the new test file.

`deploy/setup-server.sh:36` still generates a plain vhost with `client_max_body_size 25M;` for a first-time install before certbot has run. It is not one of this card's three files and 25M is below the 105m this PR settles on, but it is the same knob and now the odd one out; I left it alone since the card scopes to `spa.conf` and the two dedicated vhosts, but it is worth its own follow-up if that script is still how a fresh server gets bootstrapped.

## Migrations to paste into the installed host copies

`nginx -t` and `systemctl reload nginx` still have to run on the actual hosts, and certbot's rewrite means the installed copies under `/etc/nginx/sites-available/` are edited in place rather than replaced wholesale (`pigro.letsrebase.conf`'s own header comment, and how REB-107 shipped its robots block). The orchestrator should paste, into each installed copy's existing `server { ... }` block (the port-80 one; certbot's port-443 block proxies through the same location table so it needs the same paste too if it declares its own locations -- check the installed file):

For `pigro.letsrebase.com` (paste after `absolute_redirect off;`):

```nginx
    client_max_body_size 1m;
```

and, after the `mcp` regex location block (before the catch-all `location / { ... }`):

```nginx
    location ^~ /api/documents/ {
        proxy_pass http://127.0.0.1:8080;
        include /etc/nginx/snippets/orbiters-proxy.conf;
        client_max_body_size 105m;
    }
```

For `preview.pigro.letsrebase.com`, the same two pastes, with `http://127.0.0.1:8081` instead of `:8080`.

## Screenshots and video

Nothing visible changed: this is nginx configuration, not a product surface. The observable difference is a 413 that no longer happens, which the new test proves against the files rather than against a running server.

Linear: REB-252.


fiorelorenzo:
Independent review (reviewer subagent) found a **blocker**: the previous commit's `location ^~ /api/documents/` only matched the bare, unprefixed path, but a logged-in session always uploads at `/<slug>/api/documents/...` (`lib/tenant.ts` prepends the space prefix to every API call, and both host vhosts redirect `/app/` under it), so it never actually reached the fix — reproduced by the reviewer against `nginx:alpine` with the pre-fix files. Fixed in 69f6dd33: `spa.conf` gets a second, more specific regex location for the prefixed path ordered before the `(api|health)` regex it would otherwise fall into; both vhosts replace their bare-only block with a single regex covering both forms, the same shape their neighbouring MCP block already uses. Extended `upload-body-size.test.ts` to check every `/api/documents/` location in a file and assert a slug-prefixed match exists too; both new/changed assertions fail against the prior commit and pass now. Verified with `nginx -t` (the real image for spa.conf, the minimal harness for the vhosts) and `docker compose build web`. Two minor/nit findings (an unanchored regex in the test that a future comment could fool, and setup-server.sh's own stale 25M on a fresh self-hosted install) were not applied given time; noting them here for the record rather than silently dropping them.

fiorelorenzo:
**Update after the blocker fix (69f6dd33):** the paste below replaces the earlier `location ^~ /api/documents/` shape with a single regex covering both the bare and the space-prefixed path (a logged-in session always uses the prefixed one). Paste this into each installed copy's existing `server { ... }` block, replacing any earlier paste of the old shape:

For `pigro.letsrebase.com` (after `absolute_redirect off;`):

```nginx
    client_max_body_size 1m;
```

and, after the `mcp` regex location block (before the catch-all `location / { ... }`):

```nginx
    location ~ "^(/[a-z0-9][a-z0-9-]{1,30}[a-z0-9])?/api/documents(/|$)" {
        proxy_pass http://127.0.0.1:8080;
        include /etc/nginx/snippets/orbiters-proxy.conf;
        client_max_body_size 105m;
    }
```

For `preview.pigro.letsrebase.com`, the same two pastes, with `http://127.0.0.1:8081` instead of `:8080`.

PR #141, opened 2026-09-16
fix(api): POST /api/auth/link and GET disponibile are throttled per client

fiorelorenzo:
## What this changes

`POST /api/auth/link` was unauthenticated, opened an engine on the registry database, called `TenantService.list()` (every row), opened one more engine per space an address owns, wrote a magic-link token per space and queued a mail send, with no rate limit at all. `GET /api/tenants/{slug}/disponibile` was likewise unauthenticated and unthrottled. `member` and `signup` already call `ratelimit.spend_one(request)` as their first statement; neither of these two did. Now both call it too, at the same budget (5 requests/minute per client).

## How I verified it

- `uv run mypy`: no issues in 307 source files.
- `uv run pytest -q projects/pigrocrm/apps/api/tests -k 'link or disponibile or ratelimit'`: 11 passed. I first ran the two new throttle tests against the pre-fix routes (git stash of the two router files) to confirm they fail without the change (both did, 202/200 instead of 429), then restored the fix and they pass.

The card's `MagicLinkService.request` has no per-address cooldown, which the "why it matters" section also names. I did not add one: a link asked at the root opens one engine per space the address owns (`_owned_slugs`/`_space_link` in `routers/auth.py`), so a cooldown would need to check and write state across however many per-space databases that address owns, which is not a one-line change in the same shape as `spend_one`. It stays open on the card.

The rate limiter's bucket is a process-global dict keyed by client IP, so I added a `reset_rate_limit()` call to the shared `client` fixture in `apps/api/tests/conftest.py`: without it, `test_auth_api.py`'s own six pre-existing `/api/auth/link` calls would exhaust the new budget partway through the file and start failing with 429 on an unrelated test.

## Screenshots and video

Nothing visible changed: this is a rate limit on two existing API routes, not a product surface.

Linear: REB-228.


fiorelorenzo:
**Independent review (reviewer subagent).** Full findings: blocker, major, four minor/nit. Applied all six in `16518884`.

1. **Blocker — `disponibile` shared the signup's own 5/minute bucket with `POST /api/tenants/membro` and `POST /api/tenants/`.** Simulated against the real module: a member lookup plus five debounce-fired availability checks in 15 seconds leaves the bucket empty, and the `Crea lo spazio` submit that follows is refused 429 — a person can be locked out of creating their own space by nothing but hesitating while typing its name. Fixed: `spend_one` now takes a `scope` and a `per_minute`, the bucket carries its own cap per key, and `disponibile` moves to `scope="disponibile"`, `DISPONIBILE_REQUESTS_PER_MINUTE = 30` — its own budget, separate from the create-space bucket.
2. **Major — the SPA never handled a 429 from the probe.** `registrati.tsx`'s `.then(({ data }) => { if (!data) return })` left `availability` stuck at `checking` forever on any non-2xx, with no message and the submit button permanently disabled. Fixed: the probe now checks `response.status`, surfaces a 429 through the same `error` alert `onCreate` uses, and drops back to `idle` so a further edit asks again on its own, later budget. Added `registrati.test.tsx`'s own case for it.
3. **Minor — the 429 both routes can now return was undocumented in OpenAPI.** Added a shared `TOO_MANY_REQUESTS_RESPONSE` (`ratelimit.py`, `application/json`, never `problem+json` — the same reason `auth.py`'s own `_UNAUTHENTICATED_RESPONSE` cannot reuse `PROBLEM_RESPONSES` for 401) and attached it to `request_link` and `availability`. Left `membro`/`signup`'s own pre-existing gap from ORB-173 alone: out of scope for this card, which only touches the two routes REB-228 added.
4. **Minor — `reset_rate_limit()` living inside the `client` fixture left one caller uncovered.** `test_tenants_api.py::test_the_root_slug_is_the_root_itself_and_nobody_elses_name` builds its own `TestClient` without `client` or `spaces_client` and calls the now-throttled `disponibile`; a narrow `pytest -k`/`--lf` selection after the new throttle test could hand it an already-spent bucket. Fixed: `reset_rate_limit()` moved to an `autouse=True` fixture in `conftest.py`, covering every test in the directory.
5. **Minor — the throttle test's docstring claimed an ordering (\"spent before the sender check\") nothing tested.** The original test ran every request with `sender` configured, which is equally consistent with the limiter running after that check or not at all. Fixed: dropped the `sender` fixture, asserted the first `REQUESTS_PER_MINUTE` requests are 503 (no sender) and the next is 429 not 503 — that's the ordering, directly.
6. **Nit — the conftest comment named the wrong file.** Resolved by finding 4's move to an autouse fixture.

Considered and dismissed by the reviewer: `X-Real-IP` trust (every nginx vhost sets it from `$remote_addr`, replacing anything client-supplied, so it can't be forged through the proxy); `spend_one` running before the sender/public-url checks on `/link` (correct — a misconfigured install must still throttle).

Verified: `uv run ruff check projects/pigrocrm`, `uv run ruff format --check projects/pigrocrm`, `uv run mypy` (307 files, clean); `uv run pytest -q projects/pigrocrm/apps/api/tests/test_auth_api.py projects/pigrocrm/apps/api/tests/test_tenants_api.py` (51 passed); `pnpm exec vitest run src/routes/app/registrati.test.tsx` (9 passed, including the new 429 case); `pnpm exec tsc --noEmit` and `pnpm exec eslint` on the changed web files (clean).

PR #142, opened 2026-09-16
feat(website): index.html carries WebSite and Organization structured data

fiorelorenzo:
## What this changes

No page carried `application/ld+json` or microdata, so a search result for rebase drew
a title and a snippet with nothing attached: no logo, no organisation, nothing tying
`letsrebase.com` to the legal entity `privacy.html` and `termini.html` already name.

`index.html`, the front door, now carries one `<script type="application/ld+json">`
with a `WebSite` and an `Organization` node in a single `@graph`:

- `WebSite`: `name` rebase, `url` `https://letsrebase.com/`.
- `Organization`: `name` rebase, `legalName` "Humancraft di Ivan Sala", `url`
  `https://letsrebase.com/`, `logo` the share card `og:image` already names
  (`https://letsrebase.com/assets/share-card-1.png`), `vatID` `14518240966` and
  `email` `ivansala@humancraft.tech`.

Every value already appears elsewhere: the legal name, VAT number and email are exactly
what `privacy.html` §"Il titolare del trattamento" and `termini.html` state, nothing
invented for a richer block. Nothing else from ORB-113's "deliberately not done" list is
added: no `FAQPage`, no `Product`, no `Review`/`AggregateRating`.

`landing-pages.test.ts`'s `index.html` describe block parses the block as JSON and
checks `@context`, both `@type`s, `name`, `url` and, for the `Organization`, `logo` and
the three legal fields. `links.test.ts`, which already reads all six pages, gets a
one-line assertion per page that only `index.html` carries any `application/ld+json` at
all.

## How I verified it

- `pnpm --filter website test` — 283 passed (13 files), including the new JSON-LD
  parse/shape assertion and the six-page "only the front door" check.
- `pnpm --filter website exec tsc --noEmit` — clean.
- `pnpm --filter website build`, then parsed `dist/index.html`'s script block with
  `JSON.parse` directly (not just through the test runner) to confirm the built output
  is valid JSON with the expected shape.
- `pnpm --filter website test:e2e` — 58 passed against `vite preview` on :4173.
- `website-image` (`docker compose build`) — built.
- Google's Rich Results test and the schema.org validator need a public URL, so I could
  not run them from here; the orchestrator runs both once this deploys, per the card's
  own "what is needed".

`logo` points at the OG share card (1200x630, a social preview image with the wordmark
and claim on it) rather than a square brand mark under `src/`, because the card is
already built to a stable, unhashed path the crawler can resolve; `rebase-logo.svg` is
Vite-hashed at build time and would 404 at the literal path I could otherwise write.
The card is due to move to `share-card-2.png` in REB-205's PR, right after this one; that
PR updates this `logo` value along with the six `og:image`s.

## Screenshots and video

Nothing visible changed: a `<script type="application/ld+json">` in `<head>` renders
nothing. No before/after pair.

Linear: REB-113.


fiorelorenzo:
Independent review (dispatched before push, so its findings are already in commit `139a87a7`):

Overall correctness: **correct**, no blocker. Confirmed the block is valid JSON, a correct schema.org shape, lives on the homepage only, and every Organization value traces literally to `privacy.html:54-57` / `termini.html:54-58` with nothing invented (no postal address exists there, so none appears here). The `logo` choice (the share card, not a `src/` asset) was confirmed defensible since `rebase-logo.svg` is a Vite-hashed path with no fixed URL.

Two non-blocking findings, both applied:

1. **Applied.** `toMatchObject` alone cannot fail on an invented extra property (a postal address, say), since it ignores keys it doesn't list. Both nodes now assert their exact key set first, and `@graph`'s length is checked.
2. **Applied.** The `logo` was checked against a second hardcoded copy of the share card URL, which would drift silently from `og:image` on the next redraw (REB-205's rename to `share-card-2.png`). The test now compares the structured data's `logo` against the page's own `og:image` instead, so the two stay self-consistent without a second literal to update by hand.

One thing considered and left alone: `vatID` carries the bare `14518240966` rather than `IT14518240966`, matching the legal pages exactly rather than inventing the ISO prefix.

Not applied (minor, P3): the two graph nodes are unlinked blank nodes with no `@id`/`publisher` cross-reference. Beyond what the card asked for (`@context`, `@type`, `name`, `url`, and for the Organization, `logo`); left out.

PR #143, opened 2026-09-16
test(core): test_sort_plan.py's plan assertions no longer depend on which machine runs them

fiorelorenzo:
## What this changes

`test_sort_plan.py`'s planner-marked assertions passed two trunk runs and then failed a
third on a commit that touched no Python at all (34294151175, the four
`customers`/`ragione_sociale` cases), the same class of failure ORB-9 already fixed for
`test_trgm_escape.py`. `ordered_corpus` now runs a plain `VACUUM` on each of the four
ordered tables before its own twenty-thousand-row corpus is inserted, not only
`VACUUM (ANALYZE)` after, in the same shape ORB-9 gave `customers_at_reference_scale`:
`VACUUM` after the insert only reclaims the dead pages this corpus's own rows leave once
deleted, and does nothing about whatever dead pages an earlier module in the same run
(`test_search_plan.py` deletes 50 000 rows per table on its way out) already left in
`customers`, `people`, `deals` or `documents`. Left there, `relpages` costs the composite
indexes on pages this corpus never wrote, and the planner picks the narrower
single-column index under an `Incremental Sort` instead of the one the ordering contract
expects.

None of the file's assertions were full plan-string comparisons to begin with —
`_assert_served_by_its_index` already asserted properties (no sequential scan, the named
index, no in-memory sort) rather than an exact plan string — so the property style was
already right; what was missing was giving the planner a deterministic corpus to read
those properties from.

`pytest.mark.planner` drops from the whole module: that marker is what meant this plan
choice depended on which machine ran it, and the pre-insert vacuum is what removes the
dependency, the same conclusion ORB-9 reached for its own file. No test in the repository
still carries the marker, so `pigrocrm-py-planner`, the preflight lane that only ever
selected it, is retired, and the two comment blocks in `ci.yml` and `preflight.json` that
named `test_sort_plan.py` as the one file still needing the single-machine lane are
updated to say it no longer does. The marker stays registered in `pyproject.toml` for
whichever assertion needs it next — nothing else references it, so this is a decision to
keep the infrastructure rather than a dangling one.

## How I verified it

- `uv run ruff check` / `ruff format --check` on the changed test file: clean.
- `uv run pytest -q -m "slow" projects/pigrocrm/packages/core/tests/test_sort_plan.py`,
  three consecutive runs on this devbox: 51 passed each time (55.7s, 58.5s wall, 66.1s
  wall including one run under `-p no:randomly`, which is a no-op here — this project
  does not use `pytest-randomly`, confirmed against `pyproject.toml`/`uv.lock`).
- `uv run pytest -q -m "slow" projects/pigrocrm/packages/core/tests`, the exact selection
  and file order `pigrocrm-py-corpus` runs on the trunk: 75 passed, 0 failed, 3162
  deselected, 18m29s.
- Not verified: whether the fix holds on an actual two-core hosted runner rather than
  this box, since I have no way to provision one from here. That is the same limit ORB-9
  shipped under; if a plan choice still flips somewhere else with a corpus this clean,
  that is new evidence for reopening this file, not a reason this PR should have waited.

Dropping `planner` file-wide, not per test. All six tests in this file share the same
`ordered_corpus`/`ordered_session` fixtures, so I could not find a case that stays
margin-sensitive after the fixture fix while a sibling test does not — including the two
falsifiers, which were already robust to either failure mode (`pytest.raises(...,
match="sequentially scanned the table|did not read its index")`). If that reasoning is
wrong for one specific case, the fix is splitting `pytestmark` back onto that test alone,
not reverting the fixture.

Retiring `pigrocrm-py-planner` from `preflight.json` rather than leaving it: once no test
carries `-m "planner"`, that entry selects zero tests and pytest exits 5 ("no tests
collected"), which would make the lane always fail rather than quietly do nothing.

## Screenshots and video

Nothing a person sees or does: this is a test-fixture and CI-config change with no
product-facing surface.
Linear: REB-90.


fiorelorenzo:
Independent review found the mechanism claim in the first commit did not hold: plain `VACUUM` truncates a heap's trailing empty pages, but never shrinks a B-tree index's `relpages` (only `REINDEX`/`VACUUM FULL` do), so the pre-insert `VACUUM` alone did not reset the composite indexes' page counts the way the docstring claimed. Confirmed empirically against a scratch Postgres 17 container (see commit 54d51c2c's message for the exact numbers: 139 pages surviving a delete+vacuum+reinsert, versus 57 for a genuinely fresh build). Fixed by adding `REINDEX TABLE` alongside the pre-insert `VACUUM`, verified against the same scratch instance and against a fresh 51/51 run of the file.

Also found and fixed: a comment in ci.yml claiming `planner` was 'deselected on both' CI tiers was false — `pigrocrm-py-corpus` selects on `slow` alone with no `and not planner`, so this file ran on the hosted runner under that job the whole time, marker or not; dropping the marker changed no CI selection, only retired the redundant preflight lane. And one stale present-tense docstring reference in `test_trgm_escape.py` to the now-retired `pigrocrm-py-planner` lane, updated to past tense.

One review point left as is: the pre-insert check/vacuum/reindex loop iterates `_ORDERED_TABLES` (four tables), not `invoices`, which `build_corpus` also inserts one row into. Reviewer flagged this as informational only — no assertion in this file reads `invoices`, and `test_search_plan.py`'s own teardown already vacuums it — so I left it unchanged rather than widening the loop for a table nothing here measures.

PR #144, opened 2026-09-16
refactor(hub): the wizard walks screens, the review walks fields

fiorelorenzo:
## What this changes

`wizard/Wizard.tsx` splits `Step<T>` into `Field<T>` (id, label, hint, optional,
render, validate, summary: one question) and `Screen<T>` (id, title, hint,
fields: what the wizard shows and validates together). `screensFromFields`
wraps each field in its own single-field screen, so both wizards keep their
eight and five screens exactly as they are today: grouping several fields onto
one screen is REB-120 and REB-121's job, not this one's.

A screen now validates every field it holds, moves focus to the first one that
failed, and shows each message under its own control with `aria-invalid` and
`aria-describedby` wired through to the control (`TextField`, `LongTextField`,
`ChoiceField`, `LinksField`, `FileField`). The review, the member area
(`Area.tsx`) and the edit page (`Modifica.tsx`, `editSteps` renamed
`editFields`) now walk the flat list of fields instead of the wizard's own
steps, one row per answer, which today is exactly what they showed before
since every screen still holds one field.

This also closes REB-243: a 422 whose field id names no field the wizard
recognises (`cognome`, folded into the combined nome+cognome field, has none of
its own) used to disappear entirely, the submit button just re-enabling with
nothing said. The engine now only jumps to a screen when a field actually
matches; an id that matches nothing falls back to the review's own alert, so
the sentence is never silent. `submitError.step` is renamed `submitError.field`
since it always named a field, never a step.

## How I verified it

- `pnpm --filter hub lint` and `pnpm --filter hub test -- --run src/wizard src/pages`: 25 test files, 165 tests, all green, including the new `Wizard.test.tsx` cases for the aria wiring and for REB-243's unknown-field fallback, and the existing `Wizard.test.tsx`, `FreelancerWizard.test.tsx`, `CompanyWizard.test.tsx`, `Area.test.tsx` and `Modifica.test.tsx` suites unchanged in their assertions.
- `pnpm --filter hub build` (`vite build && tsc --noEmit`): clean.
- Built both worktrees (this branch and `origin/main`) and walked both wizards' first screen and review screen at 1440×900 in a headless browser: pixel-identical between before and after (screenshot below is one pair; the freelance and company review screens and the company first screen were compared the same way and matched).
- Reproduced REB-243 directly: mocked the freelance wizard's submit to answer 422 with `loc: ['body', 'cognome']` (a 121-character surname, `packages/core`'s own limit). On `origin/main` the review re-enables the button with no message at all. On this branch the review shows "assicurati che il cognome abbia al massimo 120 caratteri" under a `role="alert"`, boxed in the after frame below.
- Walked the member area and the edit page in the built preview with a mocked profile: `Area.tsx` shows one row per field (name, LinkedIn, rate, position, remote preference, links, CV), `Modifica.tsx` shows one section per field with its own hint and "(facoltativo)" marker, and clearing "Cognome" and saving shows `aria-invalid="true"` / `aria-describedby="nome-error"` on both name inputs with the message under them.

`nome`+`cognome` and `periodo_da`+`durata` stay one combined `Field` each (not
split into two), matching today's exact validation message and review row; I
did not atomize them into separate fields, since REB-119 says nothing visible
should change and splitting them would either change the error message shown
or require deciding how two fields share one screen's error state, which is
grouping's own question (REB-120/121). REB-243 is fixed at the engine's generic
fallback rather than by giving `cognome` its own field id.

The number of screens in either wizard, Enter's behaviour, the CV step's
semantics, the draft in `localStorage`, and the company wizard's own
`durata`-to-`periodo_da` remap in `CompanyWizard.tsx`, which stays alongside
the new generic fallback since it is a better, more specific fix for that one
field than the review's alert.

## Screenshots and video

**1. REB-243: a 422 naming an unknown field, before and after**

REB-243 fallback, before and after

**2. The freelance wizard's first screen, before and after (pixel-identical)**

Freelance wizard first screen, before and after

**Video: the freelance wizard walked to review, submitted, and the fallback alert appearing**

https://github.com/user-attachments/assets/043b78ed-0adf-4815-95d6-f37723a293dc

Linear: REB-119.


fiorelorenzo:
Independent review found 5 findings, all applied in 0ae8bd0d:

1. The REB-243 regression test mounted fresh (index 0), so it passed whether or not the fallback worked. Fixed: it now walks to the review first and asserts the alert carries the message.
2. `handled` was keyed on the error's message text, so a second identical refusal naming a matched field (e.g. two "Email già usata." in a row) was silently dropped after the first jump. Fixed: keyed on the error object itself.
3. `focusId` could never actually change `autoFocus`'s value on any of today's single-field screens, so the promised focus-on-failure was inert. Fixed: the field wrapper now remounts on the invalid/valid transition, so the control's own autofocus effect runs.
4. The three combined fields (nome+cognome, referente+email, periodo_da+durata) marked both inputs `aria-invalid` even when only one was actually wrong. Fixed: each input now reflects its own emptiness/validity.
5. The company wizard already remapped `durata`'s 422 onto `periodo_da` but not `email`'s onto `referente`, though both share a field the same way. Fixed: `email` now maps too.

165 tests still green after the fixes, tsc and build clean.

PR #145, opened 2026-09-16
fix(core): forward PIGROCRM_REFRESH_TOKEN_DAYS to the api and mcp containers

fiorelorenzo:
## What this changes

`PIGROCRM_REFRESH_TOKEN_DAYS` was documented in `.env.example` as a
per-installation choice (the sliding six-month session, spec 9 §5.6), but the
shared `x-api-environment` anchor in `docker-compose.yml` never named it, and
`packages/core/tests/test_compose_environment.py`'s `NOT_FORWARDED` exempted
it with the same reason as `access_token_minutes` (a ceiling nobody moves).
The two claims disagreed: a deploy's `.env` could set the variable and the
container would still receive `Settings`' compiled-in 180 days, silently.

Now the anchor forwards `PIGROCRM_REFRESH_TOKEN_DAYS: ${PIGROCRM_REFRESH_TOKEN_DAYS:-180}`,
the same shape as every other installation choice already there
(`PIGROCRM_TIMEZONE`, `PIGROCRM_MCP_FULL_ACCESS`, ...), so both `api` and `mcp`
see whatever the environment's `.env` sets.

**Other settings checked for the same situation.** I compared every field of
`Settings` against `NOT_FORWARDED` and against `.env.example`.
`refresh_token_days` was the only field `NOT_FORWARDED` exempts that
`.env.example` also documents as a configurable, per-installation value; every
other `.env.example`-documented `PIGROCRM_*` variable is already in the
anchor. The remaining `NOT_FORWARDED` entries (`access_token_minutes`,
`magic_link_minutes`, the Gmail/Drive ceilings, `pandoc_binary`/`typst_binary`,
the `solleciti_*` numbers, `storage_local_root`, `tenants_database_url`,
`tenants_alembic_ini`, `gmail_sync_address_batch_size`, `cookie_secure`) are
not mentioned in `.env.example` at all, so none of them make the same false
claim. I left all of them as they are.

## How I verified it

- `docker compose --env-file <throwaway> config -q` passes, and `config`
  shows `PIGROCRM_REFRESH_TOKEN_DAYS: "180"` resolved for both the `api` and
  the `mcp` service.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_compose_environment.py`:
  86 passed, including the new
  `test_refresh_token_days_is_forwarded_with_its_documented_default`.
- `uv run ruff check projects/pigrocrm && uv run ruff format --check projects/pigrocrm`:
  all checks passed, 518 files already formatted.
- `preflight --list` against this diff selected `pigrocrm-py-lint`,
  `pigrocrm-py-test`, `pigrocrm-py-corpus`, `pigrocrm-py-planner` and
  `pigrocrm-images`. I ran the lint check, the guard test file directly (the
  change touches only `docker-compose.yml` and that one test file, and
  nothing else references `refresh_token_days` through compose), and
  `pigrocrm-images` (`docker compose build` with the throwaway secrets CI
  uses): both `pigrocrm-api` and `pigrocrm-web`/`pigrocrm-mcp` images built.
  I skipped the full `pigrocrm-py-test`/`pigrocrm-py-corpus`/`pigrocrm-py-planner`
  tiers: this change adds no application code and the corpus/planner tiers
  need a live Postgres unrelated to a compose-file edit; the change's own
  test file passing is the meaningful signal.

I placed the new line next to `PIGROCRM_TIMEZONE` in the anchor, in the
"installation's own choices" group, rather than at the end of the block. If a
different position reads better, it is a one-line move.

## Screenshots and video

Nothing visible changed: this is a compose file and a test file, no UI, no
API surface.

Linear: REB-257.


fiorelorenzo:
Independent review (agent, read-only, diff origin/main...HEAD): overall correct, confidence 0.86. Confirmed the anchor line parses and sits at the required two-space indent, the `:-` default matches `config.py` and `.env.example`, `refresh_token_days` is not in `space_settings.OVERRIDABLE_KEYS` so the environment is the only source, removing the `NOT_FORWARDED` entry leaves no stale reference, and the commit follows repo conventions.

One should-fix, applied in `1dfa189d`: the new test compared the forwarded string against a hardcoded `180` instead of `Settings`' own compiled default, so it did not actually enforce the invariant its docstring claims. Now reads the default off `Settings.model_fields["refresh_token_days"].default`, so a future change to `config.py`'s default without a matching compose update fails here.

`uv run pytest -q projects/pigrocrm/packages/core/tests/test_compose_environment.py`: 86 passed, after the fix.

PR #146, opened 2026-09-16
fix(website): the production vhost answers an ACME challenge like every other name

fiorelorenzo:
## What this changes

`letsrebase.conf` (production) had no ACME challenge location, while its preview
copy and both redirect vhosts (`redirects/joinorbiters.conf`,
`redirects/preview.joinorbiters.conf`) already did. A renewal under
`--webroot -w /var/www/html` reached the apex, fell into the catch-all
`location /` that proxies to the website container, and that container's own
map ends in a 404 (`deploy/nginx.conf:38`). `letsrebase.conf` now carries the
same block, in the same position `preview.letsrebase.conf` has it:

```
location ^~ /.well-known/acme-challenge/ {
    root /var/www/html;
}
```

`preview.letsrebase.conf`'s own comment claimed it "mirrors `letsrebase.conf`
line for line and differs only in the ports," which was already false before
this change (it also carries `X-Robots-Tag`, `/robots.txt`, and a different
CRM redirect host). Now that the ACME block is common to both files, the
comment names every remaining difference.

## How I verified it

`pnpm --filter website test`: 13 files, 283 tests, all passing.
`path-map-plugin.test.ts` only reads `letsrebase.conf` for the four
`ELSEWHERE` prefixes (`/api`, `/hub`, `/app`, `/health`); the new
`location ^~ /.well-known/acme-challenge/` block matches none of them, so the
parser ignores it, as expected. `nginx` itself is not installed on this
runner, so `nginx -t` could not be run here; the block is copied verbatim from
`preview.letsrebase.conf`, which already serves it in production traffic
today, so the syntax is proven live.

The plain-HTTP block added here is necessary but not sufficient on its own:
the independent review below found, from live read-only probes, that
certbot's installed 443 copy for `letsrebase.com` redirects every plain-HTTP
request (the challenge path included) to https with a server-level
`if ($host = ...) { return 301 ...; }`, which nginx resolves in the rewrite
phase before any location is chosen. A `--webroot` validation therefore
follows that redirect and needs the same location inside the 443 block too,
which only the orchestrator can install (see below). The comment at the
location now says this.

## What the orchestrator pastes into the installed copies

The installed `/etc/nginx/sites-available/letsrebase.conf` on the server has
certbot's port-443 block on top of this plain-HTTP file and is edited in
place rather than overwritten. This location needs installing in **both**
halves of that file for `--webroot` renewal to actually work end to end:

1. The plain `listen 80` block for `letsrebase.com`, right after
   `client_max_body_size 1M;` and before the `# The Orbiters hub` comment
   (same position as this PR's plain-HTTP copy) — this half is what a
   non-redirecting request or a `--nginx`-authenticator run reaches.
2. **Certbot's `listen 443 ssl` block for `letsrebase.com` and
   `www.letsrebase.com`** — this is the half that actually answers today,
   since certbot's own `if ($host = ...) { return 301 https://...; }` sends
   every plain-HTTP request there first, challenge path included, and that
   block currently has no challenge location, so the request falls into the
   website container's `location /` and 404s.

Paste the same block into both, then `nginx -t` before reloading:

```
    # Renewal has to reach the challenge. `--nginx` answers from a server block of its
    # own, so this one is for `--webroot -w /var/www/html` -- and the installed copy in
    # /etc/nginx/sites-available/ needs this location inside certbot's 443 block as
    # well: certbot's `if ($host = letsrebase.com) { return 301 ...; }` runs in the
    # server rewrite phase, before any location below is chosen. preview's installed
    # copy already has it there; production's 443 block hands the path to the website
    # container, which 404s it.
    location ^~ /.well-known/acme-challenge/ {
        root /var/www/html;
    }
```

## Screenshots and video

Nothing visible changed: this is a host vhost file with no user-facing surface.

Linear: REB-250.


fiorelorenzo:
Independent review found three issues, all applied in dd5b7771:

1. **Blocker (priority 1):** the plain-HTTP location alone does not fix production. Live read-only probes showed certbot's installed 443 block redirects every plain-HTTP request, challenge path included, before any location is chosen, so `--webroot` renewal was still landing in the website container's 404. Fixed by rewriting the comment to say the 443 copy needs this location too, and by rewriting the PR body's "orchestrator pastes" section to say both halves, with the 443 block called out as the one that actually answers live traffic.
2. **Major (priority 2):** `preview.letsrebase.conf`'s corrected header sentence still missed one difference: the `/app` redirect targets a different CRM host (`preview.pigro.letsrebase.com` vs `pigro.letsrebase.com`). Added to the sentence.
3. **Nit (priority 3):** the file's own header list of paths that do not reach the container was missing the new `/.well-known/acme-challenge/` bullet. Added.

`pnpm --filter website test` still green (283 tests) after the fixes.

PR #148, opened 2026-09-16
fix(website): letsrebase.com/health probes the hub, not the CRM's stack

fiorelorenzo:
## What this changes

`letsrebase.com/health` and `preview.letsrebase.com/health` proxied to
`127.0.0.1:8080` and `127.0.0.1:8081`, PigroCRM's web containers per the port
table in `docs/adding-a-project.md`, not the hub's API (8084 production, 8086
preview). Every comment describing the route already disagreed with the code:
`letsrebase.conf`'s own header comment calls it "the API's probe" among the
hub's paths, `path-map-plugin.ts`'s `ELSEWHERE` entry called it "the hub API's
probe", and `deploy-hub.yml` already curls `127.0.0.1:8084/health` and
`127.0.0.1:8086/health` directly and never reads this route at all. Only the
README said the opposite: "`/health` stays on PigroCRM's stack."

Decided the way `/api/hub/` already is: the hub answers for its own API
surface on this name, and the CRM has its own name (`pigro.letsrebase.com`)
for its own probe. Both vhosts now proxy `/health` to the hub API's port, with
a one-line comment placed with the rest of the hub's locations. `rebase_api`'s
`GET /health` (`projects/hub/apps/api/src/rebase_api/main.py:56-62`) already
touches the database (`SELECT 1`) before answering, so this is not a
configuration-only probe. `path-map-plugin.ts`'s `ELSEWHERE` string and the
README sentence now agree with the vhosts and with `deploy-hub.yml`.

## How I verified it

`pnpm --filter website test`: 13 files, 283 tests, all passing (the `route()`
suite covers the `ELSEWHERE` map, including `/health`). `pnpm --filter website
lint`: clean. Read `projects/hub/apps/api/src/rebase_api/main.py` to confirm
the hub API serves `GET /health` with a database check, and
`.github/workflows/deploy-hub.yml:95,111` to confirm the deploy already curls
8086/8084 directly (unaffected by this change, since it never went through
this vhost route). `nginx` itself is not installed on this runner, so
`nginx -t` could not be run; the edited blocks keep the exact shape of the
neighbouring `/api/hub/` block they now match.

`deploy-hub.yml`'s health checks are untouched: they already hit the hub API
directly on its container port and never went through this vhost path, so
nothing about deploy verification changes here.

`path-map-plugin.ts`'s `ELSEWHERE` entries for `/api` and `/hub` still read
"the Orbiters hub API"/"the Orbiters hub SPA", the pre-rebrand product name.
That is outside this card's scope (REB-254 named only the `/health` string at
`path-map-plugin.ts:47` at audit time); I left it alone and am reporting it to
the orchestrator as a defect found outside scope rather than fixing it here.

## What the orchestrator pastes into the installed copies

Both installed vhosts (`/etc/nginx/sites-available/letsrebase.conf` and
`preview.letsrebase.conf`) have certbot's port-443 block on top and are edited
in place. Replace the existing `location = /health { ... }` block in each with:

Production (`letsrebase.conf`):

```
    # The hub API's probe (REB-254): the same host as /api/hub/ above.
    location = /health {
        proxy_pass http://127.0.0.1:8084/health;
        include /etc/nginx/snippets/orbiters-proxy.conf;
    }
```

Preview (`preview.letsrebase.conf`):

```
    # The hub API's probe (REB-254): the same host as /api/hub/ above.
    location = /health {
        proxy_pass http://127.0.0.1:8086/health;
        include /etc/nginx/snippets/orbiters-proxy.conf;
    }
```

`nginx -t` before reloading either.

## Screenshots and video

Nothing visible changed: `/health` answers `{"status": "ok"}` either way: it
now also proves the hub's database rather than the CRM's.

Linear: REB-254.


fiorelorenzo:
Independent review found one issue, fixed in adacff0b:

**Major:** projects/website/deploy/nginx.conf's own header comment (the container's own nginx, separate from the host vhost) still said "/health...belongs to PigroCRM's API and stays on its stack" and pointed at a path (deploy/nginx/letsrebase.conf) that was never this vhost's real location, a fifth record disagreeing with the four this card set out to align. Fixed to name the rebase hub and the correct path.

pnpm --filter website test still green (283 tests) after the fix.

PR #149, opened 2026-09-16
fix(web): killApi no longer throws on a pid a concurrent e2e run already reaped

fiorelorenzo:
## What this changes

`e2e/resilience.spec.ts` kills the real API mid-suite and relaunches it itself, tracking
the pid through `PIGROCRM_E2E_API_PIDFILE` — one fixed path (`/tmp/pigrocrm-e2e-api.pid`
by default), not a per-checkout handle. `projects/pigrocrm/AGENTS.md` already names the
consequence: a second `pigrocrm-e2e` run alive on this shared devbox can replace or reap
that pid before `helpers.ts`'s `killApi()` gets to it, and `process.kill` then throws
`ESRCH` — observed live on 2026-09-10, filed as REB-91.

`playwright.config.ts` runs this project with `fullyParallel: false` and `workers: 1`, so
I confirmed the two tests in `resilience.spec.ts` do not race each other within a single
run: the hazard is specifically a second, concurrent `pigrocrm-e2e` invocation on the
same box, exactly as `AGENTS.md` already diagnoses ("the pidfile trap does not show up
any other way"). There is no in-file coupling to remove, so I did not restructure the two
tests.

`killApi()` now catches `ESRCH` and treats it as "already dead" — every other `kill`
error still throws — then falls through to the same `waitUntil(!pingApi())` it already
ran on the success path, so the test that follows sees the API genuinely gone (the port
stops answering) rather than an unhandled exception in its place. `relaunchApi()` now
writes the new pid to a process-unique temporary file and `renameSync()`s it into place,
so a concurrent reader of the fixed pidfile path always sees one whole pid or the other,
never a partial write from a `writeFileSync` caught mid-flight.

## How I verified it

- `bash apps/web/scripts/e2e.sh e2e/resilience.spec.ts`, twice in a row on this devbox
  (checked no other `pigrocrm-e2e` container or process was live before each run): both
  passed, 2/2 tests, 1.1m and 1.3m. Teardown left no stray container, pidfile or uvicorn
  process either time (the second attempt's first try hit a transient
  `address already in use` on :55434 from the previous run's just-removed container
  reusing the port before the OS released it; a retry a few seconds later succeeded
  cleanly, and that collision is a pre-existing, out-of-scope port issue
  `projects/pigrocrm/AGENTS.md` already names — the same "hardcoded, not per-checkout"
  ports problem this PR does not touch).
- Not reproduced here: an actual second, concurrent `pigrocrm-e2e` invocation racing this
  one for the pidfile (I do not have two checkouts free to run at once without violating
  the project's own "one checkout at a time" convention). The fix is verified by reading
  the exact failure `AGENTS.md` and the card describe (`process.kill` throwing `ESRCH`
  when the pid is already gone) and confirming `killApi()` no longer propagates that
  specific error, not by reproducing the two-run race live.

The card's conditional "if the file's two tests share state, make them independent" —
I concluded they do not race (`workers: 1`, `fullyParallel: false`) and left them as they
are, chaining pid tracking from one test's relaunch into the next test's kill on purpose,
which is the documented, intended behaviour ("every spec file after this one, regardless
of run order, still has a working backend"). If a reviewer disagrees with that reading of
the config, that is the one place this PR could be incomplete.

## Screenshots and video

Nothing a person sees or does: this is a test-helper robustness fix with no product-facing
surface.

Linear: REB-91.


fiorelorenzo:
Independent review (overall: correct, no blocking defect) found the fix only covered half the race. Applied in commit 26397f51:

1. **ENOENT on a missing pidfile, not only ESRCH on a reaped pid.** A concurrent run's own teardown ends in `rm -f` on the pidfile, so `readFileSync` could throw outside the try/catch that only handled `kill`'s ESRCH, dying on an unhandled exception exactly as before. Fixed: a missing pidfile now reads as `pid = 0`.
2. **`pid > 0` guard against `kill(0, …)`.** `Number('')` is `0`, and `kill(0, SIGTERM)` is not an error — it broadcasts the signal to the caller's whole process group (the Playwright worker and the shell running it), which the new try/catch would not have caught since it succeeds rather than throwing. Fixed with one comparison.
3. **Leftover temp pidfile on a failed `renameSync`.** Now removed in a `finally`.
4. **Stale `projects/pigrocrm/AGENTS.md` paragraph** still described the pre-fix ESRCH-kills-the-run behavior and cited a since-moved line number. Updated to the current behavior, and added the one case this fix does not cover (a concurrent run that *replaces* the pid with its own live API, rather than reaping it).

Left as informational only, per reviewer: the pre-existing `if (child.pid)` behavior when `spawn` fails to start `uv` is unchanged (an honest timeout, not a silent failure), and the docstring's 'exactly as dead either way' phrasing was softened rather than treated as a separate defect.

Re-verified: `pnpm --filter web lint` and `tsc --noEmit` clean, `apps/web/scripts/e2e.sh e2e/resilience.spec.ts` twice more with the final code (2/2 passed, 30.7s and 27.9s), no leftover pidfile or temp file after either run.

PR #152, opened 2026-09-16
fix(infra): the CRM's Postgres data directory has no default, like the hub's

fiorelorenzo:
## What this changes

`docker-compose.yml` kept `${PIGROCRM_DATA_DIR:-../../data/postgres}` for the
`db` service's volume, a relative default that `docs/adding-a-project.md` §7
forbids: an `.env` that forgets the variable brings Postgres up on an empty
relative directory, the deploy's health check answers 200, and the real
register is never mounted. The hub's own compose file already uses
`${REBASE_DATA_DIR:?...}` for exactly that reason. The CRM is the project
that holds the fiscal register, and it was the one disagreeing with its own
runbook.

Now `PIGROCRM_DATA_DIR` uses the hub's own `:?` sentence: `set it in .env, to
this stack's own directory outside the repository`. Both live `.env` files
(production and preview) already carry the key, confirmed by the
orchestrator before this card was filed, so this changes nothing about how
either environment starts up. What changes is a deploy whose `.env` forgets
the variable: it now refuses to start instead of coming up quietly wrong.

Compose interpolates the whole file before a build, even though a build
never mounts the volume, so `PIGROCRM_DATA_DIR` also had to gain a throwaway
value in the two places that build the images without a real `.env`:
`.github/workflows/ci.yml`'s `pigrocrm-images` job and
`.github/preflight.json`'s `pigrocrm-images` entry. Both now pass
`PIGROCRM_DATA_DIR=./build-only-never-mounted`, the same value and shape the
hub's own image jobs already use (it has to stay path-shaped, or compose
reads the word as a named volume and refuses the project).

## How I verified it

- `docker compose --env-file <throwaway, with PIGROCRM_DATA_DIR set> config -q`
  passes.
- With `PIGROCRM_DATA_DIR` unset, `docker compose config -q` fails with
  `error while interpolating services.db.volumes.[]: required variable
  PIGROCRM_DATA_DIR is missing a value: set it in .env, to this stack's own
  directory outside the repository`, the clear sentence the card asks for.
- `python3 -c "import json; json.load(open('.github/preflight.json'))"` and
  `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"`
  both parse.
- `docker compose build` with the same env vars `ci.yml`/`preflight.json` now
  pass (`POSTGRES_PASSWORD=build-only PIGROCRM_JWT_SECRET=... PIGROCRM_DATA_DIR=./build-only-never-mounted`)
  built `pigrocrm-api`, `pigrocrm-mcp` and `pigrocrm-web`, and left no stray
  `build-only-never-mounted` directory on disk (a build never mounts the
  volume).
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_compose_environment.py`:
  86 passed (this card touches no `Settings` field, so the guard test is
  unaffected; run to confirm the compose file still parses the way that
  suite expects).
- `preflight --list` against this diff selected only `pigrocrm-images`
  (`docker-compose.yml` is the only file in its `when` globs among what
  changed); ran it, see above.

Both live `.env` files carrying `PIGROCRM_DATA_DIR` is the orchestrator's
own statement from before this card was filed, not something I could verify
myself from this worktree (no SSH access to either host). I did not
independently re-check the servers.

## Screenshots and video

Nothing visible changed: this is a compose file and two CI configuration
files, no UI, no API surface.

Linear: REB-258.


fiorelorenzo:
Independent review (agent, read-only, diff origin/main...HEAD): overall correct, confidence 0.85, no blockers. Confirmed the `:?` sentence matches the hub's, the throwaway value matches the hub's own build jobs, and this is the only place in the repo that builds the CRM's images without a real `.env` (the deploy workflow always passes `--env-file`, `test_compose_environment.py` parses only the anchor and service blocks so the `db` volume is out of its scope, and `docs/adding-a-project.md` §7 already prescribes this exact pattern).

Six should-fix/nit findings, all applied in `61c77238`:
1. The api/mcp documents-volume comment cross-referenced the database comment's `../../` explanation, which this PR deleted along with the relative default it justified. Made it self-contained.
2. The CI job's comment named only `PIGROCRM_JWT_SECRET` and `PIGROCRM_DATA_DIR` as `:?`, but `POSTGRES_PASSWORD` is too; the count was wrong and reads as if it were prunable. Named all three.
3. `.gitignore` and `.dockerignore` still explained their `data/` entries as "when PIGROCRM_DATA_DIR is left at its default", which no longer exists. Reworded.
4. `.env.example` never told an operator the stack now refuses to start without the key, unlike the identical `POSTGRES_PASSWORD` case two lines above it. Added the same "Required" language.
5. The compose comment asserted the present contents of two files (both live `.env`s) this repository cannot see or keep honest. Dropped; the invariant stands without it.
6. `backup-db.sh`'s `DEST` computation silently fell back to a directory inside the checkout when `PIGROCRM_DATA_DIR` was unset, the same silent-wrong failure mode this card removes from compose. Switched to `:?`; along the way found the original Italian message's apostrophe broke `bash -n` (though not `sh -n`) inside the nested substitution, and reworded it. Verified both parse and the DEST logic fails loudly / resolves correctly in isolation.

Re-verified after the fixes: `docker compose config` passes with the full env and fails with the clear sentence without `PIGROCRM_DATA_DIR`; `uv run pytest -q .../test_compose_environment.py`: 86 passed; `docker compose build` with the CI throwaway env built all three images; `.github/preflight.json` and `.github/workflows/ci.yml` still parse.

PR #154, opened 2026-09-16
feat(ci): the hub deploy fails when the mcp container does not start

fiorelorenzo:
## What this changes

The hub deploy verified one URL, the API's `/health` (8084 production, 8086
preview), then printed `docker compose ps` without inspecting it and exited
0. The stack has had a third service since REB-213: `rebase_mcp.http` serves
a `/health` of its own on `REBASE_MCP_PORT`
(`apps/mcp/src/rebase_mcp/http.py:45,148-150`), and nothing curled it. A
crashed `mcp` container, or a port that never got published, left
`/api/hub/mcp` answering 502 while the release reported success.

`_deploy-compose.yml` gains an optional `extra-health-url` input, empty by
default, and its "Check the deployed code answers" step now:

- curls `health-url`, and `extra-health-url` too when a caller sets it, in
  the same retry loop, requiring both to answer 200;
- unconditionally asserts every non-one-off service `docker compose ps -a
  --format json` reports is `State: "running"`, whatever either URL says, by
  piping the output through `jq -s` and failing (with logs) on anything not
  running. `-a` is required here: without it an exited container is absent
  from the output entirely rather than reported as not running, which an
  independent review caught (see below) and which this PR verified live
  against a throwaway stack.

`deploy-hub.yml` passes the `mcp` service's own `/health` as
`extra-health-url`: `http://127.0.0.1:8089/health` for preview,
`http://127.0.0.1:8088/health` for production, the ports
`projects/hub/AGENTS.md` records for `REBASE_MCP_PORT`.
`docs/adding-a-project.md` now mentions the new optional input where a new
project copies the deploy pattern from an existing caller.

## How I verified it

`python3 -c "import yaml, sys; yaml.safe_load(open(...))"` on both edited
files: both parse. Read `projects/hub/apps/mcp/src/rebase_mcp/http.py:44-50,
147-150` to confirm the `mcp` service answers a plain `{"status": "ok"}` on
`/health` (no database touch, unlike the API's). Extracted the embedded
health-check shell script and ran `shellcheck -s bash` on it: clean except
`SC2034` on the unused `attempt` loop variable, an idiomatic retry-loop
counter that predates this change.

**Proved live against two disposable compose stacks** (not this worktree,
torn down after): a one-service stack answering 200 exits the step with 0; a
two-service stack where the primary answers 200 but the second container
exits immediately (`sh -c "exit 1"`) correctly fails with `::error::Service(s)
not running for <project>: mcp: exited`, which is the exact scenario REB-246
describes. Also confirmed `docker compose ps -a --format json` surfaces a
`docker compose run` leftover with `Labels` containing
`com.docker.compose.oneoff=True`, which the jq filter now excludes so a
legitimate one-off container is never counted as a crashed service.

`_deploy-compose.yml` is a `workflow_run` workflow: per `AGENTS.md`, it only
ever executes in the version already on the default branch, so the real
mcp-probe path against the actual hub stack cannot be exercised from a
branch. It is proven on the trunk's next real preview deploy of the hub after
this merges, and on the next `hub-v*` tag for production. `website` and
`pigrocrm` never pass `extra-health-url`, so `EXTRA_HEALTH_URL` is empty for
them and `extra_code` stays hardcoded `200`, keeping their retry condition
exactly `[ "$code" = 200 ]`, today's behaviour; every service in both
projects' compose files is a long-running `restart: unless-stopped` process
with no one-shot container, so the new assertion introduces no new failure
mode for either.

The "every service running" assertion is unconditional for every caller, not
gated behind `extra-health-url`: deliberate, since a crashed container should
fail any project's deploy, not only the hub's, and it is a no-op for a stack
that is actually healthy.

`deploy-website.yml` and `deploy-pigrocrm.yml` are untouched, and neither
passes the new input, so their behaviour is unchanged.

## Screenshots and video

Nothing visible: this is a CI/deploy workflow file, no user-facing surface.
The mcp-crash failure path itself can only be proven on the trunk's next
deploy, but the assertion logic that detects it was proven live, as above.

Linear: REB-246.


fiorelorenzo:
Independent review found real problems, both blockers fixed in 9b799b16, verified against a live throwaway stack:

1. **Blocker:** `docker compose ps` without `-a` omits exited containers entirely (confirmed against a disposable stack: a service that ran and exited produced zero JSON lines), so a container that started, died and stayed dead did not appear in the assertion at all and the step still exited 0, the exact false success REB-246 exists to remove. Fixed by adding `-a` and excluding `docker compose run` leftovers via their `com.docker.compose.oneoff=True` label, which `-a` also surfaces and which is not part of the stack's own service set.
2. **Blocker:** the ps/jq pipeline ran unguarded inside a heredoc without `set -e`, so its own failure (e.g. `jq` missing on the deploy host, the first host-side dependency on it in these workflows) would silently report success. Both stages now exit 1 loudly with a named error if they fail.
3. **Nit:** the timeout message for a caller that never sets `extra-health-url` printed an empty `(or )` and a spurious `/ 200`. Now branches on whether the second URL exists.
4. **Nit:** `docs/adding-a-project.md`'s "four things" sentence, the one a new project copies from, did not mention the new optional fifth input. Added.

I re-verified the fixed assertion end to end against two throwaway compose stacks (not the shared worktree): a healthy one-service stack exits 0, and a two-service stack with a crashed extra container (primary health URL still answering 200) correctly exits 1 with `Service(s) not running for <project>: mcp: exited`.

PR #155, opened 2026-09-16
docs(hub): correct the MCP spec and the stdio docstring on the transport that shipped

fiorelorenzo:
## What this changes

The MCP design spec (`docs/superpowers/specs/2026-09-15-mcp-for-admins-design.md`)
and `__main__.py`'s docstring both describe a transport that is not the one
shipped: mounted inside the hub API at `/api/hub/mcp`, no separate compose
service. What actually runs is the opposite: `apps/api`'s `ruff.toml` bans
importing `apps/mcp` ("apps/api must not import from apps/mcp"), so
`rebase_mcp.http:app` ships as its own `mcp` compose service on its own port
(`REBASE_MCP_PORT`), and the host vhost proxies `/api/hub/mcp` there
directly. `projects/hub/AGENTS.md` already records the real shape.
`projects/hub/AGENTS.md` sends the next agent to this spec before changing
the shape of anything, so a spec describing a forbidden architecture is a
trap for whoever reads it first.

Per the spec's own convention ("Never rewrite a row" is `DECISIONS.md`'s
rule; this file is a dated design record with the same spirit, and
`AGENTS.md`'s English rule keeps every record it once wrote), the original
sentences stay untouched and a dated correction paragraph follows each of
the three affected passages: the Transport paragraph in section 1, the
`main.py` bullet in section 3.2, and the "a separate `mcp` service" line in
section 5's out-of-scope list.

`__main__.py`'s docstring is code documentation, not a decision record, so it
is fixed directly rather than annotated.

## How I verified it

`uv run ruff check projects/hub`: all checks passed. Read
`projects/hub/apps/api/ruff.toml` to confirm the import ban
("apps/api must not import from apps/mcp"), `projects/hub/docker-compose.yml`
to confirm the `mcp` service runs `rebase_mcp.http:app` as its own container,
and `projects/website/deploy/letsrebase.conf:53-56` to confirm the vhost
proxies `/api/hub/mcp` straight to that container's port rather than through
the API.

Whether a "Correction, <date>" paragraph is the right shape for this file
versus editing the original sentences outright: I followed
`projects/hub/AGENTS.md`'s instruction to read this spec before changing
shape and the repository's general practice of not rewriting dated records
(`docs/design/DECISIONS.md`'s own rule, and the same reasoning `AGENTS.md`
applies to grandfathered specs under `docs/superpowers/`), but this is a
project spec rather than `DECISIONS.md` itself, and a case could be made for
editing it outright since nobody currently treats hub specs as an immutable
audit trail the way `DECISIONS.md` is.

## Screenshots and video

Nothing visible changed: a design document and a module docstring, no
user-facing surface.

Linear: REB-245.


fiorelorenzo:
Independent review found two non-blocking wording nits, both fixed in 6e7974a8:

1. The 3.2 correction said "main.py mounts nothing," which contradicts the router list it sits directly under (main.py mounts plenty: signups, freelancers, companies, admin, members, pigro, tokens, /health). Narrowed to say it mounts no MCP transport and enters no lifespan for it.
2. The section 5 correction said "that last line of this list is superseded," but that line carries three unrelated items (token expiry, a PigroCRM-opening token, and a separate mcp service); only the last one shipped. Narrowed to name that one item.

Everything else the review checked out: no original sentence was rewritten (pure insertions), every factual claim matches the shipped code (ruff.toml's import ban, docker-compose.yml's mcp service, main.py mounting no MCP transport, the vhost proxying straight to the mcp container), and markdown formatting is clean.

uv run ruff check projects/hub still passes.

PR #157, opened 2026-09-16
feat(pigrocrm): every Monday a space receives its week by mail, and can switch it off

slavni96:
## What this changes

After the welcome mail a space never heard from PigroCRM again, and PostHog showed the spaces going quiet after their first day. Now every Monday at 08:00 (Europe/Rome) each space receives «La tua settimana», a mail with the week just closed: the invoices to collect (overdue, worst first, and those due in the next seven days), the ones to issue (won deals without an invoice, unbilled hours and their accrued value), the invoices issued and the ones collected in the week, the hours, the pipeline (open deals by stage, the deals that moved, the offers still waiting) and the operational signals. A section prints only when it has rows; a week in which nothing moved closes with «Settimana ferma» and a ready-made prompt for the assistant; every link carries `?da=digest` so PostHog can tell a click from the mail apart. The subject names the week's non-zero facts (`La tua settimana: 2 fatture emesse, 1.800,00 € da incassare`).

I built it as one object per week (`WeeklyDigest`, from `DigestService.build`, which copies every figure from the repository or service that owns it and adds no arithmetic of its own), one renderer (`digest_mail`), one run (`DigestRun.send_for_space`: active users with the switch on, one send per space and ISO week recorded in a new `digests` table, an activity `digest.inviato` with counts only, then a server-side PostHog event `digest_inviato` per recipient) and one command, `pigrocrm digest [--slug] [--data YYYY-MM-DD] [--forza] [--dry-run]`, that walks the tenant registry the way `ensure-space-defaults` does and prints one line per space. It is a cron job, not a worker, for the same reason the Gmail sync is one: PigroCRM has no daemon (the row is in `docs/design/DECISIONS.md`, 2026-09-16).

A person switches it off from a new settings tab, **Profilo** (`/app/impostazioni/profilo`), which every user can reach from the user menu; the mail's last line links there. This is a deviation from the plan, which put the switch on the users panel: that panel is admin-only at the service layer, so a collaborator who received the mail could never have opted out.

**Adjacent:** no open card touches the mail seam, the CLI or the settings tabs. REB-222 (the Home in three blocks), REB-223 (customers proposed from Gmail) and REB-224 (invoice PDF handed to the assistant) are the sibling cards from the same spec and are untouched here.

## How I verified it

Each of the nine plan tasks was implemented by a subagent, reviewed against the brief and the spec by a second one, and fixed where the review found something; then a whole-branch review and one more fix round. From the worktree root:

- `uv run pytest -q` on the core files the branch touches (`test_digest_models`, `test_invoice_aggregates`, `test_digest_build`, `test_mail`, `test_telemetry`, `test_digest_run`, `test_cli_digest`, `test_auth_users`, `test_migrations`, `test_architecture`, `test_clock`, `test_dashboard_no_arithmetic`): 226 passed together with the API auth file below (one `-p no:xdist` run, 50 s); the last commit added one more case in `test_mail.py`, run with its four neighbours: 100 passed. `test_cli_digest.py` provisions real spaces in the Postgres container, sends through a `RecordingSender` and reads the `digests` rows back; `test_digest_run.py` covers the five outcomes, a partial and a total refusal by the sender, `--forza`, `--dry-run`, soft-deleted-only spaces, and asserts no transaction is left open after any outcome.
- `projects/pigrocrm/apps/api/tests/test_auth_api.py` in the same run, including a real `collaboratore` PATCHing `/api/auth/me`.
- `uv run ruff check projects/pigrocrm && uv run ruff format projects/pigrocrm && uv run mypy`: clean.
- `pnpm --filter web test`: 1183 passed in 112 files; `pnpm --filter web lint` and `pnpm --filter web build`: clean.
- The full core suite (about two hours on this machine) was not run locally; CI runs it.
- The mail rendered from a realistic `WeeklyDigest` is in the Screenshots section; I read the HTML and the text bodies side by side.

## Migrations

`0035_digest_settimanale`: adds `users.digest_settimanale` (boolean, `server_default true`, not null) and creates `digests` (`id`, `settimana` varchar(8) unique, `inviato_a` jsonb, `occurred_at` timestamptz). Safe on live databases: a new column with a server default and a new table. `test_migrations.py`'s `compare_metadata` agrees. Spaces receive it at the first API boot after the tag, through `ensure-space-defaults`; `pigrocrm digest` itself never migrates a space, one behind the schema answers `saltato` and waits for the next boot.

## API changes

- `PATCH /api/auth/me`, body `MeUpdate { digest_settimanale: bool }`, response `UserRead`; 401 anonymous. Self-only by construction: there is no id in the path or the body.
- `UserRead.digest_settimanale: bool`, `UserUpdate.digest_settimanale: bool | null`.
- `projects/pigrocrm/apps/web/src/lib/api-types.ts` regenerated with `openapi-typescript` against the running API (port 8000 was held by another worktree's API on this machine, so it ran on a free port; the diff is exactly the new route and fields).

- The MCP surface: no tool sends or reads the digest, and none flips the switch. `DigestService.build` and `UserService.update_own_digest` are named exclusions in `apps/mcp/tests/test_mcp_surface_coverage.py`: every number of the report is already a tool or a resource, and the switch is a person's own preference on their inbox.
- `SollecitiService.candidates` still drives the reminder screen with its grace and chase-count filters; the mail lists every unpaid invoice past its due date instead, because a report that hides a chased invoice is a report that lies.
- The sidebar's settings group stays admin-only; a non-admin reaches Profilo from the user menu.

- The order inside `DigestRun.send_for_space`: reads, `rollback()` to close the read transaction so the dashboard's snapshot opens as the first statement, `build`, the sends, then the `digests` row, the activity and one `commit()`, then PostHog. The mails go out inside the snapshot transaction, so a space's connection sits idle-in-transaction for one Resend round-trip per recipient: fine at today's sizes, worth knowing before a space has fifty users.
- Two crons that wake together both pass the «già inviato» check and both send; the loser's insert then hits the unique constraint and answers `saltato`. The cron line runs once per host, and the comment in `run.py` says so.
- Without a Resend key the run answers `saltato (invio_non_configurato)` and records nothing, so a Monday with a missing key is retried, not burned. Production has the key.
- The production host is on UTC: the operator's crontab line is `0 6 * * 1` during CEST (`0 7` after the October change), or `timedatectl set-timezone Europe/Rome` and the README's `0 8`. The runbook says so.

## Screenshots

Three pictures and one video, attached below (1440×900 for the app, same data on both sides, the before from a worktree on `origin/main` against the branch's API).

1. `pair-1-profilo-tab.png`: Impostazioni before (opens on Campi, no profile anywhere) and after, with the new **Profilo** tab first and the switch «Ricevi il resoconto settimanale».
2. `pair-2-user-menu.png`: the user menu of a `collaboratore` before (only «Esci») and after, with «Profilo».
3. `mail-after.png`: the weekly mail rendered from a realistic week (invented data). There is no before: the mail did not exist on `main`.
4. `video-profilo.mp4`: the collaborator opens the menu, reaches Profilo, switches the report off (toast) and back on.

Nothing else is visible: the cron, the command, the migration and the PostHog event have no screen.

Linear: REB-221.

pair-1-profilo-tab

pair-2-user-menu

mail-after

https://github.com/user-attachments/assets/f29d51f3-e226-4cbb-8b2b-3658b31f45aa


slavni96:
## Review record

The branch was reviewed before this PR opened, task by task and then as a whole, by independent read-only reviewers; the whole-branch pass found nothing Critical, two Important and six Minor findings. What I did with each:

- **Important — «Da incassare» dropped invoices due inside the reported week** (`scadute` came from `SollecitiService.candidates`, which applies a 7-day grace and the chase-count limits). Fixed in 3a665e1f: `InvoiceRepository.list_scadute(oggi)` lists every unpaid issued invoice due on or before today, worst first; three tests (mid-week due date, due today, chased to the ceiling).
- **Important — the state word on «Emesse» was always «emessa».** Fixed in 3a665e1f: `incassata` / `trasmessa` (from `trasmessa_esternamente_il`) / the register state, with tests for each and for precedence.
- **Judgement call — no Resend key recorded the week as sent.** Changed in 48475245: `saltato (invio_non_configurato)`, nothing written, so the week is retried once the key is there. Now answered before the report is built (39db7873).
- Minor: runbook said «Impostazioni → Utenti», now «Profilo» (48475245); `api-types.ts` regenerated after the last route docstring (5e135ee7); the hours link pointed at a `fatturato` filter the page does not declare, now `/app/ore` (3a665e1f); `sum_incassate_in_periodo` had no caller and is gone (3a665e1f); a debt due today reads «scade oggi» rather than «0 giorni di ritardo» (39db7873).
- **Left as is:** the mails go out inside the snapshot transaction (one idle-in-transaction connection per Resend round-trip per recipient), harmless at today's sizes and noted in the body; two crons awake together could both send once, the loser answering `saltato`, documented in `run.py`.

The scoped re-review of the fix round confirmed every finding addressed (68 tests on the three files it ran).

PR #158, opened 2026-09-16
fix(core): a malformed PEC or issuer email is refused before a register number is spent

fiorelorenzo:
## What this changes

`PECDestinatario` (the customer's PEC) and `ContattiTrasmittente/Email` (the issuer's
email) were checked only for width and Latin-1 character range before a register
number was spent. The vendored FPR12 1.2.3 XSD types the first as `EmailType`, a
strict `local@domain` pattern, and the second as `EmailContattiType`, which additionally
sets a `minLength` of 7. Neither pattern nor floor was applied, so a PEC pasted whole
from a mail client (`Studio Rossi <pec@studiorossi.it>`, with the display name and
angle brackets still attached) or a placeholder email under 7 characters passed
`check_party_exportable`, `issue` consumed the year's next number, and every
`export_xml` after refused the file forever, with annulment as the only remedy.

Now `_check_latin` and the two `_dati_trasmissione` writer call sites for these two
fields carry `_EMAIL_RE` (mirroring `EmailType`'s own `xs:pattern`) and, for the
issuer's email only, `_EMAIL_MIN = 7` (mirroring `EmailContattiType`'s `minLength`).
Both are applied only when the field already carries a value: an absent PEC or issuer
email is something the writer omits, not something it refuses, and the pre-check must
not be stricter than the writer in the other direction either.

## How I verified it

- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_fatturapa.py`:
  81 passed, including three new cases — a mail-client-pasted PEC refused by both the
  pre-check and the writer with the same details, a well-formed PEC accepted and
  written unchanged, and an issuer email under 7 characters refused by both.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_invoice_issue.py`: 39
  passed, including a new service-level case that issues a draft for a customer whose
  PEC is `Studio Rossi <pec@studiorossi.it>`, asserts `ValidationFailed` with
  `field == "pec"`, and asserts the year's counter row still reads 0 afterward (the
  register number was never spent).
- Confirmed all four new tests fail without the fix: `git stash` on the source file
  alone, re-ran the same four tests, three failed with "DID NOT RAISE
  ValidationFailed" (the fourth, the well-formed PEC, passes either way by
  construction), then restored the fix and re-ran to green.
- `uv run mypy`: `Success: no issues found in 307 source files`.
- `preflight --list` against this diff selects `pigrocrm-py-lint`, `pigrocrm-py-types`
  and `pigrocrm-py-test` (the full non-slow, non-planner suite). I ran the lint and
  type checks above and the two touched test files directly rather than the full
  project suite or the 50k-row corpus, per this session's instruction not to run
  those locally; `pigrocrm-py-test` is what CI's `ci` job runs.
- CRM web checks (`pnpm --filter web lint/test/build`, `tsc --noEmit`) were not run:
  this change touches only `packages/core` Python, no `apps/web` file.

`CustomerCreate.pec` and `EmitterProfileUpsert.email` still accept any string at write
time (`customers/schemas.py`'s own comment says as much: "nothing in this codebase
validates a customer's email format today"). I deliberately left that alone: the card
asks for the issue-time gate inside `check_party_exportable`, which now refuses before
a register number is spent regardless of when the bad value was saved, and that is
where FatturaPA specifically requires the shape. Adding a write-time `EmailStr`-style
check on those two fields would change what a plain `PATCH /customers/{id}` or
`PATCH /emitter-profile` accepts today, is not needed to close the number-burning gap,
and is a product decision (does a freelancer get an error while typing, or only when
they try to issue) that belongs to Ivan rather than to this fix.

The exporter also writes the issuer's email a third time, into `CedentePrestatore`'s
own `Contatti/Email` (same `EmailContattiType` element, `_cedente`, not
`_dati_trasmissione`). I did not add `pattern=`/`min_length=` to that call site: it
reads the same `emittente.email` value that `check_party_exportable` has already
validated before either writer method runs, so there is no path where a number gets
spent for a value this call site would refuse and the pre-check would not. Adding the
same two keyword arguments there for symmetry is a one-line, no-behaviour-change
follow-up if a reviewer would rather every `_text` call naming an email carry its own
belt as well as the pre-check's braces.

`_EMAIL_RE` is applied to both `PECDestinatario` (`EmailType`) and
`ContattiTrasmittente/Email` (`EmailContattiType`) rather than giving each its own
pattern: `EmailContattiType`'s own pattern (`.+@.+[.]+.+`) is looser than `EmailType`'s
and adds nothing a value matching `EmailType`'s pattern would not already satisfy, so
one regex for both is stricter than the schema requires for the issuer's email but
never laxer, which is the direction that matters for not spending a number on a value
the SdI will reject.

## Screenshots and video

No UI changed: this is a validation rule inside `packages/core`'s FatturaPA generator,
exercised only from the service layer and its tests.

Linear: REB-227.


fiorelorenzo:
Independent review found three real correctness gaps, all fixed in 3aae5539:

1. **Blocker.** `EmailContattiType` is not implied by `EmailType`: its own `.+@.+[.]+.+` demands a literal dot after the `@`, which `EmailType`'s pattern does not require. An issuer email like `info@localhost` passed the pre-check on `_EMAIL_RE` alone and still produced a document invalid against `EmailContattiType`, exactly the failure this PR exists to close. Fixed with `_EMAIL_CONTATTI_RE`, the intersection of both patterns, applied to both elements the schema types `EmailContattiType` (`ContattiTrasmittente/Email` and `CedentePrestatore/Contatti/Email`); `PECDestinatario` stays on `_EMAIL_RE`, its real type.
2. **Blocker.** The PEC pattern applied even when the writer would never emit `PECDestinatario` (a customer routed by an SDI code keeps `pec` as an unchecked contact field). Reproduced: a customer with both an SDI code and a loosely formatted PEC, which issued and exported fine before this PR, was refused after it for a value the document never carries. Fixed by mirroring the writer's own `not codice_sdi` condition in the pre-check.
3. **Blocker.** Both checks ran on the unstripped value, but `EmailType`'s `xs:token` base means the SdI whitespace-collapses the value before the pattern applies. A padded address (a trailing space from a paste) was refused by the pre-check though the document itself would have tolerated it. Fixed by stripping once and using that value everywhere: the pre-check, the writer, and the emitted text.
4. **Minor, applied.** The pattern refusal's `expected` reached the web client as the raw ~150-character regex. Added an optional `pattern_expected` (defaulting to the regex, so every other call site is unchanged) and pass `"un indirizzo email nella forma nome@dominio"` at the two email sites.
5. **Nit, declined.** Refactor the `_check_latin` row tuple into a named type for readability. I actually did adopt this (as `_Campo`, a `NamedTuple`) since growing the row to 7 fields for finding 4 made the positional tuple genuinely hard to read, so this landed as a side effect of fixing (1) and (4) rather than as a separate change.

Four new regression tests added, each verified to fail against the pre-review-fix commit (801b0a99) and pass now.

PR #159, opened 2026-09-16
fix(web): fattura_emessa fires when the invoice is issued, not the draft

fiorelorenzo:
## What this changes

`fattura_emessa` was recorded on `POST /api/invoices`, which is `InvoiceService.create`:
it writes `stato="bozza"` and assigns no number. Every draft and every abandoned
proforma counted as an issued invoice in PostHog, and a genuinely issued one, through
`POST /api/invoices/{invoice_id}/issue`, never counted at all.

The funnel table now keys `fattura_emessa` on the issue route itself. A row segment
written `{name}` matches any single path segment, the same shape `api-types.ts` uses
for a resource id, so the invoice's own id is never enumerated. Updated the design
doc's row (`docs/design/2026-09-12-posthog-analytics-design.md:110`) to match.

## How I verified it

- `pnpm --filter web test` (the full CRM web suite, `vitest run`): 113 files, 1187
  tests, all passing, including the new `analytics.test.ts` case that posts a draft
  invoice and asserts no `capture` call, then posts to the issue route and asserts
  exactly one `capture('fattura_emessa')`.
- Skipped `pigrocrm-e2e` (Playwright against a local API/Postgres stack): outside what
  this change touches functionally, and heavy to stand up on a shared box already
  running several other agents' test suites.

This card comes from the 2026-09-16 code audit of `pigrocrm-web`. The audit's own
recommendation was either to extend `tablePath` to match a path-parameter template or
to normalise the id segment; I chose the first, turning `EVENTS` from an exact-match
map into a small ordered list matched segment by segment, since a second parameterised
route (the sibling card REB-234 does not need one, but a future one might) would
otherwise need its own bespoke regex. A product owner might instead want to count
`spazio_creato`-style "first attempt" data for invoices too (a draft started but never
issued), which this PR deliberately does not add: the card asks only to fix what
`fattura_emessa` means, not to add a second event for the abandoned case.

## Screenshots and video

Nothing a person sees or does changed: this is a server-response classifier inside a
`Middleware`, invisible in the UI. No pair, no video.

Linear: REB-232.


fiorelorenzo:
Independent review found four findings, all applied in 7ee33fe8:

1. **Empty placeholder match (real bug).** `rowMatches`'s `{name}` segment matched an
   empty request segment too, so `POST /api/invoices//issue` (an id that had not
   resolved yet) would have earned `fattura_emessa`. Fixed: a placeholder now requires
   a non-empty segment.
2. **Row order now load-bearing.** Switching `EVENTS` from a `Map` to an ordered list
   means the first matching row wins; a literal row added below the `{invoice_id}`
   wildcard would be silently shadowed. Documented in `rowMatches`'s own comment.
3. **Redundant path split.** `eventFor` now splits the request path once and passes
   the array to every row instead of re-splitting per row.
4. **Missing boundary test.** Added a case for `/api/invoices/{id}/artifacts` and
   `/confirm`, the two sibling POSTs a single invoice emission also makes, so a future
   lax trailing segment would fail a test instead of silently double-counting
   `fattura_emessa`. Added the empty-segment case from finding 1 too.

Nothing left as is; all four were real, cheap to fix, and now covered by a test.

PR #160, opened 2026-09-16
fix(web): the session user type comes from the generated UserRead schema

fiorelorenzo:
## What this changes

`auth.tsx` declared its own `SessionUser` interface by hand and cast the session
query's data onto it with `data as SessionUser | null`, even though the same
endpoint (`GET /api/auth/me`) already answers `components['schemas']['UserRead']`
in the generated `api-types.ts`. The `as` switched off the one guard that is
supposed to make a backend contract change break `tsc` here instead of showing up
as a runtime surprise: every other feature module in the app reads its type
straight from `components['schemas'][...]`, and this was the one place that did
not.

`SessionUser` is now `export type SessionUser = components['schemas']['UserRead']`
and the cast is gone; `data` already carries that type through `unwrap`.

## How I verified it

- `pnpm --filter web build && pnpm --filter web exec tsc --noEmit`: green. The
  schema turned out to carry three fields the hand-written type never had
  (`tariffa_oraria_default`, `costo_orario_default`, `created_at`), so `tsc`
  failed exactly where the card predicted it might: on the two tests that build
  a full `AuthValue` mock through `vi.mocked(useAuth).mockReturnValue(...)`
  (`ConnectAgentDialog.test.tsx`, `TokensPanel.test.tsx`). Fixed by adding the
  three fields to both mocks rather than keeping the cast.
- `pnpm exec vitest run src/lib src/features/tokens src/features/get-started`
  (the touched dirs, plus `get-started` since it also mocks a session user):
  15 files, 166 tests, all passing.
- Skipped `pigrocrm-e2e` (which `preflight --list` selects for this diff): the
  devbox has several other agents' PigroCRM stacks up right now, and
  `projects/pigrocrm/AGENTS.md` documents that two concurrent e2e runs corrupt
  each other through the fixed pidfile and ports, not just contend for them.
  Nothing in this change touches e2e-observable behaviour (a type-only fix plus
  two test-mock updates), so I judged the risk of colliding with a sibling not
  worth it; happy to run it in isolation if asked.
- Python tiers (`pigrocrm-py-*`) are all `skip` under `preflight --list`: this
  diff touches no Python path.

None from a product-decision angle: this closes an escape hatch in a contract
guard, it does not change behaviour. The one thing worth a second look is
mechanical: I filled the two tests' mock `user` objects with
`tariffa_oraria_default: null`, `costo_orario_default: null` and a
`created_at` timestamp that matches the convention already used for other
fixture dates in these files, since neither test asserts on those fields.

## Screenshots and video

Nothing visible changed: this is a type declaration and two test fixtures, no
component render or copy is different.

Linear: REB-240.


fiorelorenzo:
Independent review (fresh agent, read-only, worktree + `git diff origin/main...HEAD`): overall correct, no blockers. It independently re-verified the type change by restoring the two original mocks on a copy of the branch and confirmed they reproduce exactly the two TS2739 errors this PR fixes, and confirmed every `useAuth().user` consumer (AppShell, DayPanel, GetStartedPage, firstSteps, useCanWrite, useIsAdmin, roles.ts, analytics.ts) only reads the five original fields, so nothing leaks the three new ones.

Two P3 findings:
1. The PR body says auth.tsx 'was the one place' with a hand-shaped session-user cast; e2e/dashboard.spec.ts:100,194 and e2e/economics.spec.ts:174 still cast `await res.json() as { id: string }`, and features/settings/UsersPanel.test.tsx:17-19 builds an AuthValue via `as never`. Both are pre-existing, outside this card's scope (auth.tsx:8-14,66), and the reviewer itself calls the `as never` one the house idiom for vitest mocks. Left as is; the PR body's claim is narrower than 'the one place' but the fix itself is complete for the file this card names.
2. ConnectAgentDialog.test.tsx's added `created_at` used TokensPanel.test.tsx's fixture date instead of this file's own (2026-09-12T10:00:00Z everywhere else). Fixed in 10f1d931 -- zero functional change, nothing asserts on the field.

PR #161, opened 2026-09-16
fix(website): a campaign landing on /community keeps its attribution into the hub

fiorelorenzo:
## What this changes

`/community` has a door into the hub too (`Raccontacelo`, `<a href="/hub/aziende">`),
but `carryUtm` lived only in `landing.js`, which only `index.html` and `pigrocrm.html`
load. A campaign landing on `/community?utm_source=linkedin&...` reached that one click
into the hub with no attribution: `community.js` never wrote
`sessionStorage['orbiters.utm']`, the key the hub's `resolveUtm`
(`projects/hub/apps/web/src/lib/utm.ts:25,64`) falls back to when its own URL has none.

`carryUtm`, `readUtm` and `pageSlug` move out of `landing.js` into a new shared
`utm.js`: an IIFE publishing `window.__utm`, the same pattern `field.js` and
`typewriter.js` already use for the shared field and the title's typewriter.
`community.js` gains one guarded call, `window.__utm.carryUtm()`, in its own `start()`.
`landing.js` keeps its own `reveal()` and its own mounts, calling the same shared
function instead of the copy it used to own.

Loading the whole of `landing.js` on `community.html` was not an option: its `start()`
also mounts the shared field and the shared typewriter on the same `#field` canvas and
the same `h1 .role` `community.js` already mounts for this page, so the file would
double-bind both (two independent animation loops on one canvas, two competing typing
timers on one element). `utm.js` carries only what both pages need.

`community.js` keeps no browser state of its own: the write lives in `utm.js`, so the
existing `community.test.ts` assertion that this file's own source never mentions
`sessionStorage` still holds, unchanged.

`utm.js` loads first on all three pages, ahead of every script that calls it.

## How I verified it

- `pnpm --filter website test`: 292 passed. `landing.test.ts` (all of it about
  `carryUtm`) moved to `utm.test.ts`, which also pins that `landing.js` and
  `community.js` both call `window.__utm.carryUtm()` and that all three pages load
  `utm.js` before the script that calls it.
- `pnpm --filter website build && pnpm --filter website exec tsc --noEmit`: clean.
- `pnpm --filter website test:e2e`: 59 passed, including a new test that `/community`
  carries the UTM keys and `da=community` onto its one door into the hub, the same
  shape as the existing `/` and `/pigrocrm` checks.
- Read the production build's own chunk graph directly, because `utm.js` has no
  literal chunk of its own in `dist/`: Vite merges a script with an entry point's
  shared-across-the-same-pages set into one file, so `utm.js` lands inside the chunk
  it shares with `consent.js` and `field.js` (all three load on all three pages),
  which is guaranteed to execute, via a real ES `import`, before `landing.js` or
  `community.js` runs. The e2e suite is what actually proves the order is correct
  end to end, against the built preview server rather than the source.
- `WEBSITE_WEB_PORT=127.0.0.1:8099 docker compose build` in `projects/website`: the
  production image builds with the new script and serves it.

The generated `dist/index.html` has only three `<script type="module">` tags, not
four: `utm.js` never gets its own file, because Rollup groups modules by the exact set
of pages that load them, and `consent.js`, `utm.js` and `field.js` are loaded by the
same three pages. I read the merged chunk's contents to confirm the execution order
(`window.__consent`, `window.__utm`, `window.__pigroField` are all set before the next
script tag runs) and the e2e suite exercises it against the real preview server, but
this is worth a second look given how surprising the missing file looked at first.

## Screenshots and video

Nothing a person can see: every door still reads and looks the same, and the six
UTM keys and `da=` land on a query string nobody who is not opening dev tools reads.
No pair, no video.

Linear: REB-247.


fiorelorenzo:
**Independent review** (reviewer subagent, read-only, PR diff + utm.js/utm.test.ts/community.js/community.test.ts/e2e as context; verdict: correct, confidence 0.87, no blocking findings):

Applied (commit c443d20b):
1. privacy.html's origin-page sentence named only home and PigroCRM's page; community is a third now, added.
2. AGENTS.md and README.md's script count moved to six.
3. utm.js's own header miscounted its two callers as three; fixed.
4. utm.test.ts's byte budget left the file only 111 bytes of headroom (97.3% of ceiling); raised one step.
5. Dropped community.test.ts's duplicate of the utm.js wiring check utm.test.ts already makes more precisely, for all three pages.

Left for a separate change: the hub's own comments (utm.ts, api.ts, models.py, schemas.py, service.py) still enumerate only home/pigrocrm as origins, now incomplete since community is reachable. Comment-only, no behaviour change, and it touches projects/hub rather than this card's projects/website scope.

Also noted: the PR body overstated the ordering guarantee as one mechanism on all three pages; it is document order on index.html/pigrocrm.html and a real ES import on community.html, both spec-guaranteed. Correcting for the record here rather than re-editing the body.

PR #162, opened 2026-09-16
fix(web): give the revealed Token creato field an accessible name

fiorelorenzo:
## What this changes

The one-time reveal dialog («Token creato») rendered its raw-secret field as
`<Input readOnly value={issued?.token ?? ''} />` with no `id`, no `<Label
htmlFor>` and no `aria-label`. The dialog's own title and description name the
dialog, not the field, and the sibling copy button next to it does have a
label (`aria-label="Copia il token"`). The same field in `ConnectAgentDialog`
already carries `<Label htmlFor="mcp-token">Token</Label>` — the two surfaces
that are deliberately kept in step (`useUnsavedTokenGuard` is shared between
them precisely so they cannot drift) had already diverged here.

The field now has `id="token-issued"` and a `<Label htmlFor="token-issued">
Token</Label>` above it, the same shape `ConnectAgentDialog` uses.

## How I verified it

- `pnpm exec vitest run src/features/tokens`: 3 files, 37 tests, all passing,
  including a new one (`gives the revealed token field an accessible name`)
  that asserts `screen.getByLabelText('Token')` resolves to the field and
  holds the freshly issued value — it fails against the pre-fix markup
  (`getByLabelText` throws when nothing labels the input) and passes with the
  fix.
- `pnpm --filter web build && pnpm --filter web exec tsc --noEmit`: green.
- Skipped `pigrocrm-e2e` for the same reason as REB-240 (other agents'
  PigroCRM stacks are up on this shared devbox right now, and the project's
  own docs warn two concurrent e2e runs corrupt each other through the fixed
  pidfile and ports); this diff changes no e2e-observable flow either.

None. This is a one-line accessibility fix with a test that pins the actual
regression (no accessible name), not just the markup.

## Screenshots and video

**1. The revealed-token dialog, with a label now naming the field**

Token creato dialog, before and after

Captured on the CRM's own dev stack (a fresh local Postgres, `uv run
uvicorn`, this worktree on 5173 and a second `origin/main` worktree on
5175, both proxying the same local API) at 1440x900: create a token, and
look at the revealed field. No video: nothing here answers to a click that
did not already exist, the field's behaviour is unchanged and only its
accessible name is new — a screen reader difference a still frame cannot
show either, which is why the pair is annotated instead.

Linear: REB-239.


fiorelorenzo:
Independent review (fresh agent, read-only): correct, no findings. It confirmed the Label/id pairing produces the accessible name live (getByRole('textbox', { name: 'Token' }) resolves, exactly one #token-issued node), confirmed the sibling 'Token' label on ConnectAgentDialog's mcp-token cannot collide since that dialog cannot be open at the same time as this non-dismissable one, and confirmed the new test genuinely pins the regression: substituting origin/main's TokensPanel.tsx makes exactly that test fail (getByLabelText throws), and it passes with the fix, stable over three repeats. tsc --noEmit clean at this commit. No changes applied.

PR #163, opened 2026-09-16
fix(web): documento_creato fires on both document-creation routes

fiorelorenzo:
## What this changes

The funnel table named only `POST /api/documents`, which the shipped UI uses
exclusively for the upload flow (`useCreateDocument`, called from
`DocumentsTab.tsx`, to mint the row before the multipart upload). An offer made
through the «Nuovo da template» dialog goes to `POST /api/documents/from-template`,
exactly the path Get started's own step 4 tells a new space to take («Dal deal,
"Crea documento": il template "Offerta" è già pronto»), and it earned no event at
all.

Both routes now earn `documento_creato`. Updated the design doc's row
(`docs/design/2026-09-12-posthog-analytics-design.md:108`) to name both.

## How I verified it

- `pnpm --filter web test`: 114 files, 1200 tests passing, including the new case
  that posts to both routes and asserts each earns exactly one
  `documento_creato`, and the "leaves out" test still proves nothing else nearby
  (`/api/invoices/import`, the invoice's `/artifacts` and `/confirm` siblings) is
  mistaken for it.
- Skipped `pigrocrm-e2e` (Playwright against a local API/Postgres stack): outside
  what this change touches functionally, and heavy to stand up on a shared box
  already running several other agents' test suites.

The card explicitly asked me to decide whether the upload path should keep sharing
`documento_creato` with the template path, rather than getting its own name. I kept
them sharing it: the funnel's stated meaning (`lib/analytics.ts`'s own docstring,
and the design doc) is "a document exists now" for the activation funnel, and that
is equally true of either route. A product owner might instead want to tell the two
apart, since the from-template path is the one the product itself points new users
at and the upload path might be a different signal (an existing customer, migrating
records). Splitting them into two names is a two-line change if that's wanted; I did
not make it because the card's own wording asked for "the same `documento_creato`
name" and nothing in the design doc or the onboarding spec distinguishes the two
today.

## Screenshots and video

Nothing a person sees or does changed: this is a server-response classifier inside a
`Middleware`, invisible in the UI. No pair, no video.

Linear: REB-234.


fiorelorenzo:
Independent review found four findings, all applied in 5036b85c:

1. **Missing negative pin (P2, real regression risk).** The patch necessarily removed
   `/api/documents/from-template` from the "leaves out everything the table does not
   name" test (it now IS an event), which left the documents family as the only pair
   sharing a prefix with no negative test at all: a future lax match would have
   silently double-counted `documento_creato` on `/restore` or `/versions` with a
   green suite. Restored, mirroring the invoice siblings REB-232 already pins.
2. **Redundant test (P3).** The dedicated "counts a document made from a template
   the same as..." test asserted nothing the parameterised `it.each` row added in the
   same patch didn't already cover, since the middleware keeps no state between
   calls. Removed.
3. **Design doc row shape (P3).** The doc folded two code rows into one cell with
   "or"; split back into two rows, one call per row like every other, since the
   section calls this table "the contract".
4. **Docstring precision (P3).** "The funnel's meaning is «a document exists now»"
   is wider than the table (a restore or version regeneration also satisfies it and
   is deliberately excluded); reworded to "a new document was created".

Nothing left as is; all four were real and cheap to fix.

PR #164, opened 2026-09-16
fix(core): a failed provisioning keeps its registry row until the database is actually dropped

fiorelorenzo:
## What this changes

`TenantService._undo` deleted the registry row for a failed provisioning in a `finally`, regardless of whether `drop_database` actually succeeded. `DROP DATABASE` raises `ObjectInUse` when a connection appears between its own `pg_terminate_backend` and the drop, which a concurrent request for the same slug can cause. So a failed drop still freed the slug onto a database that survived: already migrated, already holding the failed attempt's admin user and seeded defaults. A later signup for the same slug would find `create_database_if_missing` answer `False`, `migrate_to_head` a no-op, and inherit a stranger's admin account.

Now `_undo` deletes the registry row only once `drop_database` returns successfully. A failed drop is logged (`tenant.slug`, `tenant.db_name`) for an operator to clear by hand, and the exception propagates instead of being swallowed: the row stays, the slug stays taken, and the space stays unreachable rather than handed to somebody else.

## How I verified it

- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_tenants.py`: 50 passed, including two new cases against a fake `drop_database` monkeypatched into `pigrocrm.core.tenants.service`: one that raises `psycopg.errors.ObjectInUse`, asserting the exception propagates and the registry row survives; one that succeeds, asserting the row is gone.
- Confirmed both new tests fail without the fix: stashed the source change alone, reran the two tests. The failed-drop case failed with `assert 0 == 1` (the row was gone despite the raised exception, the exact bug), the successful-drop case still passed (unchanged behavior). Restored the fix and reran to green.
- `uv run mypy`: `Success: no issues found in 313 source files`.
- `preflight --list` selects `pigrocrm-py-lint`, `pigrocrm-py-types` and `pigrocrm-py-test` for this diff; ran the first two directly and the one touched test file rather than the full suite or corpus, per this session's instruction.
- CRM web checks were not run: this change touches only `packages/core` Python.

`provision`'s `except ValidationFailed:` arm and its `except Exception:` arm have identical bodies (`self._undo(tenant, url); raise`), so the first is redundant since `Exception` already covers `ValidationFailed`. This is a real finding but on lines I did not touch (my change is inside `_undo` itself), so I left it alone rather than fold it into this fix; noted in the commit body too.

`_undo` is called from two `except` arms in `provision`, both of which then `raise` to re-raise the *original* failure (the one that triggered the undo, e.g. a domain refusal from the seed step). If `drop_database` itself now raises inside `_undo`, that new exception propagates immediately and the original one is lost except as `__context__` on the traceback (standard Python exception chaining) rather than being the exception that ultimately surfaces to `provision`'s caller. This is the correct trade-off per the card (the drop failure is the more actionable, more urgent problem an operator needs to see and the row must survive), but it does mean a caller catching the *original* exception type around `provision()` would instead see whatever `drop_database` raised in this one, rare case. Not something I think needs its own guard, but worth a second look.

## Screenshots and video

No UI changed: this is a service-layer ordering fix in `packages/core`, exercised only from tests against the registry database.

Linear: REB-230.


fiorelorenzo:
Independent review found no correctness blockers (overall_correctness: correct). Three minor doc/logging findings, all applied in ce74ad41:

1. `provision`'s own docstring still said "Any failure after the row undoes the row" unconditionally, which is exactly the guarantee this PR makes conditional. Reworded to point at `_undo`.
2. `_undo`'s docstring claimed the surviving row makes the space "unreachable", which is backwards: a surviving registry row is what `SpaceRegistry` routes a slug on, so a half-provisioned space (database migrated, admin created) stays reachable, not the opposite. Reworded to just "the slug stays taken" until an operator clears it.
3. The failed-drop log used `logger.error` with only the slug and db name, not the cause; an operator deciding whether to retry or investigate the server needs the traceback. Switched to `logger.exception`.

Also fixed a lint failure ruff caught on the first CI run (unsorted imports in the new test), unrelated to the review.

PR #165, opened 2026-09-16
fix(web): stop the connect-agent dialog asking twice before the Token page

fiorelorenzo:
## What this changes

The «Gestisci i token» link inside `ConnectAgentDialog` was guarded twice for
the same navigation. Its own `onClick` calls `close(false)`, which asks
«uscire comunque?» and, on yes, discards the just-issued token and closes the
dialog. The link's own navigation is then evaluated by the router's blocker
(`useUnsavedTokenGuard`, shared with the Token page), which asks the identical
question again — and by the time it does, `close()` has already discarded the
token, so answering «no» the second time cancels the navigation but does not
bring the token back, the opposite of what the dialog's own comment promised.

The link now passes `ignoreBlocker`, so this one hand-driven navigation goes
through `close()`'s own prompt only. Every other way to leave with a token on
screen — Back, a reload, closing the tab — still goes through the blocker
alone, unchanged.

## How I verified it

- `pnpm exec vitest run src/features/tokens`: 3 files, 36 tests, all passing.
  The mocked `Link`/`useBlocker` in `ConnectAgentDialog.test.tsx` now simulate
  the router's own blocker check closely enough to observe the double prompt
  at all — before this PR the test suite mocked `useBlocker` as a pure no-op,
  which is exactly why the bug went unnoticed. The rewritten test asserts
  `window.confirm` is called exactly once for this navigation and that the
  token stays on screen when the answer is «no».
- `pnpm --filter web build && pnpm --filter web exec tsc --noEmit`: green.
- Skipped `pigrocrm-e2e` for the same reason as REB-240/REB-239 (several other
  agents' PigroCRM stacks are up on this shared devbox right now, and the
  project's own docs warn two concurrent e2e runs corrupt each other through
  the fixed pidfile and ports).

The alternative the card offered — dropping the `onClick` confirm entirely
and letting the blocker be the only question, resetting state in its own
accept path — is not what I did. `shouldBlockFn` is a synchronous
predicate with no accept-path callback in this router version, and this
component owns no navigation lifecycle hook to hang a reset on; `ignoreBlocker`
keeps the existing, already-tested `close()` codepath as the single source of
truth for what happens when this link is followed, and only removes the
redundant second ask. If a future version of the guard exposes a real
accept-path hook, unifying around the blocker instead would remove the
`ignoreBlocker` special case.

## Screenshots and video

Nothing visible changed in markup: this reduces a duplicated confirm() to one
call so the interaction goes through the resolved dialog->Link->router chain,
with no reachable-in-time reproduction against last-question because
confirm() is a native browser dialog Playwright would have to intercept, and
the actual bug is a *count*, not a pixel difference a still frame or clip
could show. The rewritten test in this PR is the reproduction: it fails
against the pre-fix component (double `confirm` call, and the second "no"
discards a token that visibly stayed on screen after the first prompt) and
passes with the fix.

Linear: REB-238.


fiorelorenzo:
Independent review (fresh agent, read-only): the production fix is correct -- it verified ignoreBlocker is scoped to this single Link's own history.push (router-core's tryNavigation), so Back/popstate and beforeunload still go through the shared blocker single-prompted, unchanged. But it found the regression test did not actually pin the bug: the mock's Link returns at event.defaultPrevented before consulting the blocker, so the *decline* path (confirm answers no) passes with or without the fix -- it never reaches the second, stale shouldBlockFn call. Fixed in 22d84786 with an *accept*-path test (confirm answers yes): verified directly by removing ignoreBlocker and watching this new test fail (confirm called twice), then restoring it and watching it pass.

PR #166, opened 2026-09-16
fix(web): a failed first-steps read no longer burns the Get started redirect

fiorelorenzo:
## What this changes

`useFirstSteps`'s own `value()` maps an errored read to `true` on purpose, for the
panel it feeds: a request that broke should not make the Get started page nag about
an unfinished step. `useFirstVisitGoesToGetStarted` read that same collapsed value to
decide whether the person had nothing left to do, so a single 500, 503 or dropped
connection on the very first Home render made every one of the six reads look done,
called `markGetStartedSeen`, and burned the one-time redirect for that browser for
good: the flag is never cleared, so the person is marked as having seen a page they
were never shown.

`useFirstSteps` now also reports `failed`, computed from each query's own `isError`,
alongside `loading`. The redirect checks it before marking the visit seen, and leaves
the browser undecided instead when a read failed, so the next visit tries the reads
again. The panel's own optimistic `value()` is untouched: a step that could not be
read still shows as done there, for the reason the file's own comment already gives.

## How I verified it

- `pnpm --filter web test`: 113 files, 1188 tests passing, including the new case
  that makes every `api.GET` call reject and asserts neither `navigate` nor the
  `localStorage` flag fires.
- Skipped `pigrocrm-e2e` (Playwright against a local API/Postgres stack): the failure
  path is exercised at the unit level above; standing up the stack to reproduce a
  broken API round-trip through Playwright would add nothing this test doesn't
  already prove.

I left the panel's own `done` values as they are (an errored step still shows as
"done" on the Get started page itself), since that is a documented, separate decision
in the same file and this card is only about the one-time redirect. A product owner
might prefer the panel to show a step as unreadable rather than done when the read
failed, which would be a different, larger change to `FirstStep`'s shape; I did not
make it because nothing in the card or the design record asks for it and the existing
comment states the reasoning deliberately.

## Screenshots and video

Nothing a person sees changed: the Get started panel renders identically either way,
since the fix only changes when the browser marks itself as having already been
redirected. No pair, no video.

Linear: REB-235.


fiorelorenzo:
Independent review found two findings, both applied in e80872a8:

1. **Test anchored on a sleep, not the error state (P2).** The failed-read test's only
   discriminating assertion (the localStorage check) was equally true while the six
   reads were still pending, since `navigate` is never expected in this scenario
   either way. Fixed by having `renderHome` return its `QueryClient` and waiting for
   all six cached queries to actually reach `status: 'error'` before asserting.
2. **`failed` not gated on `enabled` (P3).** `loading` already is; without the same
   guard, a disabled `useFirstSteps({ enabled: false })` could report a stale
   `failed: true` from an error cached under a query key it shares with
   `GetStartedPage`'s own call. No observable impact today (the only `enabled: false`
   caller already short-circuits before reading it), but a one-token fix keeps the
   next caller from inheriting the surprise.

Nothing left as is.

PR #168, opened 2026-09-16
fix(website): pitch.css draws its colours and mark from shared/brand

fiorelorenzo:
## What this changes

`pitch.css` used to be the one stylesheet the palette plugin never touched: it
restated six values out of `shared/brand/palette.css` as its own custom
properties (`--ink`, `--quiet`, `--paper`, `--gold`, `--melon`,
`--melon-strong`) and declared its own `@font-face` for Outfit. The values
matched today, but a colour changed in the shared palette would have silently
left the deck, the one shared by link with people outside the project, on the
old value.

`pitch.css` is now a `TOKEN_CONSUMERS` entry, like `landing.css` and
`community.css`. What is true now:

- Its six variables are `var()`s of the injected `--color-*` custom
  properties, not a second copy of the palette.
- Its own `@font-face` is gone; `palette-plugin.ts` prepends
  `@rebase/brand/font.css` the same way it already does for the other two
  sheets, and `body`'s `font-family` reads the injected `--font-sans`.
- `landing-tokens.test.ts` has a sibling `describe` block for the deck: it
  resolves every one of the six variables to the shared palette's live hex,
  carries no raw hex in them other than white, and no longer declares a
  `@font-face` of its own.

The four-tile mark keeps its own drawing (`.glyph`, `.dark .glyph`) rather
than reusing `system.css`'s: the deck runs at a fixed 1920x1080 and the tile
is 14px here against the landing's 6px, with its own margin and box-shadow
offsets, plus a `.dark` variant `system.css` has no need for (it is never
shown on a dark background). The colours those rules use are the same four
tokens, now injected rather than restated, so the fork this issue is about is
gone; the second drawing itself is a deliberate, documented exception, not an
oversight.

## How I verified it

- `pnpm --filter website test`: 294 passed, including two new
  `pitch deck tokens` cases and the extended token count for
  `landing-tokens.test.ts` (13 tests in that file).
- `pnpm --filter website build`: `dist/assets/pitch-*.css` inspected directly —
  exactly one `@font-face`, and `--ink:var(--color-prussian-blue)` resolving to
  `--color-prussian-blue:#011936` in the built output.
- `/pitch` rendered from `vite preview` on this branch and from a second
  worktree on `origin/main`, at 1440x900: the cover slide and one dark slide
  are pixel-identical apart from the typewriter's currently cycling word
  (`Developer` vs `CTO`, which cycles on its own timer and is not something
  this change touches).

Whether the `.glyph` duplication is worth documenting as a
`docs/design/DECISIONS.md` row rather than just a code comment. I left it as
a comment next to the rule (`pitch.css`, right above `.glyph`) since it is an
implementation detail of one page rather than a site-wide rule, but it is a
call that could go the other way.

## Screenshots and video

No interaction changed, so there is no video. Two before/after pairs at
1440x900, before from a second worktree on `origin/main`, after from this
branch:

**1. The cover slide**

Pitch deck cover slide, before and after

**2. A dark slide ("L'obiettivo")**

Pitch deck dark slide, before and after

Linear: REB-248.


fiorelorenzo:
Independent review (agent) findings and what I did with them:

1. **--grid still spelled Prussian Blue by hand** (pitch.css line 15). Fixed: now `var(--system-grid)`, the same equivalent landing.css/community.css already consume.
2. **pitch.html's canvas field hard-coded the same four brand hexes field.js reads from tokens.** Fixed: now reads `--ink`/`--gold`/`--quiet`/`--melon` off `:root` via `getComputedStyle`.
3. **Stale "two stylesheets" comments** in palette-plugin.ts, system.css and README.md. Fixed: all three now name pitch.css as a third consumer; README's mark bullet also now says pitch.css's own larger `.glyph` drawing is not asserted against `@rebase/brand/mark` the way system.css's and BrandMark.tsx's are.
4. **PITCH_DECLARATION's comment claimed alternation order mattered for correctness.** It does not (the regex backtracks into the group on a partial match); reworded to describe what the pattern actually reaches.
5. **resolveLandingColour/resolvePitchColour were byte-identical duplicates.** Collapsed into one `resolveColour(token, css)`.
6. **Hold pitch.css's 14px .glyph to the same mark order landing-style.test.ts asserts for system.css's**, or parameterise the tile length in system.css so the deck overrides one length rather than the whole rule. Left as is: real gap, but it is either a new test surface (asserting a second, differently-scaled drawing of the mark) or a small system.css redesign (a `--system-glyph-tile` custom property), and both are bigger than this issue's diff. Filed as a defect below for a follow-up card rather than folded in here.
7. **TOKEN_CONSUMERS gives no way for pitch.css to opt out of system.css's prefers-contrast/.consent rules it doesn't use** (the deck's field canvas would hide under `prefers-contrast: more` with no compensating change, and ~2KB of unused `.consent`/`.sr-only` CSS rides along). Left as is: correct observation, but redesigning TOKEN_CONSUMERS' all-or-nothing injection is out of scope for a fork-removal issue and neither regression is covered by an existing test (nothing currently claims pitch.css supports `prefers-contrast: more`).

Second commit: 77b9a3cd.

PR #169, opened 2026-09-16
fix(web): a failed availability probe no longer dead-ends the signup wizard

fiorelorenzo:
## What this changes

`GET /api/tenants/{slug}/disponibile` swallowed a thrown fetch and any non-2xx answer other than 429 into `setAvailability({ state: 'idle' })`. With `idle`, `current` is `null`, so `problem` is `null`, «Crea lo spazio» stays disabled and the hint keeps rendering the same "still checking" ellipsis it shows mid-flight, forever, since the effect's own deps (`screen`, `slug`, `localProblem`) never change on their own.

`Availability` gets a fourth state, `failed`, carrying the API's own sentence (`toProblem(...).detail`, or the generic Italian fallback for a thrown fetch). The hint renders it in place of the ellipsis; the slug field is not marked `aria-invalid` for it, since a failed probe is not proof the slug itself has a problem, unlike a genuinely taken name. The submit button is no longer held hostage by a probe that never answered: `POST /api/tenants/` validates the slug again on the server, so `canCreate` lets a `failed` probe through the same way it already lets a `free` one.

## How I verified it

- `pnpm --filter web exec vitest run src/routes/app/registrati.test.tsx`: 11 passed, including two new cases for REB-236 (a rejected `GET` and a resolved 503).
- `pnpm --filter web build`: clean.
- `pnpm --filter web exec tsc --noEmit`: clean.

This card comes from today's audit (`docs/superpowers/` code audit of 2026-09-16, «Audit 2026-09-16: pigrocrm-web», finding 5). I chose "let the person submit anyway" over an explicit retry button, since the card names it first and the server already re-validates the slug on `POST /api/tenants/`; I did not add a second, redundant retry affordance next to it.

## Screenshots and video

Nothing new is rendered: the hint paragraph and the submit button already exist for the `idle`/`checking`/`free`/`taken` states (PR #141 set the same precedent for this exact paragraph, adding the 429 alert). Only the branch of `Availability` that feeds them changes, and the two new cases it triggers (a network failure, a 503) are not practical to record live.

Linear: REB-236.


fiorelorenzo:
Independent review, 4 findings, `overall_correctness: correct`, no blockers.

1. **Applied (3f944939).** The `failed` hint replaced the whole line with the API's reason, dropping the derived-address preview exactly when the button becomes enabled -- a person could submit an address they were never shown. The hint now appends the reason to the address (`${spaceHost()}/${slug}: ${reason}`) instead of replacing it, and both new tests assert on the address+reason pair.
2. **Applied (3f944939).** The comment above the 429 branch still said every failure drops to idle; it now says only a 429 does, matching the branch beneath it.
3. **Applied (3f944939).** Both new tests armed their failure with a one-shot mock (`mockRejectedValueOnce`/`mockResolvedValueOnce`), which assumes `user.type()` never pauses long enough between two keystrokes to let the 350ms debounce fire on a partial slug and consume the queued answer. Neither test wants a second, different answer, so both moved to a persistent mock.
4. **Left as is, filed as REB-265.** The effect's cleanup clears the timer but never marks an in-flight request stale, so a response for a superseded slug can still write its verdict and reach the same dead end from request ordering rather than a failure. This predates REB-236 and affects all four writers (`free`/`taken`/`failed`/the 429 idle branch), not only the two this card added, so it is scoped as its own card rather than widening this PR.

PR #170, opened 2026-09-16
docs(pigrocrm): .env.example says where compose reads .env, and documents the two missing variables

fiorelorenzo:
## What this changes

`.env.example` claimed twice that `docker compose` loads it from a copy "in the repository root, where docker-compose.yml also lives" -- but `docker-compose.yml` lives in `projects/pigrocrm/`, and compose reads `.env` from the directory of its own compose file, not the repository root. A local `docker compose up` run exactly as documented finds no `POSTGRES_PASSWORD` and refuses to start. On the server the deploy passes `--env-file "${DEPLOY_PATH}/.env"` explicitly, which is the only reason a root-level copy ever worked there.

Both the opening paragraph and the "From here down: only for docker compose up" section now say where each of the three consumers actually reads from: the Python `Settings` auto-load (repository root, this repository's own convention for `uv run`), local `docker compose` (`projects/pigrocrm/.env`, beside the compose file), and the server (`${DEPLOY_PATH}/.env`, via the deploy's explicit `--env-file`) -- the same shape `projects/hub/.env.example` already documents correctly.

`PIGROCRM_WEB_PORT` and `PIGROCRM_DOCUMENTS_DIR` are both interpolated by `docker-compose.yml` (the `web` service's port mapping and the `api`/`mcp` services' documents volume) but were never listed. Both are now documented, with the production ports from `docs/adding-a-project.md` §7's table, plus a commented preview block mirroring the hub example's own shape (its own password, its own two directories, its own two ports: web 8081, Postgres 55434).

## How I verified it

- `docker compose -f projects/pigrocrm/docker-compose.yml --env-file <throwaway env with every required and new variable set> config -q`: resolves cleanly.
- `uv run pytest -q projects/pigrocrm/packages/core/tests/test_compose_environment.py`: 86 passed (this file asserts against `docker-compose.yml` itself, not `.env.example`, and is unaffected by this change, but it is the closest guard this repository has and stayed green).

PR #152 (REB-258) merged into `main` before this branch was cut, adding `PIGROCRM_DATA_DIR`'s required-with-no-default comment; this PR is based on that commit and does not touch `PIGROCRM_DATA_DIR`'s own lines, only adds the two new variables after it. This card comes from today's audit (`docs/superpowers/` code audit of 2026-09-16, «Audit 2026-09-16: ci-infra», finding 9).

## Screenshots and video

Nothing visible changed: this is a comment file read by `docker compose` and by a person copying it to `.env`, not a product surface.

Linear: REB-260.


fiorelorenzo:
Independent review, 3 findings, `overall_correctness: correct`, no blockers.

1. **Applied (eb10d2a8).** `docs/adding-a-project.md` §7 requires writing the `.env` location in both `.env.example` and `AGENTS.md`; `projects/hub/AGENTS.md` already does, `projects/pigrocrm/AGENTS.md` said nothing at all. Added a `## Deploying` section stating the compose project names, all three `.env` locations, and the production/preview ports.
2. **Applied (eb10d2a8).** `docker-compose.yml`'s own comment on `PIGROCRM_COOKIE_SECURE` still called the local-dev `.env` and the one compose reads "this same file" -- the exact premise this card's `.env.example` fix corrects. Reworded to say they are separate copies; the substantive claim (compose never forwards this key to the `api` container) is untouched.
3. **Left as is.** `README.md`'s step 1 was also flagged, but its own step 2 already passes `--env-file ../../.env` explicitly on every documented command -- it never relies on compose's default lookup the way `.env.example` used to claim. That path is functionally identical to `_deploy-compose.yml`'s `--env-file "${DEPLOY_PATH}/.env"` (`DEPLOY_PATH` is the checkout root, two levels above the compose file, same as `../../` from `projects/pigrocrm/`), so the README's own procedure was never affected by the bug this card fixes.

PR #171, opened 2026-09-16
fix(website): a stored yes can be withdrawn from /privacy in one click

fiorelorenzo:
## What this changes

Accepting the cookie notice cost one click; taking back a stored `granted`
cost a trip into the browser's own site-data settings, which is not equally
easy (GDPR art. 7(3) asks for it to be). Nothing on the site cleared
`orbiters.consent`, and `/privacy`'s own text told the visitor to clear their
browser's data rather than offering a link.

What is true now:

- `consent.js` exposes `window.__consent.withdraw()`: it clears the stored
  key and reloads, so the next `start()` on a page that carries the notice
  shows it again.
- `/privacy` has a link, mid-sentence ("Se cambi idea, **ritira il
  consenso** e l'avviso torna") and in the footer next to Home and Termini,
  that does the same two steps. `/privacy` does not load `consent.js` itself
  (it carries no tracker to gate, `pixel.test.ts` holds that page carries no
  reference to the script at all), so its link reimplements the same two
  steps against the same key, with a comment tying the literal to
  `consent.js`'s `STORAGE_KEY` so a rename there has to carry over here too.
- The storage key itself, `orbiters.consent`, is unchanged.

## How I verified it

- `pnpm --filter website test`: 292 passed, including a new `consent.test.ts`
  case (`withdrawing consent`) that accepts, calls `withdraw()`, confirms the
  key is cleared and `location.reload` was called, then simulates the reload
  and confirms the notice shows again with neither tracker present.
- `pnpm --filter website build`: clean; Vite hashes the inline withdraw
  script into its own asset like it does the page's other module scripts.
- `pnpm --filter website test:e2e`: 60 passed, including a new end-to-end
  case that accepts on `/`, withdraws from `/privacy`, and confirms the
  notice reappears on `/`.
- Rendered `/privacy` from `vite preview` and clicked the link by hand: it
  clears the key and reloads.

`/privacy`'s withdraw link duplicates two lines of logic (and the literal
storage key) that already live in `consent.js`, rather than calling
`window.__consent.withdraw()` directly. That function only exists once
`consent.js` runs on the page, and `/privacy` cannot load it: doing so would
make the page reference the script `pixel.test.ts` forbids there (`carries
no pixel and no notice, because it has nothing to ask about`), and would
risk showing the full accept/deny notice, or auto-loading the trackers on a
stored `granted`, on the one page that is meant to carry neither. The
duplication is two lines behind a comment that names the key it has to stay
in lockstep with; the alternative (loading `consent.js` there and reworking
its auto-start behaviour to stay inert on this page) is a larger change than
this issue asks for.

## Screenshots and video

No new interaction beyond a link click, which is the whole feature, so
there is no separate video: the flow is exercised end to end by the e2e
test above. One before/after pair at 1440x900, before from a second
worktree on `origin/main`, after from this branch, of the paragraph that
changed:

**1. The withdrawal sentence on /privacy**

The withdrawal sentence on /privacy, before and after

Linear: REB-251.


fiorelorenzo:
Independent review (agent) findings and what I did with them:

1. **Withdrawing also erased a stored refusal, so a No got asked again.** Real regression against consent.js's own documented invariant ("A refusal is final until the visitor clears their own storage"). Fixed: both `withdraw()` and privacy.html's inline copy now clear the key only when it reads `'granted'`. Added a `consent.test.ts` case (`leaves a stored refusal alone: a no stays final`) pinning it.
2. **The duplicated storage-key literal in privacy.html was guarded by nothing CI runs.** Fixed: `pixel.test.ts` now reads `consent.js`'s `STORAGE_KEY` with a regex and asserts privacy.html's inline script both names it and gates on `'granted'` the same way — the same comparison-test idiom the file already uses for the PostHog literals.
3. **consent.js's `withdraw()` comment claimed a caller that doesn't exist.** Fixed: reworded to say no page calls it today and it's exported for the day one of the three notice pages grows its own control.
4. **landing.css's "two-link footers" comment was stale** (privacy.html's footer is three links now). Fixed.
5. **Give the withdraw click some visible result on /privacy** (it silently reloads a byte-identical page). Left as is: a real UX gap, but "confirm in place with aria-live" vs "redirect to /" is a product decision bigger than this issue, and no existing test claims the click is observable on this page.
6. **Withdrawal leaves the first-party tracker cookies (`__obref`, PostHog's) the replaced sentence used to clear.** Left as is: correct observation, but expiring specific third-party cookie names by hand is a separate, larger change (and a maintenance burden of its own) from clearing the consent flag itself; the sentence still names the browser-data route as the stronger remedy is a fair follow-up, not folded in here.
7. **No-JS fallback**: with JavaScript off the link is inert. Left as is: consent.js never runs without JS either, so nothing is loaded regardless; the practical harm is bounded to a `granted` written in an earlier session staying on record with no documented recourse, which is the same situation as today the sentence describes clearing browser data for.

Second commit: 877e17cd.

PR #172, opened 2026-09-16
fix(ci): the running-service check reads compose states without jq

fiorelorenzo:
## What this changes

The running-service assertion I added to `_deploy-compose.yml` this afternoon (PR #154, REB-246) used `jq` to read `docker compose ps --format json`, and the production host has no `jq`: the `pigrocrm-v0.17.0` deploy rolled the stack out, migrated every space, and then reported failure at that line (REB-264, run 35128649994). The check now asks compose for a Go template (`{{.Service}}\t{{.State}}\t{{.Labels}}`) and walks it with a `while read` loop, so a deploy depends on nothing beyond docker. Semantics unchanged: `-a` so an exited container is seen, one-off containers skipped by label, any non-running service fails the step with its name and state.

## How I verified it

`python3 -c 'import yaml; yaml.safe_load(...)'` parses the file. The loop body, copied verbatim, ran on the host over the live `rebase` stack (compose 2.40.3): the real list gives `not_running=[]`; the same list with an injected `web\texited\t...oneoff=False` row and an injected `db\texited\t...oneoff=True` row gives `not_running=[web: exited]`, so a dead service is reported and a `compose run` leftover is ignored. A `workflow_run`/tag workflow runs only from the default branch, so the step itself is proven by the next production tag; I cut `hub-v0.24.0` right after this merges and read its run.

`< <(printf ...)` is a bash process substitution; the step already runs `bash -s` on the host (line 222), and the heredoc it sits in is quoted (`<<'SSH'`), so nothing is expanded on the runner side.

## Screenshots

Nothing a person sees: a workflow step.

Linear: REB-246.


fiorelorenzo:
Independent review (one finding, applied in 174243fe): an empty `docker compose ps` list, which is what a wrong project name produces, passed the check on the health URL alone; it now fails with a sentence naming the project. Everything else checked out: quoting inside the quoted heredoc, `set -u` with empty `read` fields, the tab separator, the one-off filter, and `done < <(...)` keeping `not_running` in the shell. The reviewer ran the loop verbatim against live compose projects and a synthetic `api: exited, web: restarting` list.

PR #173, opened 2026-09-16
docs(repo): the board moves on PR events now, and the tracker table has the audit row

fiorelorenzo:
## What this changes

The team's pull request automation on Linear is on since this afternoon: a PR opening on a branch that carries the issue id moves the card to `In Progress` (a hand-set `In Review` is overwritten seconds later) and the merge moves it to `Done`. REB-247's history is the measurement: `In Review` at 17:30:40, `In Progress` at 17:30:43 when PR #161 was seen, `Done` at 17:46:42 as it merged, and every card merged since closed the same way. `AGENTS.md`, `docs/tracker.md` and the two skills that repeat it said the automation was off (PR #33, REB-80) and asked to be updated with the date once a merge closed its own issue. Now they say what the board does and what still needs a hand: the `In Progress` move before the branch exists, every comment (the closing evidence before or right after the merge, since `Done` will not wait), `Canceled`, and the `In production` comment when a tag ships. The tracker's project table also gains the row for `Code audit v1`, opened today.

## How I verified it

`get_issue REB-247` state history, quoted in the text. `list_issues` over the cards created today: every one whose PR merged after about 17:30 is `Done` with no hand on it. Nothing runs against these four files.

I did not switch the automation on and do not know who did, so the text records the observation with the date and points at `docs/design/DECISIONS.md` for the decision itself, and it leaves what `In Review` means to the setting's owner rather than redefining it.

## Screenshots

Nothing a person sees: Markdown.

Linear: REB-266.


fiorelorenzo:
Independent review, 8 findings, all applied in 31d7abaa: the AGENTS.md loop bullet and its `In Review` definition now agree with the paragraph below them; CONTRIBUTING.md joins the four files; "that day" reads 2026-09-10; the unverified case is a decision about the merge, not the state; the `**Merged:**` comment sits in the merge step of pr-creation; the DECISIONS pointer states what is known (the row is still to be written by whoever turned the setting on); the overwrite claim is narrowed to what REB-247 shows (an `In Review` set before the PR-open event fires).

PR #175, opened 2026-09-17
ci: check the wordmark SVGs against build-wordmark.py

fiorelorenzo:
## What this changes

I add a preflight check, `wordmark`, that reruns `shared/brand/tools/build-wordmark.py --check` whenever `shared/brand/**` changes, next to the `guide-pdf` entry it is modelled on. Nothing else ran this: the four SVGs (`wordmark.svg`, `wordmark-paper.svg`, `lockup.svg`, `lockup-paper.svg`) are generated and committed, so a hand edit or a palette change without a redraw survived every gate. I also add a `build:wordmark` script to `shared/brand/package.json` that runs the same script without `--check`, and point the README's "Changing the face..." section at the check by name instead of just at `--check` in the abstract.

One correction to the shape of the ask: the script's own shebang runs it as `uv run <script>`, not `uv run python <script>`. Its PEP 723 header declares `fonttools` and `uharfbuzz` as script-local dependencies, and `uv` only reads that header when the script itself is the run target; passed as an argument to `python` it resolves `python` inside the monorepo's shared workspace environment instead, which does not have `uharfbuzz`, and the check fails with `ModuleNotFoundError` on every invocation, unconditionally, not on drift. I use the working form (`uv run shared/brand/tools/build-wordmark.py --check`) in both the preflight entry and the new package script.

## How I verified it

Ran the check from the worktree root:

```
$ uv run shared/brand/tools/build-wordmark.py --check
4 files are what this script draws
```

No drift; the committed SVGs already match. Also ran `preflight --list` against the real diff of this PR (touching `shared/brand/README.md` and `shared/brand/package.json`) and confirmed the new `wordmark` check is selected:

```
run  wordmark                     uv run shared/brand/tools/build-wordmark.py --check
     matched: shared/brand/README.md, shared/brand/package.json
```

Smoke-tested the new script too: `pnpm run build:wordmark` from `shared/brand/` prints `4 files written, 0 of them changed`, and `git status` shows no SVG diff, confirming it targets the same four files with no side effect on a clean tree. Validated both edited JSON files parse (`node -e 'JSON.parse(...)'` on `.github/preflight.json` and `shared/brand/package.json`).

The invocation fix above: I changed the exact command from what the card specified (`uv run python ...`) to what actually runs (`uv run ...`), because the literal form throws on every run regardless of drift. Worth double-checking my read of `uv`'s PEP 723 handling if anything looks off.

## Screenshots and video

None: a CI manifest entry, a package script and a README paragraph, nothing rendered.

Linear: REB-249.


fiorelorenzo:
Reviewer findings and what I did with them:

1. **Must-fix, applied.** The README said the `wordmark` check "runs `--check` on every pull request touching `shared/brand/**`", but `.github/preflight.json` is the local, before-the-push tier (AGENTS.md, ci.yml); no workflow invokes `preflight`. Corrected the wording to say it runs locally, before the push.

2. **Nit, applied.** Added a clause to the check's `why` noting the two PEP 723 packages the script declares (fonttools, uharfbuzz) need uv's own wheel cache warm too, not only the tempdir font cache.

3. **Nit, rejected.** Pin the script's own `fonttools>=4.60`/`uharfbuzz>=0.56` ranges to exact versions for byte-for-byte reproducibility. Correct concern, but `shared/brand/tools/build-wordmark.py` is outside this card's file scope (`.github/preflight.json`, `shared/brand/package.json`, `shared/brand/README.md` only); pinning belongs with whoever owns that script.

4. **Nit, rejected.** Narrow `"when": ["shared/brand/**"]` to the specific files the script reads. The card specified `shared/brand/**` explicitly, and it matches the prevailing convention already used by six other checks in this file for the same package (`website-lint`, `website-unit`, `website-build`, `hub-web-unit`, `hub-web-build`, `website-types`, `website-e2e`, `website-image`); `guide-pdf`'s narrower glob is the outlier because it also spans an unrelated project (the hub). Kept as specified.

Full review: agent://Wordmark.ReviewWordmark

PR #177, opened 2026-09-17
fix(pigrocrm-web): carry the deep link a login interrupts back to it

fiorelorenzo:
## What this changes

I made the `/app` guard carry the current basepath-relative `href` as a `redirect` search param when it bounces an unauthenticated visitor to `/app/login`, and made `LoginPage`'s post-session effect navigate there once a session exists, instead of always landing on `/app`. It only honors a `redirect` that points to a same-origin path under `/app`; anything with a scheme, a protocol-relative address, or naming `/app/login` or `/app/entra` itself falls back to the dashboard exactly as before. `safeAppRedirect` (new, in `lib/tenant.ts`) is the check.

## How I verified it

`npx vitest run src/routes/app/login.test.tsx src/lib/auth.test.tsx src/lib/tenant.test.ts` (18 passed), `npx eslint` on the four touched files (clean), `npx tsc --noEmit` (clean). Added two cases to the existing login route test: a valid redirect (`/app/fatture/42`) is honored, an external one (`https://evil.example/steal`) is rejected and falls back to the dashboard.

`navigate({ href: redirect })` bypasses the router's typed `to` matching by design, since the redirect target is one of many possible `/app/*` routes and cannot be a compile-time literal. It still only accepts values `safeAppRedirect` has already restricted to `/app/...`, so this cannot reach another origin.

## Screenshots and video

Nothing visible changed in either page's markup; this is a navigation-target change. No before-and-after pair applies.

Linear: REB-237.


fiorelorenzo:
Reviewer subagent findings and what I changed for them:

- `safeAppRedirect` judged the raw string, so a plain or percent-encoded `..` segment (`/app/../evil`, `/app/%2e%2e/evil`) walked a visitor to an arbitrary same-origin path outside `/app`. Fixed: the value is now parsed with `URL`, decoded, and normalized before the `/app/` check.
- The public-route exclusion compared the raw string against only two of the three public routes (missing `/app/registrati`) and was defeated by a trailing slash, case, or percent-encoding. Fixed: it now shares `PUBLIC_APP_ROUTES" with the guard in `app.tsx` and compares the same normalized, decoded, lowercased path.
- The two full-page hops in `LoginPage` for a root installation with a name (the alias replace to the bare `/app/login`, and the assign to `/${rootSlug}/app/`) discarded the `redirect`, making the feature inert for that installation shape. Fixed: both now carry it.
- A stale comment in `app.tsx` attributed `safeAppRedirect` to `lib/auth.tsx`; it lives in `lib/tenant.ts`. Fixed.
- Added direct cases for `safeAppRedirect` in `tenant.test.ts` covering the shapes above, instead of only reaching it through the page's effect.

Noted but not changed, out of scope for this card: the link-by-mail sign-in (the default way in since ORB-172) lands on `/app/entra`, whose own `validateSearch` only accepts `t`, so an interrupted deep link is still lost for that path. Closing it needs the emailed link itself to carry the target, which is a server-side change.

fiorelorenzo:
Rebased onto `main` (`869d12f8`), since REB-300 moved the eighteen primitives into `@rebase/ui` and this branch was the one PR of the three that stopped merging. The conflict was a single import in `routes/app/login.tsx`: the branch adds `useRef` on the line REB-300 repointed from `sonner` to `@rebase/ui/sonner`, so the resolution keeps both. Nothing else in the diff changed.

Re-verified on the rebased tree: 1161 CRM tests, `tsc --noEmit` clean, `vite build` clean, lint clean. Still waiting on your review, @aciceri, together with #174 and #179, which are both still clean.

One thing worth knowing for those two: `toast` is now imported from `@rebase/ui/sonner` everywhere in the CRM, and only the package declares `sonner`, because the toast queue is module state and two resolutions mean the `<Toaster>` listens to one queue while `toast()` pushes onto the other. If either PR grows a new `toast()` call, it comes from there.

PR #179, opened 2026-09-17
feat(pigrocrm-web): page the Fatture list past the first fifty invoices

fiorelorenzo:
## What this changes

`useInvoices` sent no `limit` or `cursor`, so the API's own default of 50 newest invoices was silently the whole Fatture list: `next_cursor` was dropped on the floor, and any invoice past the 51st for a space was unreachable from the UI, with no notice that anything was missing.

I add a `useInvoicesPaged` hook that walks `next_cursor` with `useInfiniteQuery`, the way `fetchAllDeals` (`features/deals/queries.ts`) already walks it for the Kanban -- except that hook exhausts the cursor eagerly because the board needs every deal for its column totals, and this one stops after each page and lets the person reading the register ask for the next fifty. `useInvoicesList` (the list page) and `useInvoicesForOwner` (the tab) both sit on top of it. Both `InvoicesList` and `InvoicesTab` show a new `LoadMoreInvoices` affordance -- a "Carica altre" button and "Mostrate N fatture, ce ne sono altre." -- whenever the last loaded page's own `next_cursor` says there is more.

The API is untouched. Every filter the list takes (`tipo`, `stato`, `escludi_consumate`, `scadute`) is a server-side query parameter already (confirmed against `apps/api/src/pigrocrm_api/routers/invoices.py`), not something applied to a client-side page, so a chip or the type select changing mid-walk starts a fresh first page from the server rather than filtering rows already on screen -- there is no remaining "filters only see the loaded page" limitation to flag.

## How I verified it

- `pnpm exec tsc --noEmit -p .` (after `vite build` to regenerate the gitignored `routeTree.gen.ts`, otherwise every route file fails typecheck for an unrelated, pre-existing reason): clean.
- `pnpm exec eslint src/features/invoices/queries.ts src/features/invoices/InvoicesTab.tsx src/features/invoices/LoadMoreInvoices.tsx src/routes/app/fatture/index.tsx`: no findings.
- `pnpm exec vitest run src/routes/app/fatture/index.test.tsx src/features/invoices/InvoicesTab.test.tsx src/features/invoices/queries.test.tsx`: 25 passed, 25 total, including a new case (`offers "Carica altre" when the page is truncated, and loads the next page on click`) that mocks a first page with a `next_cursor` and a second page without one, and asserts the button appears with the right count, and that clicking it appends the second page's row and the button disappears once the server reports no more.

`ConsumedProformaNotice.tsx` still calls the original, untouched `useInvoices({ limit: 1, ... })` for a single-item lookup -- I deliberately did not fold that call site into the new paged hook, since it wants exactly one row and not a walk.

## Screenshots and video

Not captured. This wave's constraints explicitly prohibit Playwright, which is this repository's own documented mechanism for both the screenshot pair (Playwright MCP) and the video (`docs/pr-screenshots/record.mjs`, built on `@playwright/test`). The new markup is a plain `<Button variant="outline" size="sm">` and a `text-sm text-muted-foreground` span, the same primitives `NewProformaDialog`'s own row-adding button and the list's "Rimuovi il filtro" row already use elsewhere on this same page, and the new vitest case asserts the exact rendered text and ARIA role of both the button and the count.

Linear: REB-231.


fiorelorenzo:
Read-only reviewer went over the diff. It called the cursor walk itself correct (pageParam handling, getNextPageParam, filter-driven cache key, useInvoices left untouched with no collision today) and flagged five P3 findings, which I applied in a second commit:

- The paged cache shared its key shape with useInvoices's single-page InvoicePage, which the two hooks' filter types do not rule out colliding on -- I gave it its own key (invoicesList) and updated useInvoiceInvalidation to invalidate both.
- items was rebuilt with flatMap on every render, defeating DataTable's row-model memo -- wrapped in useMemo.
- A failed next-page fetch was a silent no-op (DataTable suppresses its own error banner while rows are on screen) -- now surfaced through the same QueryErrorBanner every other failed read uses.
- The truncation count was a bare span -- gave it role="status" so a screen reader announces the appended page and the control's own disappearance.
- The test only proved *some* cursor was sent on the second request, not the right one -- added an assertion on the exact value, and used a two-row first page since a truncated single-row page cannot occur against the real API.

PR #180, opened 2026-09-17
fix(pigrocrm): trust the host vhost's X-Real-IP in the web container's nginx

fiorelorenzo:
## What this changes

Fixes the rate limiter keying every visitor on the same address. The web container's `spa.conf` overwrote `X-Real-IP` with `$remote_addr` on every proxied `location`, which inside the container is always the compose bridge gateway, not the real visitor -- so `ratelimit.client_key` collapsed every internet client into one bucket and REB-228's per-client budgets protected nothing. Adds `set_real_ip_from 172.16.0.0/12`, `real_ip_header X-Real-IP` and `real_ip_recursive off` at server level, so nginx's realip module rewrites `$remote_addr` from the host vhost's trusted `X-Real-IP` before any `proxy_set_header` line runs. `docker-compose.yml` declares no custom network for this stack, so compose's default bridge network is what reaches this container, and `172.16.0.0/12` covers the whole range Docker draws it from.

## How I verified it

Confirmed the realip module is compiled into `nginx:1.27-alpine` (the image `Dockerfile.web` builds on): `docker run --rm nginx:1.27-alpine nginx -V 2>&1 | grep -o with-http_realip_module` printed `with-http_realip_module`.

A full `docker build -f projects/pigrocrm/Dockerfile.web` was not realistic to run offline in this worktree (it needs a `pnpm install` of the web app against the network). Instead validated the edited config directly against the same base image: `docker run --rm -v <worktree>/projects/pigrocrm/deploy/nginx/spa.conf:/etc/nginx/conf.d/default.conf:ro nginx:1.27-alpine nginx -t`, which printed `nginx: the configuration file /etc/nginx/nginx.conf syntax is ok` and `nginx: configuration file /etc/nginx/nginx.conf test is successful`.

`ratelimit.py`'s `client_key` docstring already describes the fixed chain accurately (it documents the general need for the header, not the pre-fix hop overwrite), so it needed no change. The existing tests `test_the_link_request_is_throttled_per_client`, `test_disponibile_is_throttled_per_client` and `test_the_member_question_is_throttled_per_client` already assert that a request carrying a real `X-Real-IP` lands in its own bucket, so no new unit test was needed for `client_key` itself. Ran `uv run pytest -q projects/pigrocrm/apps/api/tests/test_auth_api.py projects/pigrocrm/apps/api/tests/test_tenants_api.py`: 55 passed.

`172.16.0.0/12` alone, not also `10.0.0.0/8` or `192.168.0.0/16`: `docker-compose.yml` for this stack declares no custom `networks:` block, so it uses compose's default bridge network, which Docker draws from `172.16.0.0/12`. If a future change gives this stack a custom network in the `192.168.0.0/16` or `10.0.0.0/8` range, this trusted-proxy range needs to widen with it.

This PR touches only server-level directives in `spa.conf`; it does not touch the `/app/entra` location block REB-229 is adding in parallel in the same file.

## Screenshots and video

Nothing visible changed; this is a server-level nginx config fix with no UI surface.


fiorelorenzo:
Ran a read-only reviewer subagent on this diff. Its one finding: 172.16.0.0/12 does not cover Docker's 192.168.0.0/16 fallback pool, which a host could land on if a custom default-address-pool were ever configured. I'm not widening the range for that: docker-compose.yml for this stack declares no custom network today, so 172.16.0.0/12 is what the card asked for and what actually matches the observed default-bridge gateway. If this stack's compose file is ever given a custom network in the 192.168.0.0/16 or 10.0.0.0/8 range, this trusted-proxy range needs to widen with it (noted in the PR body already). Everything else the reviewer checked (directive placement/order, real_ip_recursive being inert for a single-value X-Real-IP, no route regressions) came back correct; it verified end to end against nginx:1.27-alpine that every proxied location now forwards the real visitor address instead of the bridge gateway.

PR #181, opened 2026-09-17
fix(pigrocrm): throttle POST /api/auth/login before the argon2 verify

fiorelorenzo:
## What this changes

Fixes `POST /api/auth/login` being unauthenticated and unthrottled while every attempt costs a full argon2id verify at the library's own defaults (64 MiB, time cost 3, parallelism 4). Any address could guess a password at CPU speed, and forty concurrent anonymous POSTs (uvicorn's default thread pool) would pin about 2.5 GiB in a container compose gives no memory limit.

`login` now calls `spend_one(request, scope="login")` as the very first thing it does, before `UserService.authenticate` runs, so a request past the budget never pays for the hash at all. It gets its own `LOGIN_REQUESTS_PER_MINUTE` (10) budget rather than sharing `REQUESTS_PER_MINUTE` with the signup routes: a scripted password sweep must not be able to starve someone else's signup, and a household sharing an address filling in the signup form must not eat into login's own budget. Ten, not five: a person locked out by a typo still gets two full retries before the wait, nowhere near enough attempts a minute to make guessing worthwhile. Answered with the existing `TOO_MANY_REQUESTS_RESPONSE` wording and `Retry-After`, the same as `link` and `disponibile`.

This depended on REB-269 (client_key seeing the real visitor instead of the compose bridge gateway on every request), merged as PR #180, or this budget would throttle the whole installation instead of one client.

## How I verified it

Ran `uv run ruff check projects/pigrocrm/apps/api` (all checks passed), `uv run mypy projects/pigrocrm/apps/api/src/pigrocrm_api/ratelimit.py projects/pigrocrm/apps/api/src/pigrocrm_api/routers/auth.py` (no issues), and `uv run pytest -q projects/pigrocrm/apps/api/tests/test_auth_api.py` (35 passed), including the new `test_login_is_throttled_per_client_before_the_password_is_checked`, which asserts the eleventh wrong-password attempt from one client is 429 with `Retry-After: 60` while a fresh client (a different `X-Real-IP`) still gets 401 on the same request.

The budget (10/minute) is a judgment call sized for a person mistyping their password, not derived from any existing constant; `REQUESTS_PER_MINUTE` (5) and `DISPONIBILE_REQUESTS_PER_MINUTE` (30) are the two nearby reference points and this sits between them for the reason in the commit body.

## Screenshots and video

Nothing visible changed; this only adds a 429 for a client already past its login budget, with the existing 429 response shape used elsewhere in this API.


fiorelorenzo:
Ran a read-only reviewer subagent on this diff. It found the ordering, scope namespacing and 429 response merge all correct and confirmed spend_one runs before UserService.authenticate (auth.py). Two P3 findings, both applied in a follow-up commit:

1. My rate-limit comment had replaced, not joined, the existing comment explaining why login's 401 is byte-identical for all three failure causes (anti-enumeration, pinned by test_login_failures_are_byte_identical_regardless_of_cause). Restored it.
2. The new test's docstring claimed to prove spend_one runs before authenticate, but its assertions alone could not distinguish that from a limiter placed after the verify and before the 401 raise. Rewrote it to wrap authenticate with a counting spy and assert it is entered exactly LOGIN_REQUESTS_PER_MINUTE times once the 429 arrives.

Everything else the reviewer flagged as deliberately out of scope for this card (nginx header trust is REB-269, generated web client types, per-account lockout) I agree is out of scope.

PR #182, opened 2026-09-17
fix(hub): keep the magic-link token out of PostHog

fiorelorenzo:
## What this changes

I take the hub's magic-link token out of the URL in `main.tsx` before `initAnalytics` runs, the same shape as REB-229's web half in the CRM (PR #174): a small `lib/entra-token.ts` with `stripEntraToken()`/`takeEntraToken()`, and `pages/member/Entra.tsx` reads it back with a lazy `useState` initialiser (`takeEntraToken() ?? t` from the search params), falling back to the search param for whatever reaches the page without `main.tsx` having run first.

I also add a `before_send` in `shared/analytics/browser.ts` that strips a `t` query parameter from `$current_url`, `$initial_current_url` and `$referrer` on every event, unconditionally for every surface that calls `initAnalytics`. Since the CRM initialises through the same `shared/analytics` package, it gets this scrubber for free: no change needed to its own `lib/entra-token.ts` or `main.tsx` to get the second brace this card asks for.

## How I verified it

- `cd shared/analytics && pnpm test`: 13 passed, including the two new cases in `browser.test.ts` ("removes t from $current_url, $initial_current_url and $referrer, and keeps every other param" and "is a no-op on an event with none of those properties, and on a null result").
- `cd shared/analytics && pnpm lint` and `pnpm exec tsc --noEmit -p tsconfig.json`: clean.
- `cd projects/hub/apps/web && pnpm exec vitest run src/lib/entra-token.test.ts src/pages/member/Entra.test.tsx`: 9 passed (5 new in `entra-token.test.ts`, the existing 4 in `Entra.test.tsx` unchanged in behavior since `takeEntraToken()` is null under jsdom's default location and the read falls back to the search param exactly as before).
- `cd projects/hub/apps/web && pnpm exec tsc --noEmit -p tsconfig.json` and `pnpm exec eslint` on the touched files: clean.

None deliberately left uncovered. The lazy `useState` initialiser in `Entra.tsx` follows the CRM's own comment on why Strict Mode's double-invoke does not double-spend the token.

## Screenshots and video

Nothing visible changed: this moves a token out of a URL before it is captured and scrubs a query parameter from three analytics properties. No UI to screenshot.

Linear: REB-273.


fiorelorenzo:
One read-only `reviewer` subagent looked at the diff and found the scrubber's real bug: `$initial_current_url` is not a member of a `CaptureResult`'s `properties`. posthog-js computes it once, from the first pageview this browser ever sent, and carries it as a person `$set_once` value, a sibling of `properties`, so the loop that only walked `properties` never reached it. I verified this independently against the pinned SDK (`posthog-js@1.429.5`, `dist/module.js`): `get_initial_props()` builds `$initial_current_url`/`$initial_referrer` from `$initial_person_info` and returns them into what `capture()` assigns as `f.$set_once`, never `f.properties`. Fixed in the second commit: the scrubber now walks `properties`, `$set` and `$set_once`, and the test builds the event shape posthog-js really emits (`$initial_current_url` inside `$set_once`) instead of one where it lived inside `properties`.

The reviewer also flagged that no test exercised the primary path this PR adds (a token `stripEntraToken` already took out of the URL, read back by `takeEntraToken`): every existing `Entra.test.tsx` case mounts with `?t=` still in the router's own search, so `takeEntraToken()` always returned null and the `?? fromSearch` fallback was all that ran. Added a test for it, and confirmed it catches a regression: reverting the `useState` initialiser to drop `takeEntraToken()` fails the new test.

One finding I did not act on: the reviewer noted a `useState` lazy initialiser calling a destructive read is only safe under render interruption if React never discards a partial initial mount, which is true here (no Suspense boundary, no transition, on the very first paint) but not a guarantee `useState`'s contract makes in general. I left it as is: it is the exact shape of the CRM's own `lib/entra-token.ts` (PR #174), which this card asks to mirror, and diverging here would give the two surfaces two different answers to the same one-shot-read problem for a risk that does not apply to this app's actual render tree.

PR #185, opened 2026-09-17
fix(hub): mask every text in session replay, not only inputs

fiorelorenzo:
## What this changes

`main.tsx` now calls `initAnalytics({ maskText: true })` for the hub, the same option the CRM already sets. The admin area renders every freelancer's and company's name, email (`mailto:` href), daily rate, LinkedIn URL, links and CV filename as plain text; a session recording of an admin browsing `/hub/admin/*` was a copy of the candidate database at a third-party processor, and autocapture on a clicked `mailto:` or LinkedIn anchor recorded the address in `attr__href`/`$el_text`. Inputs were already masked (the wizards' own reasoning), but nothing else was.

The wizard funnel events (`wizard_iniziato`, `wizard_passo`, `wizard_completato`) lose nothing: they are explicit `capture()` calls from `lib/analytics.ts`, sending their own structured properties (`tipo`, `perk`, `passo`, `passi`), entirely unrelated to `session_recording`'s masking config, which only ever threads into `maskTextSelector`/`maskAllInputs` for the replay recorder.

**The privacy note.** I checked what promises anything about replay. It is not the hub's own text: `Shell.tsx` links to `https://letsrebase.com/privacy`, which is `projects/website/src/privacy.html`, shared between the CRM and the hub, so I did not edit it (out of this project, and out of this card's file ownership). It currently says (§ Cookie, around line 108): "*...anche come registrazione della sessione in cui **ogni campo è mascherato, e nel CRM anche ogni testo**, così una fattura o il nome di un cliente non compaiono mai.*" -- "every field is masked, and in the CRM also every text." That sentence draws a distinction between the two surfaces that no longer holds once this PR ships: the hub now masks every text too, the same as the CRM. It does not promise anything false (the hub's own promise, "every field is masked," stays true), it just understates the hub's protection now. Reporting it here per the card; the website's own owner can decide whether and how to update the sentence.

I also amended `docs/design/2026-09-12-posthog-analytics-design.md`'s hub section to record the change, the same way the 2026-09-15 REB-215 amendment did in the same document, and corrected one now-stale test title in `shared/analytics/browser.test.ts` ("which is what the CRM asks" -- now both surfaces ask for it).

## How I verified it

- Read `lib/analytics.ts`: `useWizardAnalytics` calls `capture('wizard_iniziato'|'wizard_passo'|'wizard_completato', ...)` directly; `shared/analytics/browser.ts`'s `maskText` option only ever sets `session_recording.maskTextSelector`, never touching autocapture or explicit `capture` calls.
- `cd shared/analytics && pnpm test` (14 passed), `pnpm lint`, `pnpm exec tsc --noEmit -p tsconfig.json`: all clean (no behavior change here, only the corrected test title).
- `cd projects/hub/apps/web && pnpm exec eslint src/main.tsx` and `pnpm exec tsc --noEmit -p tsconfig.json`: clean. `main.tsx` is an entry point with no dedicated test file and nothing in the suite imports it directly; the option itself is exercised by `shared/analytics/browser.test.ts`'s existing `session_recording` assertions, which still pass.

Whether the privacy.html sentence needs its own follow-up card to reword the CRM-only text-masking claim -- I left that to whoever owns `projects/website`.

## Screenshots and video

Nothing visible changed: this only widens what a session recording masks. No UI to screenshot.

Linear: REB-274.


fiorelorenzo:
One read-only `reviewer` subagent found that `maskText: true` was not enough to close the leak this card is about. `session_recording.maskTextSelector` masks replay's own text nodes only. Two other paths stayed open: autocapture (a separate, top-level `mask_all_text`/`mask_all_element_attributes` pair in `PostHogConfig`) still sent an element's text and attributes -- including a `mailto:` anchor's `href` -- as ordinary event properties on every click, and replay's own masking never reaches DOM attributes at all, so a masked link's visible text hid nothing while its `href` still carried the address in the recording (`session_recording.maskAllElementAttributes` is the separate switch for that).

I verified both claims independently against the pinned SDK before acting: `posthog-js@1.429.5`'s own `dist/module.js` wires `mask_all_text`/`mask_all_element_attributes` straight into the autocapture config object (`t.maskAllElementAttributes=i.mask_all_element_attributes,t.maskAllText=i.mask_all_text`), and `@posthog/types@1.411.1`'s `posthog-config.d.ts` documents `session_recording.maskAllElementAttributes` as masking "every string-valued source DOM attribute, including... `src`, `href`" -- a separate mechanism from `maskTextSelector`, which its own doc comment says explicitly "do[es] not hide an image, whose `src` is still recorded."

Fixed in the second commit: `maskText` now sets all three switches. This also closes the same latent gap for the CRM, which already asked for `maskText` and had never set the other two either -- one fix in the shared package, both surfaces. Updated the test to assert all three, and confirmed it fails against the prior implementation and passes against the fix (reverted the code, watched the new assertion fail, restored it). Also corrected two now-stale "the CRM adds text masking" mentions the reviewer flagged (`README.md`, the design doc's package description) since they described the pre-fix distinction that no longer holds.

PR #188, opened 2026-09-17
feat(infra): the CRM, hub and website vhosts send HSTS, a CSP, frame-ancestors, nosniff

fiorelorenzo:
## What this changes

Neither nginx hop in front of the CRM, the hub or the website sent HSTS, a CSP,
`frame-ancestors`, a `Permissions-Policy` or, on most of them, `X-Content-Type-Options`
(REB-271, REB-275). Now every host vhost (`pigro.letsrebase.conf`,
`preview.pigro.letsrebase.conf`, `letsrebase.conf`, `preview.letsrebase.conf`) sends,
with `always`:

- `Strict-Transport-Security: max-age=31536000; includeSubDomains` (no preload)
- `X-Content-Type-Options: nosniff`
- `Referrer-Policy: strict-origin-when-cross-origin`
- `X-Frame-Options: DENY`, backed by an **enforcing** `Content-Security-Policy` that
  carries only `frame-ancestors 'none'` — this one costs nothing (nothing here is ever
  meant to be framed) so it ships live, not report-only
- `Permissions-Policy: camera=(), microphone=(), geolocation=()`
- `Content-Security-Policy-Report-Only`, the untested part, with a policy read against
  what each surface actually loads (see "Anything a reviewer should look at twice")

**Where each header lives, and why.** The host vhost is the TLS terminator and the one
hop that sees every response, so the full set lives there, in one included snippet per
project (`projects/pigrocrm/deploy/nginx/security-headers.conf`,
`projects/website/deploy/security-headers.conf`, the second covering the website and
the hub since they share the apex vhost). A location that ever needs its own
`add_header` only has to repeat the `include` line, not seven directives by hand —
nginx drops every server-level `add_header` for a location that declares its own.
`deploy/setup-server.sh` inlines the same set instead of including it: that script has
nothing that provisions `/etc/nginx/snippets/` on a fresh self-hosted install, and an
`include` of a path that does not exist fails `nginx -t` outright. The three
container-internal configs (`spa.conf`, the website's and the hub's own `nginx.conf`)
keep only `X-Content-Type-Options: nosniff`, for whoever reaches a container directly
without the host vhost in front (a local `docker compose`, a health check) — a
`Strict-Transport-Security` or a CSP set there would be meaningless on the plain HTTP
those containers speak inside the compose network. The hub's own weaker
`Referrer-Policy: strict-origin` is removed in favour of the host's stricter
`strict-origin-when-cross-origin`: sending both would leave a browser to pick between
two differently-valued instances of the same header rather than trusting one.

**The repository copies change here; the host is edited in place afterwards**, the same
shape as REB-107 and REB-252. Each vhost's own top comment now says what has to be
pasted into the installed, certbot-rewritten copy and where the snippet file has to
reach the host (`/etc/nginx/snippets/pigrocrm-security-headers.conf`,
`/etc/nginx/snippets/website-security-headers.conf`, beside the existing
`orbiters-proxy.conf`).

## How I verified it

- `pnpm --filter web test` (114 files, 1205 tests) — includes `robots.test.ts` and
  `upload-body-size.test.ts`, which read `pigro.letsrebase.conf`/`preview.pigro.
  letsrebase.conf`/`spa.conf` as fixtures; both still pass unchanged, since the new
  headers reach those files through an `include` line rather than a literal
  `add_header`, so `robots.test.ts`'s count of exactly one `add_header` before the
  first `location` (the X-Robots-Tag line) still holds.
- `pnpm --filter website test` (13 files, 300 tests) — includes `path-map-plugin.
  test.ts` and `pixel.test.ts`, which read `letsrebase.conf`/`nginx.conf`; both pass.
- `nginx -t` inside `nginx:alpine`, mounted read-only, for each of the three
  container-internal configs (`spa.conf`, the website's `nginx.conf`, the hub's
  `nginx.conf`) standalone: all three report "syntax is ok" / "test is successful".
- `nginx -t` for all four host vhosts together (`pigro.letsrebase.conf`, `preview.
  pigro.letsrebase.conf`, `letsrebase.conf`, `preview.letsrebase.conf`, `__ROOT_SLUG__`
  substituted), with both new snippet files and the existing `orbiters-proxy.conf`
  mounted at `/etc/nginx/snippets/`: "syntax is ok" / "test is successful". These files
  have no `ssl_certificate` directive to stub — that only exists in the certbot-rewritten
  copy on the host, never in the repository's plain-HTTP source of truth.
- `setup-server.sh`'s generated vhost: extracted the exact heredoc lines (`sed -n
  '32,61p'`) and ran them through real bash (so `\$host`/`` \` `` are unescaped exactly
  as the script itself would write them), then `nginx -t` on the result: passes.
- Built the real images and checked the real running containers, not just the config
  files: `docker build -f projects/pigrocrm/Dockerfile.web .`, `-f projects/hub/
  Dockerfile.web .`, and `docker compose build` in `projects/website` all succeed;
  `docker run --rm --entrypoint nginx <image> -t` passes on all three; running each
  image and `curl -sD -` against it shows `X-Content-Type-Options: nosniff` on a real
  response (`/app/` for the CRM, `/hub/` for the hub, `/` for the website), and the
  hub's response no longer carries the old `Referrer-Policy: strict-origin`.
- No `preflight.json` check builds the host vhosts or runs `nginx -t` against them
  directly (only `pigrocrm-images`, which builds `spa.conf` into the CRM's own image,
  covered above); `check-unindexable.sh` reaches live hosts over HTTPS and is excluded
  from PR/push CI, unaffected by this change (no robots-related header or location was
  touched).

**The `Content-Security-Policy-Report-Only` policy is a read of the code, not a
measurement of the running site**, and I want to be upfront about where it is a guess:

- The CRM (`script-src 'self' https://eu-assets.i.posthog.com`) and the hub apex vhost
  need `style-src 'unsafe-inline'` because several React components set a computed
  `style={{ ... }}` (`apps/web/src/features/dashboard/MonthlyBars.tsx` and `charts.tsx`
  in the CRM, `src/wizard/Wizard.tsx`'s progress bar in the hub) — that is a real inline
  style attribute a CSP has to allow explicitly, and I chose `'unsafe-inline'` over
  threading a nonce through every one of them.
- The website's apex vhost needs `script-src 'unsafe-inline'` too: `index.html`,
  `pigrocrm.html` and `community.html` each carry a three-line reveal-gate script and a
  JSON-LD block, and `privacy.html` carries the consent-withdrawal control — all three
  inline, with no bundler to hash or nonce them (the project's whole point is a
  300ms build with none). `bzrcdn.openai.com` (the ChatGPT Ads pixel's SDK) is in
  `script-src`/`connect-src` even though `consent.js` only loads it after a visitor
  says yes: the policy has to allow what the page *can* load, not what a given request
  happened to load.
- `connect-src` carries `eu.i.posthog.com` and `eu-assets.i.posthog.com` everywhere
  PostHog is used; I could not find where the ChatGPT Ads SDK's own beacon calls
  actually go (its script is minified and third-party), so `bzrcdn.openai.com` is in
  `connect-src` as the only OpenAI-related literal in the codebase. If that guess is
  wrong, report-only mode is exactly the safety net for finding out: a violation
  shows up in a browser's console, nothing breaks.
- No `report-uri`/`report-to` collector exists in this installation, so today that
  console line is the only signal — nobody here sees it unless they open devtools.
  REB-306 (filed alongside this PR, same milestone as REB-271) tracks watching one
  release with the policy live and then promoting it to enforcing; whether to also
  stand up a collector is left to whoever picks that up.

**Two separate CSP headers, deliberately**: `frame-ancestors 'none'` ships in its own,
enforcing `Content-Security-Policy` header (nothing here is ever meant to be framed, so
it costs nothing to enforce from day one), while everything else — the real guess above
— stays in `Content-Security-Policy-Report-Only`.

## Screenshots and video

No visible change: every header here is invisible in the browser chrome, and the one
thing a person could watch for (a blocked resource) cannot happen yet, by design — the
resource-loading part of the policy is report-only. Verified instead with `curl -sD -`
against the running containers, in "How I verified it".

Linear: REB-271, REB-275.


fiorelorenzo:
**Review applied:** a fresh read-only `reviewer` on this PR's diff found three real issues, all fixed in ce339cf6:

1. The report-only CSP had no `frame-src`, so the CRM's invoice PDF preview (`InvoicePdfPreview.tsx`, a `blob:` object URL in an `<iframe>`) violates it on every view and would blank out at the enforcing flip. Added `frame-src 'self' blob:` to the CRM snippet and to `setup-server.sh`'s inlined copy.
2. Both snippets used `Referrer-Policy: strict-origin-when-cross-origin`, weaker on a same-origin request than the `strict-origin` this PR removed from the hub's own `nginx.conf` — and both the hub's and the CRM's magic-link routes (`/hub/entra?t=`, `/app/entra?t=`) carry the token in the URL, which their own same-origin assets then echo back in `Referer` before the token is stripped client-side. Switched both snippets (and `setup-server.sh`) to `strict-origin`, identical cross-origin, costs nothing same-origin either app needs (checked: nothing here reads `document.referrer`).
3. `setup-server.sh`'s HSTS line dropped `includeSubDomains`: that script writes a vhost on a domain the project does not control (a self-hoster's own), where pinning every sibling subdomain to HTTPS for a year from one visit could break something the installer has no say over. The two committed production vhosts keep `includeSubDomains`, since `letsrebase.com`/`pigro.letsrebase.com` are names the project does own.

A fourth, lower-severity note (the `style-src 'unsafe-inline'` rationale cited React's `style={{}}` props, which are CSSOM writes CSP does not govern — `radix-ui`'s runtime `<style>` injection, confirmed in `package.json`, is the real reason) is also corrected in the same commit, since it was in a file whose whole body is that reasoning.

Nothing was rejected; all four points landed.

PR #192, opened 2026-09-18
feat(ui): declare the application tokens once in @rebase/ui

fiorelorenzo:
## What this changes

`shared/ui` (`@rebase/ui`), one `tokens.css` both SPAs import in place of their own copy, at the weights the application-variant record chose on 2026-09-18 (`docs/design/2026-09-18-application-variant-brief.md`, REB-298): radius zero across the derived scale, the line is the ink itself at 1px with the site's 2px named beside it (`--line`, `--line-strong`) for the primary button and the focus ring, one 4px ink step with no blur on what floats and nothing on what rests, the site's 16px grid back on the body at 4% where the site draws 7%, the Prussian Blue sidebar, the five chart tints, and no `.dark` block.

Three things that are not a straight move:

- The hub keeps `.site`, reduced to what still differs from the shared default: the public wizards are the landing continued and carry its 2px line, its 8px step and its 7% ground. Squared corners and the ink colours are no longer that scope's business, since they are now the default everywhere.
- The `dark:` variant stays bound to a class nobody sets. Without the binding Tailwind v4 falls back to `prefers-color-scheme`, and the primitives' `dark:bg-input/30` would fill a visitor's fields with ink at 30% over the paper, which is the bug ORB-138 fixed in the hub. The block those utilities would read is gone; the binding that keeps them inert is not.
- The CRM's toaster stops reading `next-themes`, which handed sonner `system`, the visitor's OS preference, and was the last path to a dark surface in a product with no dark tokens left. The dependency goes with it.

The contract test moves into the package and runs once (30 assertions): every colour resolved back through the brand palette, the five `color-mix()` chart expressions evaluated the way a browser does, and one assertion per sentence of the record. The hub's own test keeps the `.site` weights and the guard that nothing in the app ever adds the bare `dark` class; the CRM's file is deleted, since nothing in it was CRM-local.

CI gains a `shared_ui` filter and a `shared/ui · web` job (the SPA gates install the package but never run a dependency's own scripts, the same reason `shared/analytics` has one), `shared/ui/**` joins each consumer's filter so a token change re-runs both apps, and `preflight.json` gains `ui-unit` plus the same trigger on the six app checks.

## How I verified it

- `pnpm --filter @rebase/ui test` 30 passed, `lint` clean, `tsc --noEmit` clean.
- `pnpm --filter hub test` 168 passed in 26 files, `pnpm --filter web test` 1177 passed in 113 files, both apps `lint` and `build` clean.
- The compiled CSS, not the source: `dist/assets/index-*.css` of both apps carries `--radius:0px`, `--shadow-app-md:var(--step) var(--step) 0 var(--color-prussian-blue)`, `--border:var(--color-prussian-blue)`, `background-size:var(--grid-cell) var(--grid-cell)`, and zero occurrences of `prefers-color-scheme`. The hub's own file still overrides `--shadow-app-md` to `var(--landing-step) var(--landing-step) 0 var(--landing-ink)` inside `.site`. The only `.dark` left in either bundle is the `:is(.dark *)` utilities.
- Rendered, in headless Chrome against `vite preview` of both builds at 1440px: a `shadow-lg rounded-xl border` element resolves to `rgb(1, 25, 54) 4px 4px 0px 0px`, radius `0px`, border `rgb(1, 25, 54)`, and the body paints `16px 16px` tiles of the 4% mix. Screenshots below.

- The import order. `@import '@rebase/ui/tokens.css'` comes after `@import 'tailwindcss'` on purpose: before it, Tailwind's own defaults would win for the keys this file sets. Both builds were read back out of `dist` to prove the values survive, which is the same check the hub's old comment describes for the shadow indirection.
- `--shadow-xs`/`--shadow-sm` are `none`. Every generated primitive that asks for `shadow-xs` or `shadow-sm` now casts nothing, which is the record's "nothing that sits in the page carries a shadow". `shadow-md`/`lg`/`xl`/`2xl` all resolve to the one 4px step: four rungs of a blurred scale collapse into one offset, since an offset has no second magnitude to graduate.
- The chart tints moved as they were, values and comment: they were computed for contrast against both grounds and the test still records the two WARNs (a mark below 3:1 on its worse ground, a pair at deltaE 9.6) rather than hiding them.
- The primitives are untouched. A component that reads `--line-strong` or draws the 2px primary button is REB-300; this card only makes the values exist in one place. The same goes for the hub's admin area moving off its old shapes, which is REB-302 after REB-279.

## Screenshots and video

Rendered at 1440px against `vite preview` of both builds, in `~/.cache/tmp/reb299/`:

- `crm-login.png`, `hub-admin-login.png`: the application weights, squared, 1px ink lines, cards with no shadow, the 4% grid behind them.
- `hub-wizard.png`, `hub-chooser.png`: the `.site` scope unchanged, 2px lines, the 8px step under the card, the 7% ground.

Linear: REB-299.


fiorelorenzo:
Second commit fixes a hole this PR's own CI run exposed: `shared/ui · web` reported `skipping` even though `shared/ui/**` is exactly what changed. The `changes` job maps each filter to an output by hand and I had added the filter without the mapping, so `needs.changes.outputs.shared_ui` was the empty string. The next run is the proof: the job has to report a pass, not a skip.

fiorelorenzo:
Third commit applies the review. The blocker was outside the diff and CI could not have caught it on a PR: neither `Dockerfile.web` copies `shared/ui`, so pnpm linked the new workspace dependency to a path absent from the image and both web builds died resolving `@rebase/ui/tokens.css`; the image jobs are gated on `github.event_name != 'pull_request'`, so it would have broken the trunk and both deploys after the merge. Both Dockerfiles copy it now and I built both images locally to prove it.

The second real finding: with `--radius: 0` the CRM was rendering literal corners next to the squared ones (sidebar items and switcher at 10px, content panel at 16, checkbox 4, tooltip arrow 2, dashboard bar 3) plus two `input-group` rungs computing `calc(var(--radius) - 3px)`, a negative radius the browser drops. They read the scale or nothing now, the panel test says what it draws instead of pinning the old literal, and the built CSS has no `border-radius` other than 0 and the pill sentinel. Also applied: two assertions the move had dropped are back (`--secondary` stays a palette mix, `--sidebar-accent-foreground`), and `ui-unit`'s preflight trigger lists `shared/brand/font.css`, which its own test reads.

Left for REB-300 with the components: three non-semantic tint borders (`border-muted-foreground/30`, `border-foreground/30`, `border-input/30`) that the record's ink line replaces.

PR #193, opened 2026-09-18
refactor(ui): move the eighteen primitives into @rebase/ui

fiorelorenzo:
## What this changes

The eighteen radix-backed primitives move into `@rebase/ui`, one export path each, and the two applications stop carrying their own copies: the hub had seven written by hand with class-variance-authority, the CRM had twenty generated ones with the same seven among them, so a button fixed in one product stayed broken in the other. The hub's seven are deleted rather than merged, since the CRM's are the ones with the states, the variants and the tests behind them, and `cn` moves with them, because it was the same five lines in each application's `src/lib/utils.ts`. Five of the CRM's components stay where they are (avatar, calendar, command, input-group, sidebar): none has a second consumer, the hub's shell is its own and its screens have no command palette, no date picker and no avatar, so moving them would be moving one product's code into a package for three. That choice is written in the package README, where the next person will look for it.

Three things the move had to decide rather than port. `tw-animate-css` and `shadcn/tailwind.css` back the primitives' enter and exit transitions and the `data-open`/`data-checked` variants their class names use, and both have to be imported before Tailwind expands, so they are now `@rebase/ui/before-tailwind.css` and no longer two lines the hub happened to be missing. `tokens.css` carries `@source './*.tsx'`, because Tailwind scans no path under `node_modules`, which is where an application sees this package, and a class used only by a primitive would otherwise never be emitted. And every `dark:` utility is stripped from the moved files: there is no dark theme since 2026-09-18, so they were dead weight waiting for somebody to bind the variant to the OS preference again, which is the ORB-138 bug. The hub's own test now proves the word appears nowhere in that application, in any form.

The package's build is the gallery: one page with every primitive in every variant, size and state, plus `?open=dialog|sheet|menu|popover|select` for the surfaces that only exist behind a trigger and a portal. CI builds it, which is what proves the eighteen still compile against Tailwind and the tokens outside an application, and REB-301 runs `uislop` against it.

It also closes the three findings REB-299 deferred: the two tint borders in `UploadDropzone` and `TimerBar` and the command palette's `border-input/30`, which the record's ink line replaces, with the running timer now reading `--line-strong` instead of a 30% fade that would have looked lighter than the resting border.

## How I verified it

Everything that follows was run in the worktree, and the numbers are the ones the commands printed.

- `pnpm --filter @rebase/ui lint`, `test` (4 files, 55 tests: the token contract plus the three primitive suites that moved with the components), `build`, `tsc --noEmit`: all green.
- `pnpm --filter hub test`: 168 tests in 26 files. `pnpm --filter web test`: 1152 in 110, which is 1177 minus the 25 that moved into the package. Both apps lint, build and typecheck.
- **Both web images built locally** (`docker build -f projects/{pigrocrm,hub}/Dockerfile.web`), because the image jobs skip on a pull request and this is exactly the shape of failure that broke nothing on the PR and everything on the trunk in REB-299.
- In the CRM's built CSS, the only `border-radius` values are `0` and the pill sentinel, and `prefers-color-scheme` appears zero times. In the hub's, `animate-in`, `data-checked` and `shadow-lg` are present, which is the `@source` line doing its job: before it, those classes existed only in a file Tailwind never scanned.
- Rendered, in headless Chrome at 1280 and 1440: the gallery (contact sheet attached), the dialog and the sheet open, and then the two products served **from the images I built**, not from a dev server. The CRM login and signup and the hub admin login compute `border-radius: 0px`, a 1px border and `rgb(1, 25, 54)` on the field. The hub's wizard computes 2px and the site's own paddings on the same components, which is the `.site` scope still reaching the primitives after the move, and it looks identical to before.

The `@source './*.tsx'` line in `tokens.css`, because it is the one thing in here that is invisible when it breaks: if Tailwind ever stops resolving `@source` relative to the file that declares it, both products lose every utility that only a primitive uses and nothing fails, it just renders wrong. I would rather have that in the package than two copies of a path back into `node_modules` in the applications, but it is a bet on a Tailwind behaviour.

Second, `before-tailwind.css` as a separate entry point. The alternative was putting those two imports inside `tokens.css`, which is one import for an application instead of two, and I did not take it: `tokens.css` is imported after Tailwind on purpose, and an `@import` that lands after the expansion is dropped by the optimiser rather than reordered. If a third application is ever added and forgets the first line, its primitives render with no transitions and with the `data-*` variants dead, which is a quiet failure. A test cannot see it from inside the package, so the README says it in the one place someone will read.

Third, the five components that stayed in the CRM still carry their `dark:` utilities, and `tokens.css` still binds the variant to a class nobody sets. That binding is what keeps them harmless. Cleaning them up belongs with whatever card moves them, or with REB-302.

## Screenshots and video

The gallery is the screenshot: `docs/design/application-variant/` already holds the contact sheet of the brief, and this page is the implemented counterpart. Attached below: the gallery full page, the dialog and the sheet open on the step shadow, and the CRM login and the hub wizard served from the two images, which is the before-and-after that matters here (they are meant to look identical: the primitives moved, the rendering did not).

Linear: REB-300.

gallery-full

open-dialog

open-sheet

crm-login

hub-wizard


fiorelorenzo:
Second commit applies the review. The one that mattered most was not the red test: `radix-ui` and `sonner` were declared with independent ranges in two manifests, and sonner keeps its toast queue in module state, so a one-sided bump would have handed the `<Toaster>` one queue and `toast()` another and every toast in the CRM would have stopped appearing with a green build and a green suite. `@rebase/ui/sonner` exports `toast` beside `Toaster` now, the CRM's 47 call sites and 28 `vi.mock` targets go through it, and the application no longer declares sonner at all. Proving that end to end turned up one more thing: the toast was the last floating surface on a blurred shadow, since sonner hardcodes `0 4px 12px` and exposes no variable for it, so its class carries `shadow-md!` and it now casts the same 4px ink step as the dialog (measured in the browser: `rgb(1, 25, 54) 4px 4px 0px 0px`).

The red test was mine and the review is right about why: `badge.test.tsx` moved byte for byte while `badge.tsx` had its pill fill rewritten to `bg-paper` in the same commit, so the suites I reported green had been run against the working tree rather than against the commit. Fixed, and the two dot tints finish the conversion they had half of.

Also applied: `components.json` exists in the package, so a nineteenth primitive is generated where the other eighteen live, and the CRM's `utils` alias names `@rebase/ui/cn` instead of the file this branch deletes, which is exactly what the next `shadcn add --overwrite calendar` would have written. The step-shadow scan that travelled with `shapes.test.tsx` had stopped covering the five components that stayed in the CRM, so those have their own test, which also pins that they read the package rather than keeping a second copy. `font-heading` is deleted from the four files that claimed it, since the brand ships one typeface and `--font-heading` is declared nowhere, so it emitted nothing. The gallery renders the four icon sizes, the small card, the four sheet sides and the vertical tabs, because "every variant, size and state" has to be true of the page REB-301 reviews. And `preflight.json`'s image checks now trigger on `shared/ui/**`, which is the hole this branch would otherwise have left: an image no pull request builds, breaking on a package its filters never mentioned.

Re-run after the fixes: package 55 tests, hub 168, CRM 1155 in 111 files, three lints, three typechecks, both builds, both web images rebuilt, and the toast clicked in a browser.

PR #194, opened 2026-09-19
test(ui): one contrast implementation, and a browser to check it

fiorelorenzo:
## What this changes

The token contract had a hole with two halves. The WCAG contrast maths was written three times, in the website's token test, in the hub's and in the package's, and one of them wrote the sRGB knee as 0.03928, which is not the number in the specification but a typo copied around the web for twenty years. The review checked what I had claimed here and it was wrong: on eight-bit colour the two knees agree exactly, since no channel value falls between them, so nothing ever measured differently. What three copies cost is the next edit to one of them, not a past one. It now lives once in `@rebase/brand/contrast`, next to the palette it measures, with `AA_TEXT` and `AA_NON_TEXT` named rather than typed as bare numbers at each call site, and the three suites import it. What deliberately does not move is resolving a stylesheet: the site keeps its own cascade walker because it owns its CSS, and an application asks a real browser rather than re-deriving the cascade in jsdom.

That browser is the second half. `shared/ui/e2e/gallery.spec.ts` drives the gallery from REB-300 under Playwright and measures what Chromium computes, which is what no stylesheet test can reach: radius, every shadow layer, the lines and the 16px grid on rendered primitives, axe over the page and over each open overlay, and an operating system emulated in dark mode to prove that nothing anywhere reaches a `dark:` utility (ORB-138). Twelve tests, on the built gallery, in about nine seconds. `preflight` gets a new `ui-e2e` check that runs it before a pull request exists, serially, next to the website's own Playwright suite, and both `ui-*` checks now trigger on all of `shared/brand/**` rather than two files of it.

It found two real things, and both are filed rather than filtered. The watermelon read as text measures 4.17:1 on the Paper ground and 3.57:1 on its own 10% tint, against the 4.5:1 AA asks of body text, which is REB-307 and a palette decision rather than something to fix in passing; the gallery now shows the `link` variants on a card, which is the only ground they clear AA on, with the reason written next to them. And radix's own focus scope makes axe report `aria-hidden-focus` while a menu or a select is open, which is REB-308. Both sit in the spec's `KNOWN` list asserted exactly, per URL, so a new violation fails the suite and so does fixing one of these without deleting its line. While measuring I also found the palette's own comment claiming 4.58:1 for that step: true against the Mint Cream ground the CRM had until 2026-09-08, not against Paper, so it now carries the measured numbers and points at the card.

## How I verified it

- `pnpm --filter @rebase/ui test:e2e` — 12 passed (8.6s), on the built gallery served by `vite preview` on 4174.
- `pnpm --filter @rebase/ui test` — 55 passed. `pnpm --filter hub test` — 168 passed. `pnpm --filter website test` — 300 passed. All three after the last edit, including the palette comment rewrite, since the site's suite reads that file.
- `pnpm --filter website lint`, `pnpm --filter hub lint`, `pnpm --filter @rebase/ui lint`, and `tsc --noEmit` in the website and the hub: clean.
- The two findings were measured, not guessed: `contrastRatio(watermelonStrong, paper)` is 4.169, `blendSrgb(watermelonStrong, paper, 0.1)` under it is 3.567, and the candidate step on REB-307 gives 5.385 and 4.544.
- The gallery renders as before; the only visible change is the link row, which now sits on a real `Card size="sm"`.

The decision to share the maths and not the cascade. The card asks to "reuse derivePairs", and I did not: the site's walker resolves `var()` chains out of its own stylesheet text, which is exactly the thing an application must not re-derive, since between a declaration and a pixel sit Tailwind's utilities and the cascade. Sharing the half that is a pure function and answering the other half with a browser is my reading of the intent, and it is the part I would argue about.

Second, `KNOWN` being an exact assertion rather than a filter. It means a fix to REB-307 or REB-308 turns the suite red until somebody deletes the line, which is deliberate but is a footgun if you do not know it. It is written in the spec's header comment and in the package README.

The CRM half of "the two per-app tests shrink" is already done and needs nothing here: REB-299 deleted the CRM's token test and moved the chart maths into the package, so what remains app-local is the hub's `.site` weights, which stay.

## Screenshots and video

The full gallery as the suite sees it, after the link row moved onto a card. Nothing else visible changed, and the rest of this PR is tests and a shared module, so there is no before-and-after pair to compose: the one thing a reviewer can see is that row.

The gallery on this branch, full page

Linear: REB-301.


fiorelorenzo:
Reviewed by an independent agent that did not write the patch. It found one blocker and I fixed all nine findings in be418b05.

The blocker was real and embarrassing in the right way: the overlay tests waited on `[data-slot$="-content"]`, and `card-content` matches that on every path, so the wait was satisfied by the first card in the Button section. For the dialog and the popover the expected axe result is the same as the bare page's, which means those two tests asserted nothing and would have stayed green with the whole `?open=` mechanism broken; the other three raced a portal they never waited for. Each overlay now waits for its own slot, and I checked both directions in a browser rather than assuming: zero `dropdown-menu-content` on `/`, one on `/?open=menu`, same shape for the dialog.

Second, the axe helper filtered violations to critical, serious and moderate, which contradicts the exact `KNOWN` assertion the suite is sold on: a new minor finding would have vanished. Every violation counts now, and the six paths return exactly the sets already written down.

Third, the spec restated three palette colours as literals, which is the second palette the neighbouring token test goes out of its way to make impossible. It reads `@rebase/brand/palette.css` through `paletteFrom` now, so a palette edit fails pointing at the token.

The rest: `hexToRgb` throws rather than reading an `#rrggbbaa` as its opaque half, `paletteFrom` captures only the lengths it can read, the sRGB transfer function is exported and the Oklab chart maths imports it instead of keeping a fourth copy of the same constant, `AA_LARGE_TEXT` and `relativeLuminance` leave the public surface since nothing imports them, and the tsconfig arrays go back to one line after my generator reformatted a file it had no business touching.

Two claims of mine were wrong, and the reviewer checked them numerically. The drift between the three WCAG copies is not observable: 0.03928 and 0.04045 return the same value for every eight-bit channel, so what three copies cost is the next edit and not a past one, and both the module header and this PR's description said otherwise. And the palette comment I rewrote in the first commit kept "3.96:1" for Prussian Blue on Watermelon, which measures 4.1685:1 against two independent implementations; the conclusion still holds, the number did not. Both are corrected in the tree.

Re-run after the last edit: `@rebase/ui` 55 unit and 12 e2e, hub 168, website 300, lint and `tsc --noEmit` clean in all three, and both Docker web images build locally (`3c82c38f`, `b96ed01a`), which is the tier CI skips on a pull request.

Full review: it rated the maths correct, checked against an independent implementation (white on `#e5133e` 4.6726, watermelon on Paper 4.1687, on its own 10% tint 3.5666), and confirmed the three rewired suites lost no assertions.

PR #195, opened 2026-09-20
chore(pigrocrm): delete what the primitive swap left behind, and say where a primitive lives

fiorelorenzo:
## What this changes

REB-300 left five generated components in the CRM because nothing else rendered them. Two of those, `calendar` and `sidebar`, had no importer in the CRM either, and never had one: they came in with a `shadcn add` and sat there. They go, and with them `react-day-picker` and `hooks/use-mobile.ts`, which existed only for those two files. What stays is what the application actually renders: `avatar` in the shell, and `command`, which composes `input-group`, in the command palette. `shadows.test.tsx` names those three now and asserts that no arbitrary shadow appears at all, rather than allowing the `0 0 0 1px` ring, which left with the sidebar.

The CRM's `components.json` goes too, which is the part that matters beyond this diff. While it existed, `shadcn add` inside the application was the easy way to grow a twentieth local primitive, which is exactly the drift `@rebase/ui` exists to end. The repository now has one `components.json`, in the package, and that is where a primitive is generated. The three `AGENTS.md` files say so: the root one finally names `shared/ui` next to `shared/brand` and `shared/analytics`, and each application's says where its UI comes from and what is genuinely local, the CRM's three files and the hub's `.site` scope.

One thing the card asked for does not apply. `src/styles/tokens.css` is not dead: it declares nothing, but it is the three-line entry file that gets the import order right (`@rebase/ui/before-tailwind.css`, Tailwind, then the tokens), and `main.tsx` imports it. It stays, and its header already says that. Same for the card's "a grep for `components/ui/` answers nothing": it answers three files, on purpose.

Last, the comments that still pointed at things that moved: the chart maths and the test that evaluates the five `color-mix()` expressions are in `@rebase/ui` now, `styles/tokens.test.ts` no longer exists, and a filter chip's line is 1px of the ink rather than the 12% tint REB-299 deleted. A comment naming a file that is gone is worse than no comment, because the next reader goes looking for it.

## How I verified it

- `pnpm --filter web test` — 1155 passed across 111 files, after the last edit.
- `pnpm --filter web lint`, `pnpm --filter web build`, `pnpm --filter web exec tsc --noEmit`: clean. `tsc` is what proves the two deleted files and the deleted hook had no importer, since an unresolved import is an error here.
- `docker build -f projects/pigrocrm/Dockerfile.web .` — `72a402a9`, which is the tier CI skips on a pull request.
- `pnpm install` regenerated the lockfile for the dropped dependency (26 lines), and the install then passes `--frozen-lockfile`.
- Grepped the tree for `components.json` and for the two deleted primitives: what is left is the package's own config, the three `AGENTS.md` sentences, and the dated design records under `docs/superpowers/`, which are a verbatim record and are not rewritten.

Deleting `sidebar.tsx`. It is the shadcn sidebar, a big file, and the CRM's shell does not use it: `AppShell` is hand-written. If somebody was planning to adopt it rather than keep the hand-written shell, this makes that a `shadcn add` in `shared/ui` instead of a file that was already here. I think that is the right direction, given REB-303 is about to rewrite the shell against the package, but it is a decision and not a cleanup.

The other one is `components.json`. Removing it means `pnpm --filter web exec shadcn add <x>` now fails in the CRM instead of silently writing a local primitive. That is the intent, and both `AGENTS.md` files say where to run it, but it will surprise anybody with the old command in their shell history.

## Screenshots and video

Nothing visible changed: the two deleted files rendered nowhere, and everything else here is a comment, a documentation sentence or a dependency. The CRM's screens are REB-303's subject, and that card carries the before-and-after pairs at 1440 and 390.

Linear: REB-304.


fiorelorenzo:
Reviewed by an independent agent. It confirmed the deletions are genuinely dead (it searched both SPAs, the website, the shared packages, the e2e suites, `index.html` and dynamic imports, and noted the CRM's `calendario` is a hand-written month grid with native date inputs), and found six things. All applied in 8d088ded.

The one that mattered: this branch wrote into three `AGENTS.md` files and a test that a primitive is generated with `pnpm --filter @rebase/ui exec shadcn add <name>`, and that command errored out. The CLI resolves every alias in `components.json` through tsconfig paths and refuses to start without them, so it had been broken in the CRM too for as long as that config existed, which nobody noticed because nobody ran it. The package now has a `@/*` entry in `paths` and aliases pointing at it, and `shadcn info` resolves against it.

Worth knowing if you ever tidy that file: the `paths` entry deliberately has no `baseUrl`. I added one first and `tsc` went red on three circular aliases, because with `baseUrl: "."` a bare `import { toast } from 'sonner'` resolves to the package's own `sonner.tsx` before it reaches the dependency. TypeScript has allowed `paths` without `baseUrl` since 4.1, and the README now records the trap.

The rest: the eslint override still scoped `src/hooks/use-mobile.ts` and still justified `set-state-in-effect: off` with that deleted hook, so both go and the comment says why the surviving `use-media-query.ts` never needed the exemption; `date-fns` turned out to be the same orphan `react-day-picker` was, arriving as its dependency and imported by nothing, so it goes with a lockfile refresh; `tokens.css` claimed the sidebar tokens are read by "the sidebar primitive", which no longer exists anywhere, and the two hand-written shells are the real consumers; and the root `AGENTS.md` counted to a "twentieth" local copy where the README in the same commit counts a nineteenth and the CRM's file a fourth, so it now says "another".

Re-run after the last edit: CRM 1155 tests across 111 files, `@rebase/ui` 55 unit and 12 e2e, hub 168, lint and `tsc --noEmit` clean in all three, both `vite build`s, and the CRM web image built locally (`72a402a9`, before this round; nothing since touches the image inputs).

PR #196, opened 2026-09-20
test(ui): record what the aria-hidden finding actually is, and assert its nodes

fiorelorenzo:
## What this changes

REB-301 recorded this finding from the rule name, and the reading was wrong. The reported nodes are not the trigger: radix marks every sibling of the open surface `aria-hidden` (the `aria-hidden` package's `hideOthers`) and leaves their contents in the tab order, so axe reports five sections of the gallery. Radix does trap Tab inside the surface, so nothing outside is reachable in practice, but axe reads the static DOM and cannot see a focus trap, and the remedy it accepts is `inert` on the background, which radix does not use.

The part that explains why only two of the five overlays show it, which is what convinced me this is upstream and not ours: a dialog and a sheet set the identical attributes on the identical sections, with the identical 43 tabbable elements inside them, and are not reported at all. axe suppresses this rule while a modal is open, and its modal detection looks for `role="dialog"`; a `role="menu"` and a `role="listbox"` are not that, so the same DOM is a violation there and clean here.

So the card's three options collapse to one. There is no version to upgrade to (radix-ui 1.6.7 is the latest release, and the workspace is on it), the second option's premise was the wrong cause, and the upstream issues that look like this one are about a different node (the focus guards, radix-ui/primitives#3822) or a different component (#3085). It stays recorded, with the mechanism written into the spec and the package README so the next reader does not re-derive it from the rule name the way I did.

What does change is the strength of the check. `KNOWN` now carries the rule and the element it was reported on, not just the rule, so the same rule appearing on a different element fails instead of landing on a line that already looked accounted for. Targets are only appended when they are a single stable selector: the sections have ids, while the two colour findings sit on utility-class chains that change with the variant, so those stay bare rule names.

## How I verified it

- Dumped every axe node on all six paths with a throwaway spec: `aria-hidden-focus` is reported on `#button`, `#card`, `#fields`, `#floating-surfaces` and `#tabs` for the menu and the select, and on nothing for the dialog, the sheet and the popover.
- Measured the DOM per state in a browser: all four modal overlays mark the same eight sections `aria-hidden`, none of them uses `inert`, none sets `aria-modal`, and each leaves the same 43 tabbable elements, 6 with a negative tabindex and 11 disabled, inside the hidden subtree. That is what rules out "the dialog is doing something different".
- `npm view radix-ui version` is 1.6.7 and the lockfile resolves 1.6.7.
- `pnpm --filter @rebase/ui test:e2e` — 12 passed. Then, to prove the new assertion can fail, I replaced one expected node with a selector nothing matches: 1 failed, 11 passed, and the failure named the menu test.
- `pnpm --filter @rebase/ui test` — 55 passed, lint and `tsc --noEmit` clean.

Whether a known-and-explained violation should stay in an assertion at all, rather than the suite dropping the rule for those two paths. I think it should: dropping the rule would also hide a genuine `aria-hidden-focus` introduced by our own markup, and the node-level entries now make that distinction mechanical.

The one thing I did not do, because it is not mine to do: file the upstream issue. radix has no issue for this specific node on `DropdownMenu` and `Select` (the closest are about the focus guards and the navigation menu), and filing one under your account on somebody else's repository is your call, not mine. Say the word and I will write it against the evidence above.

## Screenshots and video

Nothing visible changed: this is a test assertion and two documentation passages. The gallery's render is unchanged from #194.

Linear: REB-308.


fiorelorenzo:
Reviewed by an independent agent, and it was worth it: seven findings, one blocker, all applied in 06196a73.

The blocker is the kind that would have sat there for months. `axeIds` computed the bare-rule fallback once per rule, so once `aria-hidden-focus` had its five id nodes, any further node of the same rule on an element without an id contributed nothing and the assertion still matched. The reviewer reproduced it by adding a focusable element to the page header, which is `aria-hidden` and has no id: axe reported a sixth node and the suite stayed green. The fallback is per node now, and I re-ran the same repro to watch it fail: `+ "aria-hidden-focus"` on both the menu and the select test, and the tree green again once reverted.

Then the mechanism, which I had wrong a second time in the commit that existed to fix the first wrong version. The sections are not siblings of the open surface: the surface is portalled to `<body>`, whose children are `#root` and the two focus guards. `hideOthers` keeps every `[aria-live]` region and its ancestors, sonner's toaster is one of those and sits inside `<main>`, so the marks land on `<main>`'s own children instead of stopping at `#root`. I checked it in a browser rather than taking it on trust: with the menu open the marked elements are the header plus all eight sections, and the page's only `[aria-live]` is the toaster inside `<main>`.

And axe does not suppress the rule for a dialog. `focusable-modal-open` returns undefined while a modal is open, so the finding lands in `incomplete`, on seven nodes, not in `violations`. "Suppresses" reads as "axe decided this is fine", which is exactly the belief this record exists to prevent, so the wording is fixed and the spec now says the two modal URLs assert nothing about this rule. The detection is also `dialog, [role=dialog], [aria-modal=true]`, not `role="dialog"` alone.

The rest: the five sections are declared once and shared by the two URLs, the README no longer promises element precision for the colour entries (which stay bare rule names, since their targets are utility-class chains that move with the variant), and it no longer says "the two colour findings" where axe reports one node on each of three URLs.

Re-run after the last edit: 12 e2e passed, 55 unit passed, lint and `tsc --noEmit` clean.

PR #197, opened 2026-09-20
feat(brand): give the watermelon a third step, dark enough to be read as text

fiorelorenzo:
## What and why

The watermelon had two steps and both answered the same question: how dark it has to be for **white text on it** to clear AA. Raw is 4.22:1, `-strong` is 4.67:1. Read the other way, as a label on a light ground, `-strong` measures **4.17:1 on Paper** and **3.57:1 on its own 10% tint**, so `text-destructive` (every form error in both applications) and the `link` variant were below AA anywhere but on a white card. That is what REB-301's contrast check reported, and what the gallery worked around by rendering the link variants inside a card.

Lorenzo chose option **a** of REB-307: a third step. `--color-watermelon-deep: #c50d33` gives **5.39:1 on Paper**, **4.54:1 on its own 10% tint** and **6.04:1 under white text**, so the fills gain as well. `--primary` and `--destructive` point at it and no component changes, since both slots are read as text as often as they are filled. `-strong` keeps the two fills that only ever carry white text and never text in the colour: the site's call to action and the CRM sidebar's active tile.

Two things worth flagging rather than burying. The application now has **two reds on a fill**, the primary button's and the sidebar tile's, seven points of lightness apart and never adjacent on a screen; whether the tile follows the button is a decision and not a defect, so `tokens.css` says so and this PR does not touch it. And the chosen hex is a hair more saturated rather than a pure lightness step (h=347.6 s=87.6% l=41.2%): the strict continuation of the ramp is `#c31035` and measures within 0.05 of it on every pair, so the two are interchangeable by the numbers.

## What changed

- `shared/brand/palette.css`: the third step, and the comment rewritten to say which step each surface reads and why. The old text still explained `-strong` with 4.58:1, a number measured against the Mint Cream ground the CRM lost on 2026-09-08.
- `shared/ui/tokens.css`: `--primary` and `--destructive` re-pointed; the sidebar comment says why its tile stays on `-strong`.
- `shared/ui/e2e/gallery.spec.ts`: the three `color-contrast` entries leave `KNOWN`, so axe now reports **no contrast violation on any of the six URLs**. The measurement test became the proof of the fix: it asserts the colour the browser computes for the destructive badge and the link button, and that the nearest opaque ground under both is Paper, so a revert to `-strong` fails here instead of passing on a card. The invalid-field line assertion follows `--destructive`.
- `shared/ui/gallery/Gallery.tsx`: every button variant renders on the page ground, the link included, and the link badge joins the badge row. The card wrapper and its comment are gone: they existed only because the system did not allow that pair.
- `shared/ui/tokens.test.ts`: the deep step is measured as text on the four grounds the two variants actually land on (Paper, a white card, and the 10% tint over each), and white is measured on both filled steps.
- `projects/website/`: the plugin extracts eight tokens now, with a comment saying the landing receives the new one without drawing with it. Three stylesheet headers and the brand README counted the shared tokens wrong (fifteen, seven, six, against a real seven), which this corrects.
- `docs/design/DECISIONS.md`: the row.

Deliberately not touched: `docs/design/2026-09-18-application-variant-brief.md` still says "six colours" and "never a seventh hex". It is a dated record of the decision that was taken then, and the row in `DECISIONS.md` is what supersedes it.

## Evidence

- `@rebase/ui`: **60 unit** and **12 e2e** passed, `tsc --noEmit` and eslint clean, the gallery builds.
- `website`: **300 passed** (the plugin's count guard was the one real failure this change produced), `tsc --noEmit` and eslint clean, build clean.
- `web` (CRM): **1155 passed**. `hub`: **168 passed**.
- Read off the render in Chromium at 1280 wide: the destructive badge, the link badge, the link button and the primary fill all compute `rgb(197, 13, 51)` on a body of `rgb(241, 242, 243)`.

__omp_shell("[buttons and badges on the page ground](https://github.com/user-attachments/assets/9102ef81-da48-45d9-9ab5-c3afc1d19be2)")

## Risk

The palette is shared, so every surface moves at once. The website is unaffected by design: its call to action fills with `-strong` and its own text-pair test measures that pair, and the community page's error note reads `-strong` on a white box (4.67:1), so no site pair got worse. In the applications the change is strictly upward on every pair it touches. The one thing a reviewer should disagree with if they think it is wrong is the two-reds observation above.

Linear: REB-307.


fiorelorenzo:
Reviewed by an independent agent. Ten findings, two of them blockers and both real, applied in 679f43e4.

**The test could not fail for the number it was written for.** `blendSrgb` takes a percent, and both new tint rows passed `0.1`, which blends 0.1% and hands the ground back untouched: they measured 5.38 and 6.04, the same easy pairs as the two rows above them, while the tightest number in the whole change is 4.54:1 on the 10% tint over Paper. The reviewer made that concrete: a step of `#c8112f` passes the row as written at 5.20 and fails the pair it exists to guard at 4.42. Fixed, and taken further on a second finding: the rows resolve the slot now instead of measuring `--color-watermelon-deep`, since a test that measures the hex proves a property of a colour no slot has to point at. I reverted `--primary` and `--destructive` to check: five unit tests fail now, where before the revert only the source-text regex noticed.

**The second blocker is the failure mode a shared palette has.** The hub's four public pages wrap everything in `.site`, that scope paints no fill of its own, and every primitive fills from `--primary`, so moving `--primary` moved the wizard's «Avanti» to the application's red while the landing's «Entra come talento», which is the button that leads into it, stayed on the site's. One journey, two reds, no test. The scope now re-points `--primary` and `--primary-foreground` at its own `--landing-cta`, which is what its comment had claimed all along, and the hub's token test pins it. Measured in Chromium on the built hub at `/hub/freelance`: the CTA computes `rgb(229, 19, 62)` inside `.site` while the document root stays `rgb(197, 13, 51)`.

__omp_shell("[the wizard, with the site's own red back on Avanti](https://github.com/user-attachments/assets/5d89b2e2-b279-4be9-84e9-1b454fb57ca3)")

**Three numbers were wrong.** 5.39:1 on Paper is 5.38:1 (in the palette, the record and the gallery); "within 0.05 of `#c31035` on every pair" is 0.06, the widest gap being 0.059; and the sidebar's Paper-on-blue is 15.70:1, not the 14.7:1 that comment has carried for a while. A change whose whole subject is contrast cannot state a ratio loosely.

**Two claims were wider than the evidence.** `-strong` is not "two fills that never carry text": the community page writes its error note in it (4.67:1 on the white box it sits in) and the dashboard's `costi` bar fills with it under white labels. And "may now be written anywhere" holds to 10% of its own tint and no further, 4.17:1 at 15% and 3.80:1 at 20%, both live in the CRM today (`DealStageBar`'s lost stage, the destructive badge's hover) and both better than they were before this change. That is **REB-309**, filed with the measurements and three options, and the ceiling is now written into the palette, the tokens and the test rather than implied.

Smaller: the plugin's "seventh colour" read as an ordinal and pointed at charcoal blue, `landing.css` still said five colours, and the CRM's `RequiredMark` still asked the question this card answers.

Re-run after the last edit: 62 UI unit, 12 e2e, 300 website, 169 hub, 1155 CRM, all green; lint and `tsc --noEmit` clean on all four packages; hub and website build.

PR #202, opened 2026-09-21
feat(hub): a talenti read model merges Developer e CTO and Iscrizioni into one list

fiorelorenzo:
## What this changes

Adds a `talenti` read model: one endpoint (`GET /api/hub/talenti`) that merges every freelancer card and every bare sign-up (a signup with no matching card) into a single list, newest first, with a stable id, `stato` (`lead` for the bare sign-ups, the card's real state otherwise), an `origine` (`form`/`wizard`/`admin`), the UTM source, and per-state counts unbounded by the page limit.

This is additive: `GET /api/hub/freelancers`, `GET /api/hub/companies`, `GET /api/hub/signups` and the MCP tools are untouched. REB-283 (dispatched next) migrates the SPA to the new endpoint and removes the old admin nav entries.

Per REB-282's own recommendation, the `signups` table stays separate from `freelancers` and is unified only in this read model: a sign-up that later fills the wizard carries two UTM sets, and folding the tables would drop one.

## How I verified it

`uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests` -> 287 passed.
`uv run mypy projects/hub/packages/core/src projects/hub/apps/api/src` -> no issues in 38 source files.
New tests cover: a bare signup reads `stato=lead`/`origine=form`; a wizard-filled card reads its real state/`origine=wizard`; an admin-drafted card reads `origine=admin`; per-state counts are correct and unbounded by `limit`; ordering spans both tables; the `stato` filter selects leads, a single state, or nothing for an unknown value; `AdminDep` is enforced (401 signed out, 403 signed-in non-admin).

The merge-then-truncate-in-Python approach in `TalentiService.list_recent` (fetches top-`limit` from each table, merges, truncates to `limit`) rather than a SQL `UNION`: correct because both sources are already sorted newest-first, so no row outside either source's own top-`limit` can land in the combined top-`limit`. Worth a second look if someone later changes the ordering.

## Migrations

None. `talenti` is a pure read model over the existing `freelancers` and `signups` tables; no new column or index. The bare-signup anti-join reuses the same unindexed pattern `FreelancerService._leads` already runs today.

## Screenshots and video

No UI change: this PR is API-only (a new admin-gated endpoint with no frontend consumer yet). REB-283 adds the screen this feeds.

Linear: REB-282.


fiorelorenzo:
**Review applied:** 1 finding (test coverage gap: truncation across the two merged sources was only asserted by count, never by row identity). Added the missing assertion in `dc15f5f0`. The core merge logic itself was confirmed correct by the reviewer.

PR #203, opened 2026-09-21
fix(hub): the signed-in shell runs full width and the sidebar stays pinned to the viewport

fiorelorenzo:
## What this changes

The signed-in shell (`SignedInLayout.tsx`, every member and admin page) no longer boxes page content away from the sidebar: the bordered/padded card wrapper around the page outlet is gone, so content runs full width flush against the sidebar with only its own page-level padding. The root frame is now viewport height (`h-screen` instead of `min-h-full`) and stays that way regardless of page content; the sidebar is pinned to that height and independently scrollable, and the main content area scrolls on its own axis, so a long admin table no longer scrolls the sidebar away with it.

The admin nav group also gets an "Amministrazione" eyebrow label above its separator, so it is visually unambiguous which sidebar items are admin-only.

## How I verified it

`pnpm --filter hub exec vitest run src/pages/SignedInLayout.test.tsx` -> 10 tests passed, including new assertions that the admin eyebrow renders only for `role: admin`, the sidebar carries `h-full overflow-y-auto` inside an `h-screen` parent, main carries `overflow-y-auto`, and no bordered card wrapper remains around the outlet.
`pnpm --filter hub exec eslint src/pages/SignedInLayout.tsx src/pages/SignedInLayout.test.tsx` -> clean.
`pnpm build` (apps/web) -> built successfully, no type errors.
Visual: `vite preview` on `/hub/io`, faked both an admin and a member session via an intercepted `GET /api/hub/me`, confirmed the admin eyebrow and flush layout on both, then loaded the admin list with 40 mocked rows to force overflow and confirmed via `page.evaluate` that scrolling `main` to 800px left `aside.scrollTop` at 0 with `email`/`Esci` still visible at the bottom and `window.scrollY` never moving: the sidebar is pinned, only `main` scrolls.

Page-level padding: removing the wrapper meant checking that `member/Area.tsx` and `admin/lists.tsx`'s `Header` don't rely on the wrapper's own padding. They carry their own (`Area.tsx`'s outer `mx-auto max-w-2xl`, `lists.tsx`'s `Header`'s `px-6 py-5`), so `<main>` keeps a plain `p-6` and nothing collides, but it's worth a second look on both pages.

## Screenshots and video

Not attached: this session has no interactive `gh --attach` step wired for this delegated run. The visual verification above (admin/member views, forced-overflow scroll test) was done live via the browser tool against `vite preview`; a reviewer can reproduce it with `pnpm --filter hub preview` from this branch.

Linear: REB-311.


fiorelorenzo:
**Review applied:** 2 findings, both fixed in `7f810023`. (1) `h-screen` reintroduces a viewport-unit clipping bug this codebase already fixed once (2026-09-08 incident, documented in `shared/ui/tokens.css`'s html/body/#root height:100% comment); reverted to `h-full`, matching pigrocrm's AppShell convention. (2) Dropping the bordered wrapper made `hover:bg-muted` invisible on every admin table since `--muted` equals `--background` with nothing behind it; `main` now keeps `bg-card` (without the removed border/radius).

PR #209, opened 2026-09-21
fix(hub): the member area header gets breathing room and its two perk buttons align

fiorelorenzo:
## What this changes

The member area header gets a clear top margin above «LA TUA AREA» instead of sitting flush against the shell's edge, and the two perk cards' buttons ("Apri PigroCRM", "Scarica la guida") now land at the same height regardless of description length or whether a trailing footer note follows. Measured before/after with `getBoundingClientRect()`: a 28px mismatch between the two buttons before this change, 0px after.

## How I verified it

`pnpm --filter hub exec vitest run src/pages/member/Area.test.tsx` -> 7 passing.
`pnpm build` -> clean.
Live browser check on `vite preview`, this branch versus `main`: bounding-rect comparison showed the buttons at `top: 782` and `top: 754` (28px apart) before, both at `top: 777` after.

The footer-note slot on the PigroCRM card is rendered `invisible`+`aria-hidden` to reserve identical space to the guide card's real footer text, rather than a fixed-height spacer div. Both approaches work; this one keeps the markup symmetric between the two cards.

## Screenshots and video

Not attached in this delegated run; the bounding-rect numbers above are the measured proof.

Linear: REB-312.


fiorelorenzo:
**Review applied:** 1 finding, fixed in `5d641abf`. The header's own added padding stacked on top of SignedInLayout main's existing p-6, doubling the gap to 48px (64px in the negato branch) instead of a single deliberate increment; reduced to pt-2 for a clear but modest 32px total.

PR #210, opened 2026-09-21
feat(hub): talenti and companies are searchable and cursor-paginated

fiorelorenzo:
## What this changes

`GET /api/hub/talenti` and `GET /api/hub/companies` gain cursor pagination (keyset over `(sort value, id)`, opaque base64url token) and a `q` search parameter (escaped ILIKE match, ordered by `pg_trgm` similarity among matches) across the fields the ticket names: for talenti, name/surname/email/posizione plus filters (stato, posizione, remoto, tariffa_giornaliera min/max, origin, UTM source, has a CV, has logged in, created-between); for companies, nome_azienda/referente/email/progetto plus filters (stato, budget_giornaliero min/max, periodo_da, origin, created-between). Per-state counts now respect every active filter except `stato` itself. `GET /api/hub/freelancers`, `GET /api/hub/signups` and every MCP tool are untouched; both changed endpoints stay backward compatible (new params are additive with the old defaults).

## How I verified it

`uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests` -> 306 passed (independently re-run).
`uv run pytest -q projects/hub/apps/mcp/tests` -> 17 passed, run as a safety check since `CompanyService`'s signature changed; `apps/mcp` itself has an empty diff against the branch point.
New tests cover: search hits on a partial surname and an email domain; the cursor never repeats or skips a row walking every page against a larger-than-one-page dataset; an unknown/malformed filter is a 422; `AdminDep` still enforced.

`talenti`'s cursor still merges two independently-sorted, independently-limited queries (freelancers + bare signups) in Python rather than a SQL `UNION`, now extended to a cursor-bounded page -- the same top-limit-merge argument REB-282 already made, extended here; worth a second look if the merge or sort key ever changes.

## Migrations

Revision `0013`, `depends_on` `0012`. Creates `pg_trgm` (unguarded on purpose, so a managed Postgres that forbids it fails the deploy loudly rather than starting an app that silently scans every row) and nine GIN trigram indexes (`users.nome/cognome/email`, `signups.nome/cognome/email`, `freelancers.posizione`, `companies.nome_azienda/progetto`), every statement `IF NOT EXISTS` per this package's inherited-database convention. No `CONCURRENTLY` (Alembic runs each migration inside a transaction, which forbids it), matching PigroCRM's own `0021_pg_trgm_search_indexes.py` and its stated reasoning. Safe on a database where these tables already hold real rows: adds an extension and nine indexes, touches no row or column, costs nothing to run twice.

## API changes

`GET /api/hub/talenti`: adds `q`, `cursor`, `posizione`, `remoto`, `tariffa_min`, `tariffa_max`, `origine`, `utm_source`, `has_cv`, `con_accessi`, `creato_da`, `creato_a`, all optional, additive. Response gains `next_cursor` alongside the existing `totale`/`items`/`per_stato` (per-state counts, now respecting every active filter but `stato` itself).
`GET /api/hub/companies`: same shape, with `budget_min`/`budget_max`/`periodo_da`/`origine`/`creato_da`/`creato_a` in place of the talenti-specific fields.
The hub's hand-written client (`apps/web/src/lib/api.ts`) is untouched in this PR -- REB-286 wires the frontend to these new params.

## Screenshots and video

No UI change: this PR is API and migration only. REB-286 adds the screen.

Linear: REB-285.


fiorelorenzo:
**Review applied:** the reviewer traced the keyset predicate's tie-break algebra by hand and reproduced it live against Postgres 17 with engineered ties (including a forced similarity-score tie straddling the freelancer/signup boundary), confirming the pagination and search are correct as shipped. The one real defect was in the PR description itself: the API changes section named four param/field names (`ha_effettuato_accesso`, `created_after`, `created_before`, `conteggi`) that don't exist in the diff, which risked steering REB-286's frontend work onto the wrong query params. Corrected to the actual names (`con_accessi`, `creato_da`, `creato_a`, `per_stato`).

PR #211, opened 2026-09-21
feat(hub): a talent's page shows the sign-up, the logins, the downloads and the Pigro space

fiorelorenzo:
## What this changes

A talent's detail page now shows everything the hub knows about the person in one place: the card, the state editor, the origin sign-up's UTM set, the last 5 logins and last 5 guide downloads, and the PigroCRM space slug when the address owns one, then the unchanged comments thread. `GET /api/hub/freelancers/{id}` answers all of it in one call (`FreelancerDetail`, extending `FreelancerRead`) rather than the page fanning out to four endpoints.

## How I verified it

`uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests` -> 322 passed (independently re-run).
`uv run ruff check/format --check projects/hub` -> clean after a formatting fix.
`pnpm --filter hub exec vitest run src/pages/admin/lists.test.tsx` -> 9 passed (independently re-run).
`pnpm build` -> clean (independently re-run).
Visual: seeded a scratch Postgres and a fake PigroCRM registry HTTP server, ran the real API and SPA, `uishot` at 390/768/1440 in light+dark with `--axe`: zero violations, zero console errors, all four new sections render in order with real data.

`FreelancerService`'s constructor gained optional `settings`/`http` params purely for the Pigro lookup; every existing caller (MCP server, other admin routes, ~20 test call sites) is unaffected and simply gets `pigro_slug=None`. The Pigro HTTP call is gated on `settings.pigro_registry_token` being configured, so no test or unconfigured deployment makes a stray network call. `schemas.py` is touched by both this PR and REB-314 (company self-edit, in flight separately) in non-overlapping regions of the file; should merge cleanly but the two will need normal rebase handling since they share a file.

No new admin-side editing beyond what exists today (state and comments); a member edits their own card. Found and filed separately (REB-316, not fixed here): the admin shell's sidebar doesn't collapse below ~600px, reproduced identically on `/admin/talenti`, a page this PR never touches, so it predates this change.

## Screenshots and video

Not attached in this delegated run; the `uishot` proof above (390/768/1440, light+dark, `--axe`) is the verification record.

Linear: REB-284.


fiorelorenzo:
**Review applied:** 3 findings, all fixed in `ace36438`. (1) get_freelancer/create_freelancer_from_signup silently started returning the new detail fields plus an always-null pigro_slug once get() began answering FreelancerDetail, contradicting the PR's own claim that MCP callers were unaffected; both tools now truncate back to FreelancerRead explicitly. (2) find_by_email's genuine no-match path (configured token, real 200, no matching row) had no test; added one. (3) The sign-up's own origine was fetched and never displayed; now shown.

PR #212, opened 2026-09-21
feat(hub): a company contact can edit their own request the way a freelancer edits their card

fiorelorenzo:
## What this changes

A company contact who signs in can now edit their own request the same way a freelancer edits their card: `GET /me` gains `ha_azienda` and the four editable fields (progetto, periodo_da, durata, budget_giornaliero), `PATCH /me/azienda` updates them (never stato/note/nome_azienda/referente identity, structurally impossible to post), and the member area renders a company section with its own edit form, parallel to and independent of the freelancer section -- a person can have either, both, or neither. Self-edit reaches only the most recently created company request when there are several, per the card's own decision; older requests stay admin-editable only.

## How I verified it

`uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests` -> 321 passed (independently re-run).
`uv run ruff check/format --check projects/hub` -> clean after a formatting fix.
`pnpm --filter hub exec vitest run src/pages/member/Area.test.tsx src/pages/member/ModificaAzienda.test.tsx` -> 14 passed (independently re-run).
`pnpm build` -> clean (independently re-run).

`CompanyUpdate(CompanyFields)` inherits `extra='forbid'` so `stato`/`note` are rejected at the schema level, not filtered in application code -- tested directly (`test_members.py`). `schemas.py` and `members.py` are touched by both this PR and REB-284 (talent detail, in flight separately); my edits sit in different, non-overlapping regions, so they should merge cleanly, but the two will need normal rebase handling since they share files.

## Migrations

None: this reads and writes existing `Company` columns, no schema change.

## API changes

`GET /me`: response gains `ha_azienda: bool`, `progetto`, `periodo_da`, `durata`, `budget_giornaliero`, additive alongside the existing freelancer fields.
`PATCH /me/azienda` (new): accepts only the four fields above, uses `MeDep`. 404 for a user with no company.
The hub's hand-written client (`apps/web/src/lib/api.ts`) is updated in this PR (`member.updateCompany`), unlike REB-285's backend-only pattern, since the frontend ships in the same PR here.

## Screenshots and video

Not attached in this delegated run.

Linear: REB-314.


fiorelorenzo:
**Review applied:** 1 finding, fixed in `5d02ad98`. A server-side durata refusal fell through to the generic failure banner instead of highlighting "Per quanto", since durata has no field id of its own in this form; now remapped to periodo_da the same way CompanyWizard.tsx already does for the identical mismatch.

PR #214, opened 2026-09-21
feat(hub): Talenti and Aziende get a debounced search, filters, and infinite scroll, all in the URL

fiorelorenzo:
## What this changes

Talenti and Aziende both get a debounced (300ms) search box, a filter row for the fields REB-285's backend now offers (posizione, remoto, daily rate range, origin, UTM source, has-a-CV, has-logged-in, created-between for Talenti; budget range, start date, origin, created-between for Aziende), the existing state pills kept, and infinite scroll (a "Mostra altri" button plus an auto-load sentinel) walking the backend's `next_cursor`. All of search, filters and state pills live in the URL via TanStack Router's `validateSearch`, so a reload or a shared link reproduces the exact view. Two distinct empty states in Italian: no results for the active filters versus a genuinely empty table.

## How I verified it

`pnpm --filter hub exec vitest run src/pages/admin/lists.test.tsx src/lib/api.test.ts` -> 29 passed (independently re-run).
`pnpm --filter hub build` (vite build + tsc --noEmit) -> clean (independently re-run).
`pnpm --filter hub exec eslint` on every touched file -> clean.
Live browser check against the real Vite dev server with request interception: confirmed the filter row and state pills render at 1440px, typing in the search box updates the URL to `?q=...` only after the debounce settles, a state pill and a search term coexist in the URL together, and the IntersectionObserver sentinel auto-loads a second page with no click.

Worked around a real TanStack Router type-inference collapse (`useNavigate({ from })` combined with `useSearch` under a pathless layout id resolves to `never` in this router version) by calling `useNavigate()` with no `from` and an explicitly-typed search reducer at each call site instead. Isolated with a minimal repro; worth a second look if the router version ever changes.

No visual restyle beyond the new controls themselves (Shared UI v1). Ruff was not run: this PR touches zero Python files.

## Screenshots and video

Not attached in this delegated run; the live browser verification above is the record.

Linear: REB-286.


fiorelorenzo:
**Review applied:** 1 finding, fixed in `cac55faf`. TanStack Router's default parseSearch runs JSON.parse on every raw query value before validateSearch, so a purely numeric value in the URL (?tariffa_min=50) arrived as a JS number, silently discarded by strParam's string guard; invisible on an in-app navigate(), only showing up on a reload or a shared link, which is exactly what this PR's own description promises. Fixed strParam to coerce the post-parse number/boolean back to string, exported it, and reused it in the test harness (which had drifted into a hand-duplicated copy that never actually called it); the new regression test fails without the fix and passes with it.

PR #215, opened 2026-09-21
feat(hub): Amministratori, Agenti, Istanze Pigro and Accessi get search and pagination

fiorelorenzo:
## What this changes

Amministratori, Agenti and Accessi's login log all get real server-side cursor pagination and trigram search, reusing REB-285's `rebase_core.pagination`/`search` helpers (Amministratori stays oldest-first with no term, per its own "reads as a history" convention, which needed a new `descending` flag on the shared keyset predicate; Accessi's four aggregate counters are unchanged, only its `recenti` sub-list is no longer hardcoded to 20). Istanze Pigro is genuinely different: PigroCRM's registry endpoint takes no query params of its own, so this one stays a client-side filter over the already-fetched full list rather than pretending to a server-side capability that doesn't exist.

## How I verified it

`uv run pytest -q projects/hub/packages/core/tests projects/hub/apps/api/tests projects/hub/apps/mcp/tests` -> 366 passed (independently re-run, mcp included to confirm the untouched `AdminTokenService.list()` and `PigroRegistry` stayed that way).
`uv run mypy projects/hub/packages/core/src projects/hub/apps/api/src projects/hub/apps/mcp/src` -> clean, 45 source files (independently re-run).
`uv run ruff check/format --check projects/hub` -> clean (independently re-run).
`pnpm --filter hub exec vitest run src/pages/admin src/lib/api.test.ts` -> 45 passed (independently re-run).
`pnpm --filter hub build` -> clean (independently re-run).

Accessi.tsx dropped a `Link` to `/admin/freelance/$id` keyed on a login's `user_id` -- it was wrong on two counts (a login isn't necessarily a freelancer's own id, and the field it read never actually existed on the wire), now plain text. Worth confirming nothing else relied on that link.

## Migrations

Revision `0014`, depends on `0013`. Adds one GIN trigram index on `admin_tokens.nome` (the one searchable column with no index yet), `IF NOT EXISTS`, no `CONCURRENTLY`, same reasoning as `0013`. Safe on a database with existing rows.

## API changes

`GET /api/hub/admins`: now takes `q`/`cursor`/`limit`, returns a wrapper (`items`/`next_cursor`) instead of a bare array.
`GET /api/hub/tokens`: same shape change, via a new `AdminTokenService.list_page` (the original unpaginated `.list()` is untouched, still used by `apps/mcp`).
`GET /api/hub/logins`: gains `q`/`cursor` on its `recenti` field; `totale`/`membri`/`membri_totali`/`ultimi_7_giorni` unchanged.
The hub's hand-written client (`apps/web/src/lib/api.ts`) is updated in this same PR.

## Screenshots and video

Not attached in this delegated run.

Linear: REB-313.


fiorelorenzo:
**Review applied:** 1 finding, fixed in `6c24b916`. Istanze Pigro stayed entirely client-side, but the justification (PigroCRM's own endpoint takes no query params) only constrains the upstream call, not the hub's own facade route, which already holds the full registry in memory per request. GET /api/hub/pigro/istanze now searches and cursor-paginates that already-fetched list server-side, the same shape the other three lists in this card use.

